PR #1, opened 2021-07-25
Expose parser in lib.zig

moosichu:
This is quite a simple PR - but just makes the parser part of the public-facing API.


Vexu:
Thanks!

PR #2, opened 2021-07-25
Add appending include dirs to the public facing API

moosichu:
This is quite a simple PR - but just makes adding include directories part of the public-facing API.


Vexu:
I don't think I want to add functions like this at least yet since the API is probably going to change and struct fields generally are a part of the public facing API.

moosichu:
Good point! Cheers

PR #4, opened 2021-08-30
handle implicit int-to-ptr cast in condexpr

ehaas:


Vexu:
Thanks! Be sure to share any thoughts on the design based on your experience in working on translate-c.

ehaas:
Looks good so far - still familiarizing myself with the code. I'll take a shot at finishing up the stuff in the binary expressions test file.

PR #5, opened 2021-09-03
condexpr type resolution + null ptr cast + some tests

ehaas:
Adds a new implicit cast type `qual_cast` for implicit casts due to mismatching pointers on conditional expressions

Adds a new implicit cast type `null_to_pointer` to convert `0` to a pointer of appropriate type

Adds a way to test expression types.

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -200,6 +200,7 @@ pub const Tag = enum {
     align_ignored,
     zero_align_ignored,
     non_pow2_align,
+    pointer_mismatch,
 };
 
 const Options = struct {
@@ -224,6 +225,7 @@ const Options = struct {
     @"cast-qualifiers": Kind = .warning,
     @"array-bounds": Kind = .warning,
     @"int-conversion": Kind = .warning,
+    @"pointer-type-mismatch": Kind = .warning,
 };
 
 list: std.ArrayList(Message),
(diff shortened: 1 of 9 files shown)


src/Diagnostics.zig
```diff
@@ -22,6 +23,10 @@ pub const Message = struct {
             expected: u32,
             actual: u32,
         },
+        incompatible_types: struct {
```
Vexu, line 26: This adds 16 bytes to `Message` in release modes, I think it would be better to add a `typePairStr` function to `Parser` and put that in the error message.

`Tag` should also probably be changed to a tagged union.

test/README.MD
```diff
@@ -34,6 +34,22 @@ long float a;
...
+in the function
+
+**example:**
+```c
+void my_test(void) {
+    int x = 5;
+    1L;
+    (void)1L;
+}
+#define EXPECTED_TYPES \
```
Vexu, line 49: This seems like a decent solution for now but eventually we'll wan to switch to using `__builtin_types_compatible` and similiar functions combined with static asserts.

Vexu (approved):
This will need to be rebased and updated to take into consideration decayed array types.

src/Diagnostics.zig
Vexu: I did the first part in 9d6ef755c6d3066453bef7836527550ee70d8425.

ehaas: I rebased everything onto the latest and got rid of the`incompatible_types` field. It ended up being easier to squash all my commits together since a lot of the changes were intermingled with the latest updates.

Vexu:
Thanks! This doesn't yet handle decayed array types but fixing that is a bit more involved so it can be done separately.

ehaas:
Ah, I misunderstood your comment about rebasing. I can take a look at that over the weekend.

Vexu:
No worries, here's a small test case.
```c
void foo(void) {
    int a[3], b[3], c[4];
    (1 ? a : b)[4];
    (1 ? a : c)[5];
}
```
Clang and gcc won't complain about the out-of-bounds access on the first one but I don't see a reason why we shouldn't give a warning if we have the information like we also do with parameters:
```c
void foo(int foo[2]) {
    foo[3]; // clang and gcc won't complain about this, but we should since the user bothered to give us a size
}
```

PR #7, opened 2021-09-03
improve _Static_assert

ehaas:
Compile error if the condition is not constant
Nicer error message if the message is not included (this is a C++17 extension)


Vexu:
> Nicer error message if the message is not included (this is a C++17 extension)

It does seem quite likely that one-argument _Static_assert will be included in C2x so I think another good option here would be to implement it and put it before a `-Wc2x-extension` warning especially considering that clang and gcc both accept it.

PR #8, opened 2021-09-06
Parser: add typeof support

ehaas:
Two questions:

1) bare `typeof` is a GNU extension, so if GNU extensions are turned off it should be an identifier. But it doesn't seem like the Tokenizer has any concept of language options - is that something we should add?

2) I'm not storing the typeof nodes in the AST. I think that would be fine for `translate-c` but other things may want them. However I think it might be a more invasive change since it means we'd need a way to link types to nodes.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -472,6 +472,9 @@ fn nextExternDecl(p: *Parser) void {
             .keyword_union,
             .keyword_alignas,
             .identifier,
+            .keyword_typeof,
+            .keyword_typeof1,
+            .keyword_typeof2,
             => if (parens == 0) return,
             else => {},
         }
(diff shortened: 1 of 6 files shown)


Vexu:
1.  `Tokenizer` does seem like the easiest place to handle this. All this detection would require is a `comp: *Compilation` field so this shouldn't be too hard to add. This is not a blocker though so this could be turned into an issue.
2. We will eventually need "meta" types for attributed types so this could be solved by adding a `typeof` type like clang and store a node index like `VLA` does, this could also be done for `typedef` but I'm not sure what use that would have. This can also be dealt with a tracking issue.

src/Parser.zig
```diff
@@ -813,6 +816,27 @@ pub const DeclSpec = struct {
...
+        if (try p.typeName()) |ty| {
+            try p.expectClosing(l_paren, .r_paren);
+            return ty;
+        } else {
+            p.tok_i = start;
+            if (p.eatToken(.r_paren)) |r_paren| {
+                try p.errTok(.expected_expr, r_paren);
+                return error.ParsingFailed;
+            }
+            const typeof_expr = try p.expr();
```
Vexu, line 832: This expression should not be evaluated, see `genericSelection` for how to do that.

The same applies to `sizeof` and `_Alignof` as well so this should probably be extracted into an `exprNoEval` helper.

src/Parser.zig
```diff
@@ -813,6 +816,27 @@ pub const DeclSpec = struct {
     }
 };
 
+fn typeof(p: *Parser) Error!?Type {
+    if (p.eatToken(.keyword_typeof) orelse p.eatToken(.keyword_typeof1) orelse p.eatToken(.keyword_typeof2)) |_| {
```
Vexu, line 820: This would probably be better expressed with a switch on `tok_ids` and early return.

PR #12, opened 2021-09-07
Add language options module

ehaas:
Allows setting the language standard, which controls which keywords are recognized.

### DIFF
--- a/src/LangOpts.zig
+++ b/src/LangOpts.zig
@@ -23,6 +23,12 @@ const Standard = enum {
     c2x,
     /// Working Draft for ISO C2x with GNU extensions
     gnu2x,
+
+    const Invalid = error{InvalidStandard};
+
+    fn fromString(name: []const u8) Invalid!Standard{
+        return std.meta.stringToEnum(Standard, name) orelse error.InvalidStandard;
+    }
 };
 
 standard: Standard = .gnu17,
@@ -36,4 +42,8 @@ pub fn hasGNUKeywords(langopts: LangOpts) bool {
 
 pub fn hasC99Keywords(langopts: LangOpts) bool {
     return @enumToInt(langopts.standard) >= @enumToInt(Standard.c99);
+}
+
+pub fn setStandard(self: *LangOpts, name: []const u8) Standard.Invalid!void {
+    self.standard = try Standard.fromString(name);
 }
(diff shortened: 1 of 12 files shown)


src/LangOpts.zig
```diff
@@ -0,0 +1,49 @@
...
+    gnu17,
+    /// Working Draft for ISO C2x
+    c2x,
+    /// Working Draft for ISO C2x with GNU extensions
+    gnu2x,
+
+    const Invalid = error{InvalidStandard};
+
+    fn fromString(name: []const u8) Invalid!Standard {
+        return std.meta.stringToEnum(Standard, name) orelse error.InvalidStandard;
```
Vexu, line 30: This should use a `ComptimeStringMap` since there are some common aliases.

Here's the list given by clang when specifying an invalid version:
```
error: invalid value 'c87' in '-std=c87'
note: use 'c89', 'c90', or 'iso9899:1990' for 'ISO C 1990' standard
note: use 'iso9899:199409' for 'ISO C 1990 with amendment 1' standard
note: use 'gnu89' or 'gnu90' for 'ISO C 1990 with GNU extensions' standard
note: use 'c99' or 'iso9899:1999' for 'ISO C 1999' standard
note: use 'gnu99' for 'ISO C 1999 with GNU extensions' standard
note: use 'c11' or 'iso9899:2011' for 'ISO C 2011' standard
note: use 'gnu11' for 'ISO C 2011 with GNU extensions' standard
note: use 'c17', 'iso9899:2017', 'c18', or 'iso9899:2018' for 'ISO C 2017' standard
note: use 'gnu17' or 'gnu18' for 'ISO C 2017 with GNU extensions' standard
note: use 'c2x' for 'Working Draft for ISO C2x' standard
note: use 'gnu2x' for 'Working Draft for ISO C2x with GNU extensions' standard
```

src/LangOpts.zig
```diff
@@ -0,0 +1,49 @@
...
+    }
+};
+
+standard: Standard = .gnu17,
```
Vexu, line 34: Clang seems to default to `gnu17` (or at least plain `typeof` works by default), maybe we should too?

Vexu (changes requested):
We'll eventually need to also update `Compilation.generateBuiltinMacros` to take these into consideration.

src/LangOpts.zig
Vexu: Whoops misread this as `c17`.

PR #15, opened 2021-09-08
Do not evaluate typeof/sizeof/alignof arguments

ehaas:
Note: `sizeof` and `_Alignof` previously used `unExpr` to parse their expression but it seems like it should be `assignExpr` for them (like `genericSelection` uses). If they actually should be `unExpr` I can change them back.

### DIFF
--- a/src/LangOpts.zig
+++ b/src/LangOpts.zig
@@ -44,20 +44,16 @@ const Standard = enum {
     pub fn atLeast(self: Standard, other: Standard) bool {
         return @enumToInt(self) >= @enumToInt(other);
     }
+
+    pub fn isGNU(standard: Standard) bool {
+        return switch (standard) {
+            .gnu89, .gnu99, .gnu11, .gnu17, .gnu2x => true,
+            else => false,
+        };
+    }
 };
 
 standard: Standard = .gnu17,
-
-pub fn hasGNUKeywords(langopts: LangOpts) bool {
-    return switch (langopts.standard) {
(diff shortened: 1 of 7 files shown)


src/Parser.zig
```diff
@@ -3737,12 +3739,10 @@ fn unExpr(p: *Parser) Error!Result {
                     try p.expectClosing(l_paren, .r_paren);
                 } else {
                     p.tok_i = expected_paren;
-                    res = try p.unExpr();
-                    try res.expect(p);
+                    res = try p.assignExprNoEval();
```
Vexu, line 3742: `sizeof` and `_Alignof` should parse an `unExpr` if there are no parentheses.

src/LangOpts.zig
```diff
@@ -44,28 +44,25 @@ const Standard = enum {
...
-}
-
 pub fn setStandard(self: *LangOpts, name: []const u8) error{InvalidStandard}!void {
     self.standard = Standard.NameMap.get(name) orelse return error.InvalidStandard;
 }
 
 pub fn suppress(langopts: LangOpts, tag: DiagnosticTag) bool {
     return switch (tag) {
         .static_assert_missing_message => langopts.standard.atLeast(.c2x),
+        .alignof_expr => langopts.standard.isGNU(),
```
Vexu, line 65: Clang always shows this error and gcc never does, should we maybe distinguish between `gnu17` and `default` and allow gnu extensions with `default` but show errors if a gnu version was not explicitly specified?

PR #17, opened 2021-09-08
Parser: Warn on divide or remainder by zero

### LINKED ISSUE #16: Division by zero should issue a warning
Currently panics:

```c
void foo(void)
{
   (void)(0/0);
}
```

ehaas:
Fixes #16

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -3136,6 +3136,13 @@ fn assignExpr(p: *Parser) Error!Result {
         .div_assign_expr,
         .mod_assign_expr,
         => {
+            if (rhs.isZero()) {
+                switch (tag) {
+                    .div_assign_expr => try p.errStr(.division_by_zero, div.?, "division"),
+                    .mod_assign_expr => try p.errStr(.division_by_zero, mod.?, "remainder"),
+                    else => {},
+                }
+            }
             _ = try lhs_copy.adjustTypes(tok, &rhs, p, .arithmetic);
             try lhs.bin(p, tag, rhs);
             return lhs;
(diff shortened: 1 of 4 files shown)


src/Parser.zig
```diff
@@ -3494,21 +3501,28 @@ fn mulExpr(p: *Parser) Error!Result {
...
             if (mul != null) {
                 try lhs.mul(mul.?, rhs, p);
             } else if (div != null) {
-                lhs.val = switch (lhs.val) {
-                    .unsigned => |v| .{ .unsigned = v / rhs.val.unsigned },
-                    .signed => |v| .{ .signed = @divFloor(v, rhs.val.signed) },
-                    else => unreachable,
-                };
+                if (rhs.isZero()) {
+                    try p.errStr(.division_by_zero, div.?, "division");
```
Vexu, line 3508: This warning should be emitted even if `lhs` is not constant and result value should be `.unavailable`.

Vexu (changes requested):

src/Parser.zig
ehaas: Sorry for all the force pushes...I think it should be good now. Thanks for all your helpful feedback on these PRs.

Vexu: No worries, thanks for helping improve Aro.

ehaas: Just out of curiosity why is it called Aro?

Vexu: The project needed some kind of an actual name so I named it after a kind of grassland since I already named my hobby language after a kind of swamp for some reason.

PR #19, opened 2021-09-10
Parser: implement __builtin_choose_expr

ehaas:
Similar to conditional operator, except:
1. First argument must be an integer constant
2. return type matches that of chosen expression and does not undergo promotion
3. unchosen expression is not evaluated
4. can return an lvalue if the chosen expression is an lvalue

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -2764,7 +2764,7 @@ const Result = struct {
             res.ty.decayArray();
             res.ty.alignment = 0;
             try res.un(p, .array_to_pointer);
-        } else if (!p.in_macro and Tree.isLval(p.nodes.slice(), res.node)) {
+        } else if (!p.in_macro and Tree.isLval(p.nodes.slice(), p.data.items, p.value_map, res.node)) {
             res.ty.qual = .{};
             res.ty.alignment = 0;
             try res.un(p, .lval_to_rval);
@@ -3126,7 +3126,7 @@ fn assignExpr(p: *Parser) Error!Result {
     try rhs.expect(p);
     try rhs.lvalConversion(p);
 
-    if (!Tree.isLval(p.nodes.slice(), lhs.node) or lhs.ty.qual.@"const") {
+    if (!Tree.isLval(p.nodes.slice(), p.data.items, p.value_map, lhs.node) or lhs.ty.qual.@"const") {
         try p.errTok(.not_assignable, tok);
         return error.ParsingFailed;
     }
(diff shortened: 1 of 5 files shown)


ehaas:
Needing to add 2 arguments to `Tree.isLval` seems like a bit of a code smell - is there something I'm missing that would be easier?

src/Tree.zig
```diff
@@ -438,11 +440,20 @@ pub fn isLval(nodes: Node.List.Slice, node: NodeIndex) bool {
...
         },
         .paren_expr => {
             const data = nodes.items(.data)[@enumToInt(node)];
-            return isLval(nodes, data.un);
+            return isLval(nodes, extra, value_map, data.un);
+        },
+        .builtin_choose_expr => {
+            const data = nodes.items(.data)[@enumToInt(node)];
+
+            if (value_map.get(data.if3.cond)) |val| {
```
Vexu, line 452: This is a debugging fearure and should print bith branches unconditionally. The chosen branch can easily be seen by looking at the saved value of the condition.

src/Parser.zig
```diff
@@ -3809,6 +3815,43 @@ fn unExpr(p: *Parser) Error!Result {
...
+    var then_expr = if (cond.getBool()) try p.assignExpr() else try p.parseNoEval(assignExpr);
+    try then_expr.expect(p);
+
+    _ = try p.expectToken(.comma);
+
+    var else_expr = if (!cond.getBool()) try p.assignExpr() else try p.parseNoEval(assignExpr);
+    try else_expr.expect(p);
+
+    try p.expectClosing(l_paren, .r_paren);
+
```
Vexu, line 3839:
```suggestion
    try cond.saveValue(p);
```

Vexu (commented):
> Needing to add 2 arguments to `Tree.isLval` seems like a bit of a code smell - is there something I'm missing that would be easier?

The only other way of doing this I can come up with would be embedding an unfinished tree in the parser.

src/Tree.zig
ehaas: For the printer we could do that but isn't this function also used by the parser for deciding if things can be assigned to / have their address taken?

Vexu: My bad, thought this was `dump`. I really should double-check my comments when doing reviews in the morning.

src/Parser.zig
Vexu: Both my comments turned out to be wrong as `constExpr` already calls `saveValue`. Sorry about the extra work.

PR #24, opened 2021-09-12
Update predefined macros

ehaas:
Use correct value of `__STDC_VERSION__`
Add placeholder values for `__DATE__` and `__TIME__`
Update test runner to create a new compilation for each test since the predefined macros can change based on the test.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const assert = std.debug.assert;
 const mem = std.mem;
 const Allocator = mem.Allocator;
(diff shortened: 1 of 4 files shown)


ehaas:
Updated it to use the epoch API you linked, for when that gets merged.

I changed the way the test runner handles the predefined macros but I'm not entirely sure about what you mean regarding removing the magic numbers from `Compilation` - I assume you mean the `+ 2`'s but I don't know how adding a new Id `builtin` lets me get rid of those.

Vexu:
Thanks! I fixed a small off-by-one error in `__DATE__`, `month_day.day_index` is `[0,30]` we want `[1,31]`.

PR #27, opened 2021-09-13
Parser: improve typeof support

ehaas:
Add two new type specifiers: `typeof_type` and `typeof_expr`, which
are the types returned by `typeof` (depending on whether it's called
with a type or an expression)

This allows us to track the underlying type or expression that was
used.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -2045,11 +2045,10 @@ fn findScalarInitializer(p: *Parser, il: **InitList, ty: *Type) Error!bool {
             if (try p.findScalarInitializer(il, ty)) return true;
         }
         return false;
-    } else if (ty.is(.@"struct")) {
+    } else if (ty.get(.@"struct")) |struct_ty| {
         var index = il.*.list.items.len;
         if (index != 0) index = il.*.list.items[index - 1].index + 1;
 
-        const struct_ty = ty.*;
         const max_elems = struct_ty.data.record.fields.len;
         if (max_elems == 0) {
             if (p.tok_ids[p.tok_i] != .l_brace) {
@@ -2066,15 +2065,15 @@ fn findScalarInitializer(p: *Parser, il: **InitList, ty: *Type) Error!bool {
             if (try p.findScalarInitializer(il, ty)) return true;
         }
         return false;
-    } else if (ty.is(.@"union")) {
-        if (ty.data.record.fields.len == 0) {
+    } else if (ty.get(.@"union")) |union_ty| {
+        if (union_ty.data.record.fields.len == 0) {
             if (p.tok_ids[p.tok_i] != .l_brace) {
                 try p.err(.empty_aggregate_init_braces);
                 return error.ParsingFailed;
             }
             return false;
         }
-        ty.* = ty.data.record.fields[0].ty;
+        ty.* = union_ty.data.record.fields[0].ty;
         il.* = try il.*.find(p.pp.comp.gpa, 0);
         if (try p.findScalarInitializer(il, ty)) return true;
         return false;
@@ -2253,12 +2252,12 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
             },
         }
         return try p.addNode(arr_init_node);
-    } else if (init_ty.is(.@"struct")) {
+    } else if (init_ty.get(.@"struct")) |struct_ty| {
         const list_buf_top = p.list_buf.items.len;
         defer p.list_buf.items.len = list_buf_top;
 
         var init_index: usize = 0;
-        for (init_ty.data.record.fields) |f, i| {
+        for (struct_ty.data.record.fields) |f, i| {
             if (init_index < il.list.items.len and il.list.items[init_index].index == i) {
                 const item = try p.convertInitList(il.list.items[init_index].list, f.ty);
                 try p.list_buf.append(item);
@@ -2285,13 +2284,13 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
             },
         }
         return try p.addNode(struct_init_node);
-    } else if (init_ty.is(.@"union")) {
+    } else if (init_ty.get(.@"union")) |union_ty| {
         var union_init_node: Tree.Node = .{
             .tag = .union_init_expr,
             .ty = init_ty,
             .data = .{ .union_init = .{ .field_index = 0, .node = .none } },
         };
-        if (init_ty.data.record.fields.len == 0) {
+        if (union_ty.data.record.fields.len == 0) {
             // do nothing for empty unions
         } else if (il.list.items.len == 0) {
             union_init_node.data.union_init.node = try p.addNode(.{
@@ -2301,7 +2300,7 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
             });
         } else {
             const init = il.list.items[0];
-            const field_ty = init_ty.data.record.fields[init.index].ty;
+            const field_ty = union_ty.data.record.fields[init.index].ty;
             union_init_node.data.union_init = .{
                 .field_index = @truncate(u32, init.index),
                 .node = try p.convertInitList(init.list, field_ty),
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -986,14 +986,17 @@ fn initDeclarator(p: *Parser, decl_spec: *DeclSpec) Error!?InitDeclarator {
 
         var init_list_expr = try p.initializer(init_d.d.ty);
         init_d.initializer = init_list_expr.node;
-        if (init_d.d.ty.get(.incomplete_array)) |incomplete_arr_type| {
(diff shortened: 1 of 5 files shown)


ehaas:
Curious what you think of this approach.

My biggest concern is that the various `typeof_` types could be a bit of a footgun if they're not explicitly checked any time types are dealt with.

There are a bunch of places in `Parser` where a type specifier is directly checked for equality; these would all need to be updated to use something like `ty.is(specifier)` in order to check inside the typeof types.

src/Parser.zig
```diff
@@ -850,12 +850,36 @@ fn typeof(p: *Parser) Error!?Type {
...
+            .data = .{ .sub_type = typeof_ty },
+            .specifier = .typeof_type,
+        };
     }
     const typeof_expr = try p.parseNoEval(expr);
     try typeof_expr.expect(p);
     try p.expectClosing(l_paren, .r_paren);
-    return typeof_expr.ty;
+
+    const inner = try p.arena.create(Type.VLA);
```
Vexu, line 869: It would probably be appropriate to rename `Type.VLA` to `Type.Expr` or something.

src/Type.zig
```diff
@@ -1214,6 +1326,16 @@ pub fn dump(ty: Type, w: anytype) @TypeOf(w).Error!void {
             try w.writeAll("[<expr>]");
             try ty.data.vla.elem.dump(w);
         },
+        .typeof_type, .decayed_typeof_type => {
+            try w.writeAll("typeof(");
+            try ty.data.sub_type.dump(w);
+            try w.writeAll(")");
+        },
+        .typeof_expr, .decayed_typeof_expr => {
+            try w.writeAll("typeof(<expr>: ");
```
Vexu, line 1335: Someday I'll go through the effort of adding a `token` field to `Node` and then this can be changed to pretty print the expression like clang.

Vexu (approved):
> Curious what you think of this approach.
> 
> My biggest concern is that the various `typeof_` types could be a bit of a footgun if they're not explicitly checked any time types are dealt with.
> 
> There are a bunch of places in `Parser` where a type specifier is directly checked for equality; these would all need to be updated to use something like `ty.is(specifier)` in order to check inside the typeof types.

Those places would've needed to be updated sooner or later anyways for attributed types. That is unless we find some other, better way to represent them.

src/Parser.zig
ehaas: Renamed it to `Type.Expr` and renamed the members to be (I think) more consistent. It's in its own commit so should be easy to revert or change if you don't like it.

src/Parser.zig
```diff
@@ -959,7 +983,7 @@ fn initDeclarator(p: *Parser, decl_spec: *DeclSpec) Error!?InitDeclarator {
 
         var init_list_expr = try p.initializer(init_d.d.ty);
         init_d.initializer = init_list_expr.node;
-        if (init_d.d.ty.specifier == .incomplete_array) {
+        if (init_d.d.ty.is(.incomplete_array)) {
```
Vexu, line 986: This will lead to invalid usage of `ty.data`.

src/Parser.zig
```diff
@@ -2039,7 +2063,7 @@ fn findScalarInitializer(p: *Parser, il: **InitList, ty: *Type) Error!bool {
             if (try p.findScalarInitializer(il, ty)) return true;
         }
         return false;
-    } else if (ty.specifier == .@"union") {
+    } else if (ty.is(.@"union")) {
```
Vexu, line 2066: Also invalid.

src/Parser.zig
```diff
@@ -2203,7 +2227,7 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
             .data = .{ .bin = .{ .lhs = .none, .rhs = .none } },
         };
 
-        if (init_ty.specifier == .incomplete_array) {
+        if (init_ty.is(.incomplete_array)) {
```
Vexu, line 2230: `arr_init_node == init_ty` so the below assignments should happen on a duped type (at least if init_ty is not a meta type).

src/Parser.zig
```diff
@@ -2226,7 +2250,7 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
             },
         }
         return try p.addNode(arr_init_node);
-    } else if (init_ty.specifier == .@"struct") {
+    } else if (init_ty.is(.@"struct")) {
```
Vexu, line 2253: Will also lead to invalid use.

src/Parser.zig
```diff
@@ -2258,7 +2282,7 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
             },
         }
         return try p.addNode(struct_init_node);
-    } else if (init_ty.specifier == .@"union") {
+    } else if (init_ty.is(.@"union")) {
```
Vexu, line 2285: Same.

src/Parser.zig
```diff
@@ -1788,7 +1812,7 @@ fn paramDecls(p: *Parser) Error!?[]Type.Func.Param {
         } else if (param_ty.isArray()) {
             // params declared as arrays are converted to pointers
             param_ty.decayArray();
-        } else if (param_ty.specifier == .void) {
+        } else if (param_ty.is(.void)) {
```
Vexu, line 1815: The use of `qual.any()` below is invalid.

src/Parser.zig
```diff
@@ -2070,8 +2094,8 @@ fn coerceArrayInit(p: *Parser, item: *Result, tok: TokenIndex, target: Type) !bo
         return true; // do not do further coercion
     }
 
-    if (target.specifier == .array) {
-        assert(item.ty.specifier == .array);
+    if (target.is(.array)) {
+        assert(item.ty.is(.array));
```
Vexu, line 2098: The use of `target.data.array.len` below should be replaced with `arrayLen` and the assert should be left as is since `item` is supposed to be a string literal.

Vexu (changes requested):
I propose adding a `fn get(ty: *const Type, specifier: Specifier) ?*const Type` function to make correct usage easier.

ehaas:
Good catch, that was pretty careless on my part. I'll add the `.get` method you suggested and try to write some tests to cover the various places where the wrong union field would be used.

ehaas:
Couple questions:

1) Should `arrayLen` be defined for `static_array`? The original code didn't try to get the length of any static arrays but maybe it should be defined?

2) I believe some of the qualifier code may be incorrect with initializers + incomplete arrays and also with conditional expressions - pulling the qualifier from the unwrapped type. Do we need a method that merges all the qualifiers from the child types of typeof-types?

Vexu:
> Couple questions:
> 
>     1. Should `arrayLen` be defined for `static_array`? The original code didn't try to get the length of any static arrays but maybe it should be defined?

It doesn't matter for initializers as the `init_ty` can never be a static array but if you do include them then the function can also be used in `checkArrayBounds`.
 
>     2. I believe some of the qualifier code may be incorrect with initializers + incomplete arrays and also with conditional expressions - pulling the qualifier from the unwrapped type. Do we need a method that merges all the qualifiers from the child types of typeof-types?

Not sure I understand, can you give examples?

Also you seem to have accidentally commited a binary, could you rebase that away?

ehaas:
> Not sure I understand, can you give examples?

Specifically thinking of line 2992 in `Parser.zig` `adjusted_elem_ty.qual = a_elem.qual.mergeCV(b_elem.qual);` and line 995 in `Parser.zig` `.qual = incomplete_arr_type.qual,`

ehaas:
Any idea why the double-free would only happen on windows? I can dig out a windows box for debugging but it would be nice if it happened on mac or linux too :)

Vexu:
> Specifically thinking of line 2992 in `Parser.zig` `adjusted_elem_ty.qual = a_elem.qual.mergeCV(b_elem.qual);` and line 995 in `Parser.zig` `.qual = incomplete_arr_type.qual,`

995 is definitely wrong, for example:
```c
void foo(void) {
    const typeof(int[]) arr = {1, 2, 3};
    // var: '[3]int' should be [3]const int
    //   name: arr
}
```
I don't see anything wrong with 2992 though.

> Any idea why the double-free would only happen on windows?

No idea, I wouldn't be surprised if it was a bug in Zig's std.

ehaas:
I'll get my Windows machine up to date and see if I can at least bisect it to one of my commits. Will also double check the qualifiers.

ehaas:
Ok, I updated things to make the qualifiers work better now, although I'm not super happy with the `canonicalize` function - the `preserve_quals` option is only used in the `elemType` function, to pick up all the qualifiers that were added by various levels of `typeof()`; and it's only used to get the qualifiers and not the type itself. So it seems like it would possibly make sense to have a separate `getQuals` function for doing that, but most of the logic would be similar to that in the `canonicalize` function.

I get the sense that there is something I'm missing that would simplify this; possibly in the data representation.

Vexu:
Thanks! That seems like something that could be improved but it's probably not worth investigating before attributed types are implemented.

PR #29, opened 2021-09-20
Fix recursive macro func substitution

doppioandante:
The current approach of having a map of the previously expanded macros
comes close, but isn't quite correct as it is too conservative.
The `expansion_log` was changed to a simple `expansion_stack`, which
is searched linearly to check for already expanded macros.

I have not written much zig yet, so suggestions for style or logic are welcome.

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -72,9 +72,12 @@ generated: std.ArrayList(u8),
 token_buf: RawTokenList,
 char_buf: std.ArrayList(u8),
 pragma_once: std.AutoHashMap(Source.Id, void),
-// It is safe to have pointers to entries of defines since it
-// cannot be modified while we are expanding a macro.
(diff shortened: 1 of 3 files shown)


Vexu (approved):
Thanks for the fix! Style and logic look fine but it does need to be formatted with `zig fmt` to pass CI.
I'd also prefer if you prefixed commits with the area of the compiler they touch.

doppioandante:
I suppose the windows failure is unrelated?

Vexu:
Strange, that's the same issue as with the other open pr. I wasn't able to reproduce it locally on Windows 10, I'll have to ask for help from the Windows users on Discord.

Vexu:
It seems like the failure is triggered by adding a 46th test case to `test/cases`, next up is figuring out why.

doppioandante:
Now that the 46th test bug has been fixed, I'm closing this as #33 is more general.

PR #33, opened 2021-09-26
Fix preprocessor expansion strategy

### LINKED ISSUE #31: Preprocessor misses some expansion opportunities
```c
#define CREATE_CALL F
#define F(x) (x + 2)

CREATE_CALL(2)

#define EXPECTED_TOKENS (2 + 2)
```
This test case fails as arocc preprocessor generates `F ( 2 )`

doppioandante:
Fixes #31 
Fixes #29 
Supersedes #29 
Still WIP, I need to reintegrate the logic needed to handle __VA_ARGS__ and some other things, but overall it works.

I will tidy up the commits when done.


src/Preprocessor.zig
```diff
@@ -510,14 +513,400 @@ fn skipToNl(tokenizer: *Tokenizer) void {
...
+    var variable_arguments = ExpandBuf.init(pp.comp.gpa);
+    defer variable_arguments.deinit();
+
+    if (func_macro.var_args) {
+        var i: usize = func_macro.params.len;
+        while (i < expanded_args.items.len) : (i += 1) {
+            try variable_arguments.appendSlice(args.items[i]);
+            try expanded_variable_arguments.appendSlice(expanded_args.items[i]);
+            if (i != expanded_args.items.len-1 ) {
+                // TODO: use generated buffer?
```
Vexu, line 555: Should not be needed as `.comma` will always map to `","`.

src/Preprocessor.zig
```diff
@@ -510,14 +513,400 @@ fn skipToNl(tokenizer: *Tokenizer) void {
...
+}
+
+fn shouldExpand(tok: Token, macro: *Macro) bool {
+    const macro_loc = switch (macro.*) {
+        .simple => |smacro| smacro.loc,
+        .func => |smacro| smacro.loc,
+        else => unreachable
+    };
+    var maybe_loc = tok.loc.next;
+    while (maybe_loc) |loc| {
```
Vexu, line 676: A future improvement will be to convert the source location mechanism into using contiguous arrays instead of linked lists now that they are being iterated upon a lot more.

src/Preprocessor.zig
```diff
@@ -510,14 +513,400 @@ fn skipToNl(tokenizer: *Tokenizer) void {
...
+                            var args_count = @intCast(u32, args.items.len);
+                            // if the macro has zero arguments g() args_count is still 1
+                            if (args_count == 1 and func_macro.params.len == 0) args_count = 0;
+
+                            // Validate argument count.
+                            const extra = Diagnostics.Message.Extra{ .arguments = .{ .expected = @intCast(u32, func_macro.params.len), .actual = args_count } };
+                            var validated = true;
+                            if (func_macro.var_args and args_count < func_macro.params.len) {
+                                try pp.comp.diag.add(.{ .tag = .expected_at_least_arguments, .loc = buf.items[idx].loc, .extra = extra });
+                                validated = false;
```
Vexu, line 811:
```suggestion
                                idx += 1;
                                continue;
```
We're there some issues here earlier?

src/Preprocessor.zig
```diff
@@ -893,6 +1281,45 @@ pub fn expandedSlice(pp: *Preprocessor, tok: Token) []const u8 {
     return tmp_tokenizer.buf[res.start..res.end];
 }
 
+fn pasteTokens2(pp: *Preprocessor, lhs: []const Token, rhs: []const Token) Error![]const Token {
```
Vexu, line 1284: This doesn't seem at all correct, as far as I understand pasting is supposed to form a single token without doing any expansion.

src/Preprocessor.zig
```diff
@@ -510,14 +513,400 @@ fn skipToNl(tokenizer: *Tokenizer) void {
...
+    const name_tok = buf.items[start_idx.*];
+    const l_paren_tok = try nextBufToken(tokenizer, buf, start_idx, end_idx, extend_buf);
+    if (l_paren_tok.id != .l_paren) {
+        // Not a macro function call, go over normal identifier.
+        return null;
+    }
+
+    // collect the arguments.
+    var parens: u32 = 0;
+    var args = MacroArguments.init(pp.comp.gpa);
```
Vexu, line 717: Missing `errdefer`.

src/Preprocessor.zig
```diff
@@ -510,14 +513,400 @@ fn skipToNl(tokenizer: *Tokenizer) void {
...
+                        do_rescan = true;
+                    },
+                    .func => |func_macro| {
+                        var macro_scan_idx = idx;
+                        if (try pp.collectMacroFuncArguments(tokenizer, buf, &macro_scan_idx, &moving_end_idx, extend_buf)) |args| {
+                            // TODO: bug?
+                            //defer for (args.items) |item| {
+                            //    pp.comp.gpa.free(item);
+                            //};
+                            //defer args.deinit();
```
Vexu, line 801: Putting the for loop inside a block in the defer seems to fix this for me:
```zig
defer {
    for (args.items) |item| {
        pp.comp.gpa.free(item);
    }
    args.deinit();
}
```

Vexu (commented):
So far this seems mostly fine, I tried answering most of the questions in the TODOs.
I'd be nice if more allocations could be avoided but correctness comes first.

src/Preprocessor.zig
doppioandante: Yeah, it's wrong, I need to revisit this.

test/cases/unspecified expansion.c
```diff
@@ -0,0 +1,6 @@
+#define EXPECTED_TOKENS 2*f(9)
```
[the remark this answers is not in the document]
doppioandante: This is actually undefined behavior (in my C draft, 6.10.3.4, the example snippet) and it can expand either as `2*f(9)` or as `2*9*g`. It might make sense to find why clang and gcc do differently, but from the standard viewpoint this is fine. I added this test to set in stone the current preprocessor behavior.

src/Preprocessor.zig
```diff
@@ -19,16 +19,16 @@ const max_include_depth = 200;
 const Macro = union(enum) {
     /// #define FOO
     empty,
```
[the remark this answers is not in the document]
doppioandante: Yeah, removing `empty` is definitely on my todo list.

doppioandante:
Small status update, don't review yet because the commits are a mess right now, It seems that the preprocessor is working ok. I've tested it with [libpp](https://github.com/mcinglis/libpp) and it passes all its tests. I also tested it with [cloak](https://github.com/pfultz2/Cloak), but that didn't work completely, I suspect it relies on the undefined behaviour that arocc does differently respect to gcc/clang.

I will test it with more libraries from [awesome-c-preprocessor](https://github.com/Hirrolot/awesome-c-preprocessor)

A note about commit cf776f3 . That commit deals with the following:
```c
#define h(s) 30
#define H h
#define SS (s)

H SS // should this be h(s) or 30?
```
gcc/clang expand as h(s), while arocc before cf776f3 expands to 30, after it does like gcc. What gcc does is reasonable, but I'm not sure it is strictly mandated by the standard

src/Preprocessor.zig
doppioandante: Yeah, I will open a issue/pr after this one is done

src/Preprocessor.zig
doppioandante: I'm not getting what you are suggesting, if validated = true the next if would be entered, which must expand the macro func, but there are not enough arguments to do so.. am I missing something?

doppioandante:
@Vexu I'm stopping here with the fixes and enhancements, this is getting a bit out of hand already.
Can you please tell me if you prefer to review this after I have rebased/reordered the commits or if you want to review now? Being this big, it is a bit easier to be able to see the timeline of the changes, but it's the same for me.

After the rebasing, I will add documentation to all the parts I touched, and I will do following development in future PR:
- remove `.empty` from the Macro enum and rework the func/simple distinction
- make the location linked list a vector
- FIx the commented parts of test `test/cases/standard-concatenation-strings-example.c`
A bit of explanation on the last one: this is what fails now
```c
#define str(x) #x
str(f(x)+1)
str(f(x) + 1)
```
This should expand to "str(f(x)+1)" and "str(f(x) + 1)" respectively. The previous implementation always did the second, the current always does the former.
As far as I can tell, the current preprocessor completely deletes simple spaces between token, is that correct? Unfortunately I think those must be kept track of, what do you think about this?

src/Preprocessor.zig
```diff
@@ -715,12 +715,20 @@ fn collectMacroFuncArguments(pp: *Preprocessor, tokenizer: *Tokenizer, buf: *Exp
...
-    const l_paren_tok = try nextBufToken(tokenizer, buf, start_idx, end_idx, extend_buf);
-    if (l_paren_tok.id != .l_paren) {
-        // Not a macro function call, go over normal identifier, rewind
-        tokenizer.index = initial_tokenizer_index;
-        end_idx.* = old_end;
-        return null;
+    var lparen_found = false;
+    while (!lparen_found) {
+        const l_paren_tok = try nextBufToken(tokenizer, buf, start_idx, end_idx, extend_buf);
+        if (l_paren_tok.id == .nl) {
```
doppioandante, line 721: I don't know if this suffices, but it works and I called it a day lol

Vexu:
> Can you please tell me if you prefer to review this after I have rebased/reordered the commits or if you want to review now? 

I can review it now, should make squashing any fixups easier too.

> Unfortunately I think those must be kept track of, what do you think about this?

Yes that seems to be the case, it should also make properly implementing the `-E` command easier and shouldn't add too much memory usage as we don't have to pass the whitespace tokens out of the preprocessor and they don't need proper source locations.

src/Preprocessor.zig
```diff
@@ -1232,3 +1303,45 @@ fn printInBetween(slice: []const u8, w: anytype) !void {
...
+        } else if (std.mem.eql(u8, slice, "")) {
+            std.debug.print("({}) ", .{buf[i].id});
+        }else {
+            std.debug.print("{s} ", .{slice});
+        }
+    }
+    std.debug.print("[{} tokens]\n", .{buf.len});
+}
+
+fn addTestSource(pp: *Preprocessor, path: []const u8, content: []const u8) !Source {
```
doppioandante, line 1333: Oops, I had used this function for unit testing, it is not needed anymore, so I'll remove it.

src/Preprocessor.zig
```diff
@@ -731,136 +655,278 @@ fn expandFunc(pp: *Preprocessor, source: *ExpandBuf, start_index: *usize, macro:
...
-        if (commas_seen == index) for (args[i..]) |a_2, j| {
-            if (parens == 0 and a_2.id == .comma) {
-                return args[i..][0..j];
+fn expandMacroExhaustive(pp: *Preprocessor, tokenizer: *Tokenizer, buf: *ExpandBuf, start_idx: usize, end_idx: usize, extend_buf: bool) Error!void {
+    var moving_end_idx = end_idx;
+    var advance_index: usize = 0;
+    // rescan loop
+    var do_rescan = true;
+    while (do_rescan) {
+        do_rescan = false;
```
Vexu, line 777: This logic looks like it could be simplified but I can't reason about it right now.

A slight improvement which allows simplifying other parts of the function would be to move the condition from the end into a continue expression:
```suggestion
    var do_rescan = true;
    while (do_rescan) : {
        // last phase before rescan: remove placeholder tokens
        // NOTE: only do this if there were expansion (i.e. do_rescan is true)
        if (do_rescan) {
            var i: usize = start_idx;
            while (i < buf.items.len) {
                const tok = &buf.items[i];
                switch (tok.id) {
                    .empty_arg => {
                        _ = buf.orderedRemove(i);
                        moving_end_idx -= 1;
                    },
                    else => i += 1,
                }
            }
        }
    } {
        do_rescan = false;
```
You'll probably want to double check these changes for correctness before adding them.

src/Preprocessor.zig
```diff
@@ -731,136 +655,278 @@ fn expandFunc(pp: *Preprocessor, source: *ExpandBuf, start_index: *usize, macro:
...
+        // expansion loop
+        var idx: usize = start_idx + advance_index;
+        //std.debug.print("Scanning ", .{});
+        //try pp.debugTokenBuf(buf.items[start_idx+advance_index .. moving_end_idx]);
+
+        while (idx < moving_end_idx) {
+            const macro_entry = pp.defines.getPtr(pp.expandedSlice(buf.items[idx]));
+            if (macro_entry != null and shouldExpand(buf.items[idx], macro_entry.?)) {
```
Vexu, line 785:
```suggestion
            if (macro_entry == null or !shouldExpand(buf.items[idx], macro_entry.?)) {
                idx += 1;
                continue;
            }
```

src/Preprocessor.zig
```diff
@@ -731,136 +655,278 @@ fn expandFunc(pp: *Preprocessor, source: *ExpandBuf, start_index: *usize, macro:
...
+
+                        try buf.replaceRange(idx, 1, res.items);
+                        idx += res.items.len;
+                        moving_end_idx += (res.items.len - 1);
+                        do_rescan = true;
+                    },
+                    .func => |func_macro| {
+                        var macro_scan_idx = idx;
+                        // to be saved in case this doesn't turn out to be a call
+                        if (try pp.collectMacroFuncArguments(tokenizer, buf, &macro_scan_idx, &moving_end_idx, extend_buf)) |args| {
```
Vexu, line 811:
```suggestion
                        const args = try pp.collectMacroFuncArguments(tokenizer, buf, &macro_scan_idx, &moving_end_idx, extend_buf) orelse {
                            idx += 1;
                            continue;
                        };
```

src/Preprocessor.zig
Vexu: The `validated` flag is complicating the control flow. In both cases if `validated == false` the equivalent of `idx +=1; continue;` will be executed. Removing it simplifies the control flow and reduces the nesting depth.

I originally didn't notice the `do_rescan` at the end of the loop but that should now be accounted for by an earlier comment.

Vexu (approved):
This change looks good and removes some of the ugliest parts of the entire codebase.

I didn't make notes about all the debug comments and prints since you'll probably remove them as part of the rebasing.

src/Preprocessor.zig
```diff
@@ -731,136 +652,256 @@ fn expandFunc(pp: *Preprocessor, source: *ExpandBuf, start_index: *usize, macro:
...
+                        if (buf.items[idx].loc.next) |ln| {
+                            try pp.markExpandedFrom(tok, ln.*);
+                        }
+                        try pp.markExpandedFrom(tok, func_macro.loc);
+                    }
+
+                    try buf.replaceRange(idx, macro_scan_idx-idx+1, res.items);
+                    // TODO: moving_end_idx += res.items.len - (macro_scan_idx-idx+1)
+                    // doesn't work when the RHS is negative (unsigned!)
+                    moving_end_idx = moving_end_idx + res.items.len - (macro_scan_idx-idx+1);
```
doppioandante, line 877: Any suggestion about this? It was moving_end_idx -= res.items.len - (macro_scan_idx-idx+1);
But the rhs can become negative, which is harmless but zig doesn't like it. It works right now, but I don't know if this is the zig way

src/Preprocessor.zig
```diff
@@ -70,11 +70,7 @@ defines: DefineMap,
 tokens: Token.List = .{},
 generated: std.ArrayList(u8),
 token_buf: RawTokenList,
-char_buf: std.ArrayList(u8),
```
doppioandante, line 73: this was only used during stringification, so I moved it there locally, is that okay?

doppioandante:
Rebased and applied zig fmt. I originally wanted to also add some docs to this PR, but maybe I'll leave that for another one if that's okay for you?
I left another couple of comments, but other than that I think it's good to go.

src/Preprocessor.zig
Vexu: The reason it was in the `PreProcessor` was so that the buffer could be reused between stringifications redusing redundant allocations.

PR #34, opened 2021-09-29
Readme: use sh-session highlighting instead of shell

HappyFacade:


Vexu:
Neat, didn't know this existed. Thanks!

PR #35, opened 2021-10-01
Parser: Rudimentary attribute parsing

ehaas:
Putting this up as a draft to see what you think; right now it just consumes attributes (in some but not all places where they are allowed) but doesn't do anything with them.

Remaining work includes but is not limited to:

- [ ] Implement `__has_attribute` preprocessor function
- [ ] Verify arg count and types for each attribute
- [ ] Strategy for attributed statements, var decls, function decls, types, etc
- [ ] Actually do something with the attributes


Vexu:
I don't think attributes should emit nodes but instead be more like types. All the values should be constants so it would be easier to consume them without all the hassle that comes with nodes, If some attribute does require a node we can just have it work like `Type.Expr`.

ehaas:
Any thoughts on how to handle parameters? As far as I can tell they can either be integer constant expressions (e.g. `aligned(2+2)` is valid); identifiers (e.g. `format(printf, 2, 3)`), or strings (`section("mysegment,mysection"))`. Tagged union would be easier to implement I think but use more space; but if we know what kind of parameters each attribute can take then we could probably get away with regular unions.

src/Attribute.zig
```diff
@@ -0,0 +1,279 @@
+const std = @import("std");
+const mem = std.mem;
+const Tree = @import("Tree.zig");
+const NodeIndex = Tree.NodeIndex;
+
+const Attribute = @This();
+
+/// Unknown attributes have null tag
+tag: ?Tag,
```
Vexu, line 9: I don't think anything should be emitted for unknown attributes, it's not like we can do anything meaningful with them.

Vexu (commented):
This is looking pretty mergeable already, all of those check boxes can be done separately including using a more appropriate data structure over AST nodes.

src/Attribute.zig
ehaas: Updated it to not emit anything for those. I also switched it to storing the `TokenIndex` instead of the parsed tag because that made it easier to consume unknown/ignored attributes.

ehaas:
Looks like windows CI is complaining again...will see if I can reproduce it. The error message in the CI runner isn't very descriptive.

ehaas:
Looks like it might have something to do with windows newlines and the test runner? I get this when running `zig build test` in windows:

```
All 9 tests passed.
Test [3/50] alignment... 1 test skipped
Test [8/50] binary expressions... 3 test skipped
Test [20/50] generic... 1 test skipped
Test [23/50] hello world... 1 test skipped
Test [25/50] initializers... 2 test skipped
Test [26/50] invalid types... 5 test skipped
Test [34/50] redefinitions... 1 test skipped
Test [50/50] __has_attribute...
======= expected to find error =======
__has_attribute.c:3:7: error: attribute exists

=== but output does not contain it ===
C:\Users\Evan\source\arocc\test\cases\__has_attribute.c:3:7: error: attribute exists
      #error attribute exists
      ^
1 error generated.


49 passed; 1 failed.
```

Vexu:
Here's the culprit: https://github.com/Vexu/arocc/blob/39a7528cdd3fc7974701876520c23ab963c35334/src/Preprocessor.zig#L134

I fixed a similar issue for `#pragma` but didn't think to check `#error` too.

ehaas:
Once this PR is reviewed/merged I can create issues for all the extant tasks I know about

ehaas:
Also there were some false starts along the way so let me know if you want me to squash it all into one commit

Vexu:
> let me know if you want me to squash it all into one commit

That would be nice if you don't mind.

PR #46, opened 2021-10-12
Preprocessor: make __has_attribute work with simple #define

ehaas:
Note: does not yet handle the case of builtin feature macros used in user-defined function-like macros.

It seems like we'll need some logic in `expandFuncMacro` to handle that, but I'm not sure of the best way to do it while also not repeating the logic in `evalBuiltin`

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -925,6 +925,7 @@ fn expandMacro(pp: *Preprocessor, tokenizer: *Tokenizer, raw: RawToken) Error!vo
 
 // mark that this token has been expanded from `loc`
 fn markExpandedFrom(pp: *Preprocessor, tok: *Token, loc: Source.Location) !void {
+    if (loc.id == .generated) return;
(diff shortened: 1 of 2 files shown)


ehaas:
I think you're right that it needs to b e unified with the macro expansion logic, since something like this is valid:
```c
#define foo(x,y) __has_attribute(x##y)
#if foo(us,ed)
#error "it is defined"
#endif
```

ehaas:
Updated. We currently have a `Compilation.generateBuiltinMacros` which generates textual macros, as well as `Preprocessor.addBuiltinMacros` which generates symbolic ones. Should those two be combined somehow?

Vexu:
> Should those two be combined somehow?

Ideally yes but since they work on different levels it would probably end up messier. A cleaner solution would probably be to either rename `Compilation.generateBuiltinMacros` to `generateBuiltinMacroSource` or something, or making `Preprocessor.addBuiltinMacros` a part of `Preprocessor.init`.

PR #49, opened 2021-10-14
Fix bug causing warning when assignment increases qualifiers

### LINKED ISSUE #28: Spurious warning when assignment increases qualifiers
I believe this is the opposite of #25

```C
void foo(void) {
   char *bar = 0;
   const char *baz = bar;
}
```

```
/Users/ehaas/source/c/test.c:3:22: warning: initializing 'const char *' from incompatible type 'char *' discards qualifiers
   const char *baz = bar;
                     ^
1 warning generated.
fn_def: 'fn () void'
 name: foo
 body:
  compound_stmt: 'void'
    var: '*char'
     name: bar
     init:
      null_to_pointer: '*char'
        int_literal: 'int'
         value: 0

    var: '*const char'
     name: baz
     init:
      lval_to_rval: '*char'
        decl_ref_expr: '*char' lvalue
         name: bar

    implicit_return: 'void'
```

iddev5:
Adds a new method ``Qualifiers.hasQuals(a, b)`` which checks if a has all qualifiers of b.
This method is used to check if in a pointer assignment, lhs's base element type has all the qualifiers of rhs's base element type. If not, then only create the warning.

Closes #28

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -2346,7 +2346,7 @@ fn coerceInit(p: *Parser, item: *Result, tok: TokenIndex, target: Type) !void {
                 try p.errStr(.incompatible_ptr_init, tok, try p.typePairStrExtra(target, e_msg, item.ty));
                 try item.ptrCast(p, unqual_ty);
             } else if (!unqual_ty.eql(item.ty, true)) {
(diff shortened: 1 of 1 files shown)


Vexu:
Thanks!

PR #52, opened 2021-10-18
Implement __has_warning

ehaas:
Additionally refactor Tokenizer and Preprocessor a bit.


doppioandante:
I like the simplification in the preprocessor respect to the previous code

PR #56, opened 2021-10-19
Preprocessor: move `addBuiltinMacros` into `init`

ehaas:
Discussed here https://github.com/Vexu/arocc/pull/46#issuecomment-945182261

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -65,8 +65,8 @@ char_buf: std.ArrayList(u8),
 pragma_once: std.AutoHashMap(Source.Id, void),
 include_depth: u8 = 0,
 
-pub fn init(comp: *Compilation) Preprocessor {
-    return .{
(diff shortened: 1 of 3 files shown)


ehaas:
Since they're just regular macros from the preprocessor's perspective, the `-U` option (which corresponds to just using `#undef`) can be used to undefine them. Is there a dedicated option in gcc/clang for removing specific builtin feature check macros? I did a quick check here https://clang.llvm.org/docs/ClangCommandLineReference.html and didn't see anything

ehaas:
I can see the value in that. I'm agnostic about the API; as it stands you can do `preprocessor.defines.remove(macro_name)` to remove the named builtin macro before running it against a source file.

Vexu:
I don't like having to allocate things like this and I don't think having two similarly named function is reason enough to justify it especially when things are still quite volatile so I'm closing this in favor of fixing it later when untangling the dependency loops between the various components.

PR #72, opened 2021-10-28
Parser: Check for function redefinition for same signature

iddev5:
Check for function redefinitions having exactly same signature, i,e redefinition of exact same function, which were previously missing from the parser. Test case included.

```c
int foo(int a) { 
    return 10; 
}

int foo(int a) { 
    return a; 
}
```


Vexu:
Thanks!

PR #73, opened 2021-10-30
Diagnostics: show the respective warning and error options besides warning and error messages

iddev5:
Right now, warnings and errors are showed as follows:
```sh-session
warning: implicit declaration of function '_Static_asset' is invalid in C99
error: multi-character character constant
```

This PR adds the corresponding CLI option to explicitly enable the error or warning setting besides the messages as follows:
```sh-session
warning: implicit declaration of function '_Static_asset' is invalid in C99 [-Wimplicit-function-declaration]
error: multi-character character constant [-Werror=multichar]
```

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -663,6 +663,15 @@ pub fn renderExtra(comp: *Compilation, m: anytype) void {
             .zero_width_named_field => m.write("named bit-field has zero width"),
             .bitfield_too_big => m.write("width of bit-field exceeds width of its type"),
         }
+
+        if (comp.diag.tagOption(msg.tag)) |opt| {
+            if (msg.kind == .@"error") {
+                m.print(" [-Werror={s}]", .{opt});
+            } else {
+                m.print(" [-W{s}]", .{opt});
+            }
+        }
+
         m.end(lcs);
 
         if (msg.loc.id != .unused) {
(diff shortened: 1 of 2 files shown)


test/runner.zig
```diff
@@ -279,7 +279,7 @@ pub fn main() !void {
                 const expected_error = buf.items;
 
                 const index = std.mem.indexOf(u8, m.buf.items, expected_error);
-                if (index == null or m.buf.items[index.? + expected_error.len] != '\n') {
+                if (index == null) {
```
iddev5, line 282: I am not too sure about this modifications here so please let me know. But basically the two options are: 1) Change all existing EXPECT_ERRORS (or) 2) Don't check till eof, but I am not too sure for what reasons this limitation exists.

src/Diagnostics.zig
```diff
@@ -694,6 +703,61 @@ pub fn renderExtra(comp: *Compilation, m: anytype) void {
...
+        .pragma_warning_message => "#pragma-messages",
+        .newline_eof => "newline-eof",
+        .empty_translation_unit => "empty-translation-unit",
+        else => null,
+    };
+}
+
 pub const Kind = enum { @"fatal error", @"error", note, warning, off };
 
 fn tagKind(diag: *Diagnostics, tag: Tag) Kind {
```
iddev5, line 763: Technically, tagKind can be modified as such:
```zig
var kind: Kind = undefined;

if (diag.tagOption(tag)) |opt| {
    inline for (std.meta.fields(Options)) |field| {
        if (mem.eql(u8, field.name, opt)) {
            kind = @field(diag.options, field.name);
        }
    }
} else kind = @as(Kind, switch (tag) {
```
but, I am not getting a good feeling about it as I think it can introduce bugs. But I would make this or similar changes if requested in review.

Vexu:
I don't like how this is implemented but I haven't been able to come up with a better way yet so I'll merge this and open an issue to track improvements to the diagnostics system.

iddev5:
I agree

PR #83, opened 2021-11-07
Fix pragma length confusion in parser

ehaas:


Vexu:
Seems like the problem is caused by `skipTo` and `next*` not handling pragmas while they probably should.

Is there a reason why we can't just include a terminating newline or eof token? This `pragma_lens` solution is starting to look over-engineered.

ehaas:
I think a newline (or special end-pragma if necessary) token would do it. I'll look into that + make sure `skipTo` and `next*` functions handle pragmas properly.

ehaas:
I'll rebase this after the mutex update is merged

PR #94, opened 2021-11-13
Add reference to IRC channel

doppioandante:
Hi, I registered a IRC channel on libera just in case, feel free to join. I'm the channel op currently but I can cede or share the role with anyone interested.


Vexu:
I'm not very active on IRC/Discord/etc so I'd miss most if any discussion happening. Would GitHub discussions work for what you were planning to use this for?

doppioandante:
I'm fine with github discussions, I'll close this.

PR #104, opened 2021-11-17
Parser: stricter validation of preprocessor expressions

### LINKED ISSUE #101: Unary operator in `#if` causes index out of bounds in Tree.isLvalExtra
The following C code crashes the preprocessor:

```c
#if c--
#endif
```
clang reports `error: token is not a valid binary operator in a preprocessor subexpression`, pointing to the `--`

Partial trace:
```
/Users/ehaas/source/arocc/src/Tree.zig:463:30: 0x108972a7c in Tree.isLvalExtra (arocc)
    switch (nodes.items(.tag)[@enumToInt(node)]) {
                             ^
/Users/ehaas/source/arocc/src/Tree.zig:458:23: 0x10891086f in Tree.isLval (arocc)
    return isLvalExtra(nodes, extra, value_map, node, &is_const);
                      ^
/Users/ehaas/source/arocc/src/Parser.zig:4660:29: 0x108a01572 in Parser.suffixExpr (arocc)
            if (!Tree.isLval(p.nodes.slice(), p.data.items, p.value_map, operand.node) or operand.ty.isConst()) {
                            ^
/Users/ehaas/source/arocc/src/Parser.zig:4579:48: 0x1089fad86 in Parser.unExpr (arocc)
                const suffix = try p.suffixExpr(lhs);
                                               ^
/Users/ehaas/source/arocc/src/Parser.zig:4327:20: 0x1089f99e1 in Parser.castExpr (arocc)
    return p.unExpr();
```

ehaas:
When using the parser to parse a preprocessor expression, report
an error and give up if an illegal token is encountered. This prevents
the parser from trying to use a non-existent tree to validate expressions

Fixes #101


Vexu:
Both of these checks seem like they belong in `Preprocessor.expr`.

src/Preprocessor.zig
```diff
@@ -408,7 +408,16 @@ fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) Error!bool {
...
                 tok.id = .zero;
             }
+        } else if (pp.tokens.len == start and !tok.id.validPreprocessorExprStart()) {
+            try pp.err(tok, .invalid_preproc_expr_start);
+            skipToNl(tokenizer);
+            return false;
+        } else if (tok.id.invalidPreprocessorBinOp()) {
+            try pp.err(tok, .invalid_preproc_operator);
+            skipToNl(tokenizer);
+            return false;
```
Vexu, line 418: This won't work you have to put the check in the switch below.
```c
#define FOO --
#if 0 FOO
#endif
```

ehaas: Made the requested change + rebased against master. One thing I noticed is that diagnostics generated from within that switch can end up pointing to the wrong token:
```c
int foo;
#if "test"
#endif
```
```
test.c:1:1: error: string literal in preprocessor expression
int foo;
^
```

ehaas: I think maybe we just need `pp.tokens.items(.loc)[start + i]` instead of `pp.tokens.items(.loc)[i]` for error locations

PR #105, opened 2021-11-18
Source: require source files to be utf-8 encoded

ehaas:
Currently this just scans the whole file up-front to ensure that it is
utf-8 encoded. The scan is performed in preprocessor even though addSource
seems like a more natural fit because returning an error from that function
invalidates the source and path buffers which are needed for error messages.

Subsequent tokenizer enhancements will be needed for consuming unicode
codepoints instead of bytes (e.g. for identifiers); at that time the extra
scan can be removed and tokenizer.next() can either return an error or a special
token to indicate invalid utf8

### DIFF
--- a/src/Source.zig
+++ b/src/Source.zig
@@ -1,3 +1,4 @@
+const std = @import("std");
 const Source = @This();
 
 pub const Id = enum(u32) {
@@ -15,6 +16,7 @@ pub const Location = struct {
 path: []const u8,
 buf: []const u8,
 id: Id,
+invalid_utf8_loc: ?Location = null,
 
 pub const LCS = struct { line: u32, col: u32, str: []const u8 };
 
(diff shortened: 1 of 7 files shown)


src/Source.zig
```diff
@@ -36,3 +37,25 @@ pub fn lineColString(source: Source, byte_offset: u32) LCS {
...
+            if (i + cp_len > buf.len) return i;
+            if (std.meta.isError(std.unicode.utf8Decode(buf[i .. i + cp_len]))) return i;
+            i += cp_len;
+        } else |_| return i;
+    }
+    return null;
+}
+
+pub fn locationOfInvalidUtf8(source: Source) ?Location {
+    if (offsetOfInvalidUtf8(source.buf)) |offset| {
```
Vexu, line 57: Not sure how big of an impact it will have but caching the result of this check would be a really simple enhancement to make.

ehaas: are you thinking something like `is_utf8: ?bool` as a member of `Source`?

ehaas: nm, I see we could just store the location itself. Just need to distinguish "haven't checked yet" and "does not have invalid utf8"

ehaas:
Updated to check and cache utf8 status when adding the source, but not actually log an error until preprocessing. That way valid files aren't repeatedly scanned every time they're used as includes.

PR #112, opened 2021-11-19
Parser: check for incomplete types before attempting field access

### LINKED ISSUE #111: Accessing field of incomplete struct triggers assertion
```c
struct foo;
void bar(struct foo *s) {
    s->a;
}
```

```
thread 21513479 panic: reached unreachable code
/Users/ehaas/source/zig/build/lib/zig/std/debug.zig:226:14: 0x1035b22f8 in std.debug.assert (arocc)
    if (!ok) unreachable; // assertion failure
             ^
/Users/ehaas/source/arocc/src/Type.zig:534:29: 0x1036efeef in Type.getField (arocc)
            std.debug.assert(!ty.data.record.isIncomplete());
                            ^
/Users/ehaas/source/arocc/src/Parser.zig:4751:37: 0x103726a18 in Parser.getFieldAccessField (arocc)
    const field = record_ty.getField(field_name) orelse {
                                    ^
/Users/ehaas/source/arocc/src/Parser.zig:4715:55: 0x10371ecda in Parser.suffixExpr (arocc)
            const field_ty = try p.getFieldAccessField(lhs.ty, name, true);
                                                      ^
/Users/ehaas/source/arocc/src/Parser.zig:4579:48: 0x1037176c6 in Parser.unExpr (arocc)
                const suffix = try p.suffixExpr(lhs);
                                               ^
/Users/ehaas/source/arocc/src/Parser.zig:4327:20: 0x103716321 in Parser.castExpr (arocc)
    return p.unExpr();
```

ehaas:
Fixes #111

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -269,6 +269,7 @@ pub const Tag = enum {
     bitfield_too_big,
     invalid_utf8,
     implicitly_unsigned_literal,
+    deref_incomplete_ty_ptr,
 };
 
 pub const Options = struct {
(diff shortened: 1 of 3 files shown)


Vexu:
It's generally good to copy clang but this error message is pretty lousy, surely we can come up with something more useful.

ehaas:
gcc says `error: dereferencing pointer to incomplete type ‘struct S’` - how about that?

ehaas:
That also prompted me to check `*s` and it looks like there's no error for that either - clang says `error: incomplete type 'struct S' where a complete type is required` and gcc says the same `dereferencing pointer to incomplete type ‘struct S’` - so maybe we use the gcc error text for both cases?

PR #117, opened 2021-11-20
Parser: handle remainder operator correctly for signed integers

### LINKED ISSUE #113: Parser: panic on modulo with negative constant
```c
void foo(void) {
    int x = 1 % -1;
}
```

```
thread 21566079 panic: remainder division by zero or negative value
/Users/ehaas/source/arocc/src/Parser.zig:4262:49: 0x105e89f88 in Parser.mulExpr (arocc)
                    .signed => |v| .{ .signed = @rem(v, rhs.val.signed) },
                                                ^
/Users/ehaas/source/arocc/src/Parser.zig:4206:28: 0x105e88d34 in Parser.addExpr (arocc)
    var lhs = try p.mulExpr();
                           ^
/Users/ehaas/source/arocc/src/Parser.zig:4174:28: 0x105e882c4 in Parser.shiftExpr (arocc)
    var lhs = try p.addExpr();
                           ^
/Users/ehaas/source/arocc/src/Parser.zig:4145:30: 0x105e87974 in Parser.compExpr (arocc)
    var lhs = try p.shiftExpr();
```

ehaas:
Fixes #113

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -4224,6 +4224,14 @@ fn addExpr(p: *Parser) Error!Result {
     return lhs;
 }
 
+/// Implements C's % operator for signed integers, for evaluating constant expressions
+/// caller guarantees rhs != 0
+/// caller guarantees lhs != std.math.minInt(i64) OR rhs != -1
+fn signedRemainder(lhs: i64, rhs: i64) i64 {
+    if (rhs > 0) return @rem(lhs, rhs);
+    return lhs - @divTrunc(lhs, rhs) * rhs;
+}
+
 /// mulExpr : castExpr (('*' | '/' | '%') castExpr)*´
 fn mulExpr(p: *Parser) Error!Result {
     var lhs = try p.castExpr();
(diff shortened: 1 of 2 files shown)


src/Parser.zig
```diff
@@ -4257,11 +4265,15 @@ fn mulExpr(p: *Parser) Error!Result {
                     else => unreachable,
                 };
             } else {
-                lhs.val = switch (lhs.val) {
-                    .unsigned => |v| .{ .unsigned = v % rhs.val.unsigned },
-                    .signed => |v| .{ .signed = @rem(v, rhs.val.signed) },
-                    else => unreachable,
-                };
+                if (lhs.val == .signed and lhs.val.signed == std.math.minInt(i64) and rhs.val.signed == -1) {
```
ehaas, line 4268: I'm not sure if the Zig compiler couldn't figure out the type inference or if I was just doing something wrong, but I wasn't able to incorporate this check as part of an if expression in the `.signed` prong of the switch

src/Parser.zig
```diff
@@ -4257,11 +4265,15 @@ fn mulExpr(p: *Parser) Error!Result {
...
+                    lhs.val = .unavailable;
```
Vexu, line 4269: Setting the value to `.unavailable` needs to be special cased for macros but here setting it to zero seems to match clang/GCC behavior in:
```c
#if (-9223372036854775807LL - 1LL) % -1 != 0
#error foo
#endif
```

Vexu (changes requested):

PR #120, opened 2021-11-24
Add unicode identifier support

ehaas:
Add support for unicode identifiers as specified by C99 / C11 standards. Matches clang behavior of allowing C99 identifiers in pre-C99 standards.


src/Tokenizer.zig
```diff
@@ -535,8 +542,8 @@ pub const Token = struct {
     /// belong to the implementation namespace, so we always convert them
     /// to keywords.
     /// TODO: add `.keyword_asm` here as GNU extension once that is supported.
-    pub fn getTokenId(comp: *const Compilation, str: []const u8) Token.Id {
-        const kw = all_kws.get(str) orelse return .identifier;
+    pub fn getTokenId(comp: *const Compilation, str: []const u8, default: Token.Id) Token.Id {
```
Vexu, line 545: Noticed late but this will always return `.extended_identifier` if it is called on a unicode identifier. So instead of adding a default parameter this shouldn't be called at all.

Vexu (commented):
Also basic unicode awareness will need to be added to diagnostics column calculations.

src/Tokenizer.zig
ehaas: Addressed in https://github.com/Vexu/arocc/pull/129

PR #129, opened 2021-11-26
Tokenizer: remove call to getTokenId for extended identifiers

ehaas:
keywords will never use characters outside the basic character set
so there's never a need to remap extended keywords.

### DIFF
--- a/src/Tokenizer.zig
+++ b/src/Tokenizer.zig
@@ -542,8 +542,8 @@ pub const Token = struct {
     /// belong to the implementation namespace, so we always convert them
     /// to keywords.
     /// TODO: add `.keyword_asm` here as GNU extension once that is supported.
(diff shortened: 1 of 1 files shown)


src/Tokenizer.zig
```diff
@@ -1453,7 +1444,7 @@ pub fn next(self: *Tokenizer) Token {
         switch (state) {
             .start, .line_comment => {},
             .u, .u8, .U, .L, .identifier, .extended_identifier => {
-                id = Token.getTokenId(self.comp, self.buf[start..self.index], state.identifierType());
+                id = if (state != .extended_identifier) Token.getTokenId(self.comp, self.buf[start..self.index]) else .extended_identifier;
```
Vexu, line 1447: It'd be easier to just add a `.extended_identifier => id = .extended_identifier,` prong.

Same could be said for the previous case but at least that has some logic beyond just a single assignment.

PR #136, opened 2021-11-27
Preprocessor: ensure token pasting rhs exists

### LINKED ISSUE #134: Crash: index out of bounds in macro expansion
```c
#define debug(s, t) foo(x ## # n
debug(1, 2);
```

```
thread 25571143 panic: index out of bounds
/Users/ehaas/source/arocc/src/Preprocessor.zig:713:51: 0x1043621ce in Preprocessor.expandFuncMacro (arocc)
                const raw_next = func_macro.tokens[tok_i + 1];
                                                  ^
/Users/ehaas/source/arocc/src/Preprocessor.zig:1003:53: 0x104305d44 in Preprocessor.expandMacroExhaustive (arocc)
                    var res = try pp.expandFuncMacro(macro_tok.loc, macro, &args, &expanded_args);
                                                    ^
/Users/ehaas/source/arocc/src/Preprocessor.zig:1056:33: 0x1042d540c in Preprocessor.expandMacro (arocc)
    try pp.expandMacroExhaustive(tokenizer, &buf, 0, 1, true);
```

ehaas:
Fixes #134

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -710,6 +710,7 @@ fn expandFuncMacro(
         const raw = func_macro.tokens[tok_i];
         switch (raw.id) {
             .hash_hash => {
+                if (tok_i + 1 == func_macro.tokens.len) break; // missing rhs due to previous error
                 const raw_next = func_macro.tokens[tok_i + 1];
                 const placeholder_token = Token{ .id = .empty_arg, .loc = .{ .id = raw_next.source, .byte_offset = raw_next.start } };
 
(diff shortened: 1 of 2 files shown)


doppioandante:
LGTM, is your fuzzer on a killing spree lately? :P

Vexu:
This fixes the problem but isn't the correct solution. Clang and GCC do not define the macro at all when it contains errors like this and we shouldn't either.

```c
#define debug(s, t) foo(x ## # a

#if defined debug
#error foo
#endif
```
This snippet does not hit the `#error` on clang/GCC but does on Aro.

doppioandante:
Oh right, nevermind, I was reading the change quickly. It should go in the defineFn and the other function that deals with simple macros.

PR #137, opened 2021-11-27
Tets runner: set language version using a pragma

Vexu:
Now that we have a nice pragma system we can use it instead of some hacky comment, especially since c++ style comments aren't even valid pre c99.

@ehaas there seems to be some bug in the way tokens are handled that appears depending on whether there is whitespace after the pragma or not, mind taking a look?

### DIFF
--- a/src/lib.zig
+++ b/src/lib.zig
@@ -2,6 +2,7 @@ pub const Codegen = @import("Codegen.zig");
 pub const Compilation = @import("Compilation.zig");
 pub const Diagnostics = @import("Diagnostics.zig");
 pub const Parser = @import("Parser.zig");
+pub const Pragma = @import("Pragma.zig");
 pub const Object = @import("Object.zig");
 pub const Preprocessor = @import("Preprocessor.zig");
 pub const Source = @import("Source.zig");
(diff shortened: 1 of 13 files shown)


ehaas:
Github is having lots of problems right now so I can't look at the CI output, but from running it on my own machine it seems like everything is working properly - it's just that the extra newline causes the line of the reported error to not match what is in `EXPECTED_ERRORS`. For example in the `__is_identifier.c` test case, `EXPECTED_ERRORS` has `"__is_identifier.c:25:25: error: missing ')', after builtin feature-check macro"` but that is referring to an error that is actually on line 26 - `#if __is_identifier(foo bar)`. The parser correctly reports line 26 for the error, which is why the test is failing. As far as I can tell it's just that the extra newline wasn't there when we were using the comment-based system for setting the version.

The other issue I'm seeing is that the `__STDC_VERSION__` static assert in `predefined macros.c` is not going to work, because those macros are defined before any pragmas are processed.

Vexu:
> it's just that the extra newline causes the line of the reported error to not match what is in `EXPECTED_ERRORS`.

Right, of course, I'm just too distrustful of whitespace.

> The other issue I'm seeing is that the `__STDC_VERSION__` static assert in `predefined macros.c` is not going to work, because those macros are defined before any pragmas are processed.

Hmm, I can't think of any simple workaround so this probably won't be worth it unless we need to add some other settings in the future.

PR #139, opened 2021-11-28
Type: handle integerPromotion of attributed types

ehaas:


Vexu:
There are a lot of other places where `.attributed` is not handled but I have those fixed locally.

Vexu:
Oh wait I had already pushed those and just missed this one, nvm.

Issue #81, opened 2021-11-06
Improve the diagnostics system


Vexu:
Currently the diagnostics system is a bit messy and verbose, some things that could improve it are:
* print the messages to strings when they are added to reduce the amount of tags and switches
* store info regarding the options in an array to avoid duplication in switches like `tagKind` and `tagOption`
* colored printing on Windows will need to be done eventually
* something else?

PR #141, opened 2021-11-28
Diagnostics: Resolve logic duplication between tagOption() and tagKind()

iddev5:
PR #73 introduced a new function tagOption() which duplicated some switches from tagKind()
This PR tries to deduplicate some logic as per suggested by #81 (point 2)

#81 suggest using an array to store options. I looked into that approach but all solutions turned out to be overengineered.

If anyone has some other plannings on this, please let know.
Feel free to close this if its not upto its target, or there are other plannings for this.

Implementation notes:
> - added new fields to Options: fatal error, error, warning... etc. These
are special fields which do not directly corelate to any CLI option
> - tagOption() now switches on every tag, returning the correct option
when available, otherwise returning what would have been its kind in
string
> - isValidOption() is used to check if its an option or a kind
> - tagKind() is refactored to first get the option name using tagOption()
and then move through Options searching for the corresponding field

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -327,6 +327,12 @@ pub const Options = struct {
     @"c99-compat": Kind = .off,
     @"unicode-zero-width": Kind = .warning,
     @"unicode-homoglyph": Kind = .warning,
+
+    @"fatal error": Kind = .@"fatal error",
+    @"error": Kind = .@"error",
+    warning: Kind = .warning,
+    note: Kind = .note,
+    off: Kind = .off,
 };
 
 list: std.ArrayList(Message),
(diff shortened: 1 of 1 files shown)


Vexu:
I'm going to look into how this will work with pedantic and some other stuff tomorrowish and then get back to you.

iddev5:
I think -pedantic should be handled in LangOpts.zig. Afaik, it is guaranteed for -pedantic to only give out warnings. The flag in particular need to be specifically stored somewhere because it also sets the language standard to base version (i.e -std=g11 to -std=c11).

As for dealing with -pedantic-errors, the pedantic field can be have to store an ``?enum { .@"error", .warning }`` which determines what version the user wants.

Vexu:
My alternative to this is #147. It not only solves the duplicated logic but also addresses the much bigger issue of having the data spread across 4-5 separate places.

iddev5:
Noticed it, was waiting for it to stabilize.

PR #147, opened 2021-11-29
Make diagnostics declarative

### LINKED ISSUE #81: Improve the diagnostics system

Vexu:
A long overdue quality of life improvement to the diagnostics system so that you only have to modify one plave to add a new error instead of 4-5 spread accross many files.

Closes #81.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -452,11 +452,6 @@ pub fn fatal(comp: *Compilation, tok: Token, comptime fmt: []const u8, args: any
     return comp.diag.fatal(source.path, lcs, fmt, args);
 }
 
-pub fn addDiagnostic(comp: *Compilation, msg: Diagnostics.Message) Error!void {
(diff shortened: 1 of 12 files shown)


iddev5 (commented):
Looks great! Was just wondering how would special options like -Wpedantic or -Wextra would be handled? Like, handling them as special cases in Diagnostics.set()?

Vexu:
> Was just wondering how would special options like -Wpedantic or -Wextra would be handled? Like, handling them as special cases in Diagnostics.set()?

Probably yeah, `-Wfatal-errors` is already special cased in that way.

ehaas:
Very nice, great example of Zig's metaprogramming capabilities

PR #150, opened 2021-11-30
Type: add helper function for anonymous records

### LINKED ISSUE #146: Inactive union field access via typeof-type
```c
union U { int x; };
void foo(void) {
    typeof(union U) u = { 1 };
    typeof(u);
}
```

```
thread 26706946 panic: access of inactive union field
/Users/ehaas/source/arocc/src/Parser.zig:766:58: 0x109bef3fd in Parser.decl (arocc)
        if (decl_spec.ty.isRecord() and decl_spec.ty.data.record.name[0] != '(') return true;
                                                         ^
/Users/ehaas/source/arocc/src/Parser.zig:3319:19: 0x109c23922 in Parser.compoundStmt (arocc)
        if (p.decl() catch |er| switch (er) {
                  ^
/Users/ehaas/source/arocc/src/Parser.zig:887:41: 0x109bedeb4 in Parser.decl (arocc)
        const body = (try p.compoundStmt(true)) orelse {
```

ehaas:
Fixes #146

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -758,8 +758,11 @@ fn decl(p: *Parser) Error!bool {
     };
     var init_d = (try p.initDeclarator(&decl_spec)) orelse {
         _ = try p.expectToken(.semicolon);
-        if (decl_spec.ty.is(.@"enum")) return true;
-        if (decl_spec.ty.isRecord() and decl_spec.ty.data.record.name[0] != '(') return true;
+        if (decl_spec.ty.is(.@"enum") or
+            (decl_spec.ty.isRecord() and !decl_spec.ty.isAnonymousRecord() and
+            !decl_spec.ty.isTypeof())) // we follow GCC and clang's behavior here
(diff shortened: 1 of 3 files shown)


ehaas:
In the associated test, for the `typeof(u);` line, clang warns `warning: declaration does not declare anything [-Wmissing-declarations]`. However it doesn't do that if you replace `typeof(u)` with `union U`. aro does not issue the warning either way. Not sure what the most correct/consistent behavior would be.

Vexu:
> In the associated test, for the `typeof(u);` line, clang warns `warning: declaration does not declare anything [-Wmissing-declarations]`. However it doesn't do that if you replace `typeof(u)` with `union U`. aro does not issue the warning either way. Not sure what the most correct/consistent behavior would be.

GCC and clang both seem to only error on `typeof` so we should too. It was simple to implement so I added it along with some comments to explain why the anonymous record check works.

PR #151, opened 2021-11-30
Change the way source locations are stored

### LINKED ISSUE #38: [Source.zig] Change Source.Location from linked list to array
> A future improvement will be to convert the source location mechanism into using contiguous arrays instead of linked lists now that they are being iterated upon a lot more.
> 
> _Originally posted by @Vexu in https://github.com/Vexu/arocc/pull/33#discussion_r716839133_

`Source.Location` is a linked list that works like this:
- the first item is always the original source location where the current token has been originally found, if any
- the following items are the source locations of the macro expansions used to form the current token, the first one being the latest and the last one being the first expansion

Vexu:
Closes #38

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -407,7 +407,7 @@ pub fn addSource(comp: *Compilation, path: []const u8) !Source {
     return source;
 }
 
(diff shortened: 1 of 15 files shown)


ehaas:
I rebased my `_Pragma` branch against this and everything works as before 👍

PR #152, opened 2021-12-01
Add -fdollars-in-identifiers and -fno-dollars-in-identifiers option

### LINKED ISSUE #132: Add `-fdollars-in-identifiers` / `-fno-dollars-in-identifiers` option
By default clang and gcc allow dollar signs in identifiers even though they are not part of the basic character set. Some target machines do not support dollars in identifiers because the target assembler does not allow them, so `-fno-dollars-in-identifiers` can be used to make them illegal.

iddev5:
Also adds the related warnings -W(no-)dollar-in-identifier-extension

Closes #132

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -143,12 +143,13 @@ fn eatIdentifier(p: *Parser) !?TokenIndex {
             var it = std.unicode.Utf8View.initUnchecked(slice).iterator();
             var loc = p.pp.tokens.items(.loc)[p.tok_i];
 
-            if (mem.indexOfScalar(u8, p.tokSlice(p.tok_i), '$')) |i| {
+            if (mem.indexOfScalar(u8, slice, '$')) |i| {
                 loc.byte_offset += @intCast(u32, i);
                 try p.pp.comp.diag.add(.{
                     .tag = .dollar_in_identifier_extension,
                     .loc = loc,
                 });
+                loc = p.pp.tokens.items(.loc)[p.tok_i];
             }
 
             while (it.nextCodepoint()) |c| {
@@ -165,7 +166,7 @@ fn eatIdentifier(p: *Parser) !?TokenIndex {
     if (!p.pp.comp.langopts.dollars_in_identifiers) {
         if (p.tok_ids[p.tok_i] == .invalid and p.tokSlice(p.tok_i)[0] == '$') {
             try p.err(.dollars_in_identifiers);
-            defer p.tok_i += 2;
+            p.tok_i += 1;
             return error.ParsingFailed;
         }
     }
@@ -1269,34 +1270,10 @@ const InitDeclarator = struct { d: Declarator, initializer: NodeIndex = .none };
 fn attribute(p: *Parser) Error!Attribute {
     const name_tok = p.tok_i;
     switch (p.tok_ids[p.tok_i]) {
-        .identifier, .keyword_const, .keyword_const1, .keyword_const2 => {},
-        .extended_identifier => {
-            const slice = p.tokSlice(p.tok_i);
-            var it = std.unicode.Utf8View.initUnchecked(slice).iterator();
-            var loc = p.pp.tokens.items(.loc)[p.tok_i];
-
-            if (mem.indexOfScalar(u8, p.tokSlice(p.tok_i), '$')) |i| {
-                loc.byte_offset += @intCast(u32, i);
-                try p.pp.comp.diag.add(.{
-                    .tag = .dollar_in_identifier_extension,
-                    .loc = loc,
-                });
-            }
-
-            while (it.nextCodepoint()) |c| {
-                if (try checkIdentifierCodepoint(p.pp.comp, c, loc)) break;
-                loc.byte_offset += std.unicode.utf8CodepointSequenceLength(c) catch unreachable;
-            }
-        },
-        else => {
-            try p.errExtra(.expected_token, p.tok_i, .{ .tok_id = .{
-                .expected = .identifier,
-                .actual = p.tok_ids[p.tok_i],
-            } });
-            return error.ParsingFailed;
-        },
+        .keyword_const, .keyword_const1, .keyword_const2 => p.tok_i += 1,
+        else => _ = try p.expectIdentifier(),
     }
-    p.tok_i += 1;
+
     switch (p.tok_ids[p.tok_i]) {
         .comma, .r_paren => { // will be consumed in attributeList
             return Attribute{ .name = name_tok };
@@ -1362,13 +1339,6 @@ fn attribute(p: *Parser) Error!Attribute {
                 .name = name_tok,
                 .params = try p.addNode(node),
             };
-        },
-        .invalid => {
-            if (!p.pp.comp.langopts.dollars_in_identifiers and p.tokSlice(p.tok_i)[0] == '$') {
-                try p.err(.dollars_in_identifiers);
-                p.tok_i += 2;
-            }
-            return error.ParsingFailed;
         },
         else => return error.ParsingFailed,
     }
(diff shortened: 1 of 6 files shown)


iddev5:
This is actually complete but with -fno-dollars-in-identifier, the errors are quite different from clang. This is due to differences in parsing.

Taking the added test:
arocc:
```
./test.c:12:6: error: expected identifier or '('
./test.c:12:6: error: expected ';', found invalid bytes
./test.c:15:7: error: expected identifier or '('
./test.c:15:7: error: expected ';', found invalid bytes
./test.c:18:6: error: expected identifier or '('
./test.c:18:6: error: expected ';', found invalid bytes
./test.c:18:10: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
./test.c:18:18: warning: non-void function 'ther' does not return a value [-Wreturn-type]
```

clang:
```
test.c:12:6: error: variable has incomplete type 'void'
test.c:12:9: error: expected ';' after top level declarator
```

Here, it seems arocc keeps on parsing even after the $ char is found.

Vexu:
> Here, it seems arocc keeps on parsing even after the $ char is found.

You can stop that by returning `error.ParsingFailed`.

src/Parser.zig
```diff
@@ -137,9 +137,13 @@ fn checkIdentifierCodepoint(comp: *Compilation, codepoint: u21, loc: Source.Loca
 
 fn eatIdentifier(p: *Parser) !?TokenIndex {
     if (p.tok_ids[p.tok_i] == .identifier) {
+        if (mem.indexOf(u8, p.tokSlice(p.tok_i), "$")) |_|
```
ehaas, line 140: Can use `indexOfScalar` (with `'$'`) instead of `indexOf`

ehaas:
If we wanted to have better error messaging than clang - an alternative approach could be to treat identifiers with `$`'s as extended identifiers. Then the dollar check could be isolated to `checkIdentifierCodepoint`, which means we wouldn't have to run it for every identifier, which is nice since they're probably pretty uncommon in real code. Finally we could have a separate diagnostic like `illegal_dollar_in_identifier` that prints exactly what is wrong - clang it treats it like any other random illegal character which might be confusing.

iddev5:
Combining some parts from here and there, this is the new change:
- any identifier containing '$' is now exclusively an extended_identifier
- no matter whether dollars_in_identifiers is true or false, it is always tokenized as regular
- eatIdentifier() extends the check to take into account the previous case and then emit an error ``illegal character '$' in identifier

Overall, i think the error messages are much better now and there are no false warnings

src/Parser.zig
iddev5: Done in recent commit

Vexu (changes requested):
This currently ignores the preprocessor.
```
#define foo$ bar
foo$
// clang
// a.c:1:12: warning: ISO C99 requires whitespace after the macro name [-Wc99-extensions]
$ bar$
// Aro
bar
```
Splitting the token at the dollar sign fixes that but you'll need to add some extra logic in `Parser` to make the errors nice again.
```diff
diff --git a/src/Tokenizer.zig b/src/Tokenizer.zig
index e3502bc..eaae120 100644
--- a/src/Tokenizer.zig
+++ b/src/Tokenizer.zig
@@ -771,8 +771,7 @@ pub fn next(self: *Tokenizer) Token {
                 'u' => state = .u,
                 'U' => state = .U,
                 'L' => state = .L,
-                'a'...'t', 'v'...'z', 'A'...'K', 'M'...'T', 'V'...'Z', '_' => state = .identifier,
-                '$' => state = .extended_identifier,
+                'a'...'t', 'v'...'z', 'A'...'K', 'M'...'T', 'V'...'Z', '_', '$' => state = .identifier,
                 '=' => state = .equal,
                 '!' => state = .bang,
                 '|' => state = .pipe,
@@ -1055,7 +1054,10 @@ pub fn next(self: *Tokenizer) Token {
             },
             .identifier, .extended_identifier => switch (c) {
                 'a'...'z', 'A'...'Z', '_', '0'...'9' => {},
-                '$' => state = .extended_identifier,
+                '$' => {
+                    id = if (state == .identifier) Token.getTokenId(self.comp, self.buf[start..self.index]) else .extended_identifier;
+                    break;
+                },
                 else => {
                     if (c <= 0x7F or !Token.mayAppearInIdent(self.comp, c, .inside)) {
                         id = if (state == .identifier) Token.getTokenId(self.comp, self.buf[start..self.index]) else .extended_identifier;
```

iddev5:
There are some more problems now, so marked as draft

iddev5:
I think this is ready now, or at least the error messages can't get any better without any huge breaking changes

Explaining the recent commits:
As suggested by ethan, $ are exclusively part of extended identifiers again, and checked in ``Token.mayAppearInIdent`` once.
Moreover, in -fno-dollars-in-identifiers mode, the token immediately after the identifier is check if its a $ symbol, if so, then it is illegal and error must be shown (because in ``xyz$`` for ex, xyz is an identifier followed immediately by a $ symbol)

src/Parser.zig
```diff
@@ -136,21 +136,36 @@ fn checkIdentifierCodepoint(comp: *Compilation, codepoint: u21, loc: Source.Loca
...
-        const slice = p.tokSlice(p.tok_i);
-        var it = std.unicode.Utf8View.initUnchecked(slice).iterator();
-        var loc = p.pp.tokens.items(.loc)[p.tok_i];
-        while (it.nextCodepoint()) |c| {
-            if (try checkIdentifierCodepoint(p.pp.comp, c, loc)) break;
-            loc.byte_offset += std.unicode.utf8CodepointSequenceLength(c) catch unreachable;
+    switch (p.tok_ids[p.tok_i]) {
+        .identifier => {
+            defer p.tok_i += 1;
+        },
```
Vexu, line 142:
```suggestion
        .identifier => {},
```
The `defer p.tok_i += 1;` can be added to the end of the function.

src/Parser.zig
```diff
@@ -136,21 +136,36 @@ fn checkIdentifierCodepoint(comp: *Compilation, codepoint: u21, loc: Source.Loca
...
+        .extended_identifier => {
+            if (mem.indexOfScalar(u8, p.tokSlice(p.tok_i), '$')) |_|
+                try p.err(.dollar_in_identifier_extension);
```
Vexu, line 145:
```suggestion
            if (mem.indexOfScalar(u8, p.tokSlice(p.tok_i), '$')) |i| {
                loc.byte_offset += i;
                try p.pp.comp.diag.add(.{
                    .tag = .dollar_in_identifier_extension,
                    .loc = loc,
                });
            }
```

src/Parser.zig
```diff
@@ -136,21 +136,36 @@ fn checkIdentifierCodepoint(comp: *Compilation, codepoint: u21, loc: Source.Loca
...
+            }
+        },
+        else => return null,
+    }
+
+    // Handle illegal '$' characters in identifiers
+    if (!p.pp.comp.langopts.dollars_in_identifiers) {
+        if (p.tok_ids[p.tok_i] == .invalid and p.tokSlice(p.tok_i)[0] == '$') {
+            try p.err(.dollars_in_identifiers);
+            defer p.tok_i += 2;
```
Vexu, line 163:
```suggestion
            p.tok_i += 2;
```
Useless defer.

Vexu (changes requested):
Just some small nitpicks and this is done.

I duplicated some of the `eatIdentifier` logic in `attribute` to handle `const` and its variants so you'll have to modify it, sorry about that.

test/cases/no dollars in identifiers.c
```diff
@@ -0,0 +1,7 @@
+void foo$() { }
+
+void fib() {
+  int $test;
+}
+
+void ano$ther() {}
```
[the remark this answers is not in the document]
iddev5: I didn't added the ``no dollars in identifiers`` case because I think sooner or later, we would need more such options and thus a proper solution is needed, like parsing the whole (or atleast a larger) set of cli args from the first comment?

iddev5:
Made all the necessary changes!

Vexu:
Thanks, I made some final small tweaks and extracted that suggestion into an issue so that it's not forgotten.

Issue #42, opened 2021-10-11
Attributes: validate argument counts and types


ehaas:
Currently all attributes are allowed to have an arbitrary number of arguments of any type

PR #153, opened 2021-12-01
Parser: parse attributes at end of record field decls

### LINKED ISSUE #140: Attributes are not parsed for record fields
```c
typedef struct {
  long long __clang_max_align_nonce1
      __attribute__((__aligned__(__alignof__(long long))));
  long double __clang_max_align_nonce2
      __attribute__((__aligned__(__alignof__(long double))));
} max_align_t;
```

ehaas:
Fixes #140

I manually verified that the provided test case matches clang's behavior (`__clang_max_align_nonce1` gets an `__aligned__(8)` attribute, `nonce2` gets `__aligned__(8)` and `packed`, `nonce3` gets `__aligned__(8)`, and `__clang_max_align_nonce2` gets `__aligned__(8)`). I don't have any good ideas of how to test this though - any thoughts?

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1733,9 +1733,16 @@ fn recordDecls(p: *Parser) Error!void {
 
 /// recordDeclarator : keyword_extension? declarator (':' constExpr)?
 fn recordDeclarator(p: *Parser) Error!bool {
+    const attr_buf_top = p.attr_buf.items.len;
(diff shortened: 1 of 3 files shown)


Vexu:
> I don't have any good ideas of how to test this though - any thoughts?

I don't think there is any good way right now with how incomplete attribute handling is. The most important part for me right now is that they are parsed properly. The rest can be handled as part of #42.

PR #157, opened 2021-12-02
Preprocessor: implement _Pragma operator

### LINKED ISSUE #68: Implement C99 `_Pragma` operator
https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html#index-_005fPragma

ehaas:
error.StopPreprocessing can now occur when expanding a macro, so I had to
increase the size of the expandMacro error set. Preprocessor.preprocess is
now a simple wrapper around preprocessImpl which catches StopPreprocessing
and ignores it; this way the StopPreprocessing logic can exist in just
one place.

Closes #68


ehaas:
Two issues with this one currently:

1. In the provided test case - clang treats `_Pragma("GCC diagnostic error \"-Wpointer-integer-compare\"")` as also activating `-Werror`. So subsequently reducing the error level to warning doesn't actually make it a warning. Our behavior makes more sense to me but I don't think it would be hard to adopt clang's behavior (I think it would be a separate issue from this one).
2. More importantly - if any generated pragmas need to do diagnostics which refer to a source location (e.g. if there is an error, or via `GCC warning "foo"`, etc) , it will crash, because there is no actual associated `Source`. clang seems to use the concept of `<scratch space>` for this:
```c
_Pragma("GCC warning \"This is a warning\"")
```
```
test.c:1:1: warning: This is a warning [-W#pragma-messages]
_Pragma("GCC warning \"This is a warning\"")
^
<scratch space>:2:6: note: expanded from here
 GCC warning "This is a warning"
     ^
1 warning generated.
```

It seems like `pp.generated` might need to be made available to `Compilation` to handle this?

Vexu:
> 1. In the provided test case - clang treats `_Pragma("GCC diagnostic error \"-Wpointer-integer-compare\"")` as also activating `-Werror`. So subsequently reducing the error level to warning doesn't actually make it a warning. Our behavior makes more sense to me but I don't think it would be hard to adopt clang's behavior (I think it would be a separate issue from this one).

Does GCC behave the same as clang? Regardless i think we should try to match GCC here.

> 2. More importantly - if any generated pragmas need to do diagnostics which refer to a source location (e.g. if there is an error, or via `GCC warning "foo"`, etc) , it will crash, because there is no actual associated `Source`. clang seems to use the concept of `<scratch space>` for this:
> It seems like `pp.generated` might need to be made available to `Compilation` to handle this? 

I'll look into this in #151, it'd simplify some other things too and enable one other change I want to make.

ehaas:
Looks like GCC works in the expected way:
```c
#define PRAGMA(X) _Pragma(#X)
void bar(void);
void foo(void) {
    int x, y;
    _Pragma("GCC diagnostic error \"-Wuninitialized\"")
    if (x) bar();
    PRAGMA(GCC diagnostic warning "-Wuninitialized")
    if (y) bar();
}
```

```
test.c: In function ‘foo’:
test.c:6:8: error: ‘x’ is used uninitialized in this function [-Werror=uninitialized]
     if (x) bar();
        ^
test.c:8:8: warning: ‘y’ is used uninitialized in this function [-Wuninitialized]
     if (y) bar();
        ^
cc1: some warnings being treated as errors
```

ehaas:
Do you think the updated location tracking will be enough for me to fix generated pragma issue I mentioned, or will some other change need to happen first?

Vexu:
You could solve it by calling `addExpansionLocation` on each of the destringified tokens as is done in `makeGeneratedToken` or if you'll give me a day or two I'll try to make that change to how the generated source is being handled.

ehaas:
just did a quick proof of concept locally and it worked :)

```
./test.c:1:9: warning: Foobar [-W#pragma-messages]
_Pragma("GCC warning \"Foobar\"")
        ^
1 warning generated.
```

PR #162, opened 2021-12-04
Parser: fix standalone __extension__ not followed by any other token crashes

### LINKED ISSUE #145: Crash: standalone __extension__
An `__extension__` keyword not followed by any other tokens will crash, e.g. a file that just contains `__extension__`

```
thread 26702182 panic: index out of bounds
/Users/ehaas/source/arocc/src/Parser.zig:169:18: 0x108f30147 in Parser.eatToken (arocc)
    if (p.tok_ids[p.tok_i] == id) {
                 ^
/Users/ehaas/source/arocc/src/Parser.zig:595:22: 0x108f1abf8 in Parser.parse (arocc)
    while (p.eatToken(.eof) == null) {
                     ^
/Users/ehaas/source/arocc/src/main.zig:265:32: 0x108efcfd9 in processSource (arocc)
    var tree = try Parser.parse(&pp);
```

iddev5:
Now it shows an error which complains about missing declaration after ``__extension__`` in global scope
Closes #145

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -591,7 +591,7 @@ pub fn parse(pp: *Preprocessor) Compilation.Error!Tree {
     _ = try p.addNode(.{ .tag = .invalid, .ty = undefined, .data = undefined });
     try p.defineVaList();
 
-    while (p.eatToken(.eof) == null) {
+    while (p.tok_i < p.tok_ids.len and p.eatToken(.eof) == null) {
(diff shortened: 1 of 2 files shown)


Vexu:
The issue is caused by the `__extension__` keyword being eaten twice, first by `eatToken` and then later in the loop by `tok_i += 1`. I'd prefer to see that being fixed instead of a workaround like this.

PR #167, opened 2021-12-05
Pragma: improve misleading warning messages

### LINKED ISSUE #166: Pragmas: improve warning messages
Taking the following code as example,
```c
#pragma GCC gibberish
#pragma GCC diagnostics gibberish
```

The warning messages generated by arocc are:
```
warning: unsupported #pragma directive 'GCC' [-Wunsupported-pragma]
#pragma GCC gibberish
        ^
warning: unsupported #pragma directive 'GCC' [-Wunsupported-pragma]
#pragma GCC diagnostic gibberish
        ^
```

These are mostly misleading and don't match clang/gcc's behavior (note that both of them dont warn on first case). On top of that, both of them have their own warning messages but different controlling options (weird...)

GCC: ``warning: expected [error|warning|ignored|push|pop] after ‘#pragma GCC diagnostic’ [-Wpragmas]``
Clang: ``warning: pragma diagnostic expected 'error', 'warning', 'ignored', 'fatal', 'push', or 'pop' [-Wunknown-pragmas]``

iddev5:
An attempt to fix #166 

This creates two different warning messages for both cases mentioned here. But controlling option is only made available for the second (diagnostics case) following clang's style.

### DIFF
--- a/src/pragmas/gcc.zig
+++ b/src/pragmas/gcc.zig
@@ -131,7 +131,7 @@ fn preprocessorHandler(_: *Pragma, pp: *Preprocessor, start_idx: TokenIndex) Pra
                 error.UnknownPragma => {
                     const tok = pp.tokens.get(start_idx + 2);
                     return pp.comp.diag.add(.{
-                        .tag = .unknown_pragmas_directive,
+                        .tag = .unknown_gcc_pragma_directive,
                         .loc = tok.loc,
                     }, tok.expansionSlice());
                 },
@@ -163,7 +163,7 @@ fn preprocessorHandler(_: *Pragma, pp: *Preprocessor, start_idx: TokenIndex) Pra
         }
     } else {
         return pp.comp.diag.add(.{
-            .tag = .unknown_pragmas,
+            .tag = .unknown_gcc_pragma,
             .loc = directive_tok.loc,
         }, directive_tok.expansionSlice());
     }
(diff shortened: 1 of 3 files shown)


iddev5:
I think this is pretty complete. However I think ``#pragma GCC xyz`` warning also deserves an option, though I am not sure what it should be called given that gcc and clang both
- dont have it 
- have different option name for case ``#pragma GCC diagnostic xyz``

src/pragmas/gcc.zig
```diff
@@ -126,7 +127,16 @@ fn preprocessorHandler(_: *Pragma, pp: *Preprocessor, start_idx: TokenIndex) Pra
...
-            .diagnostic => return diagnosticHandler(pp, start_idx + 2),
+            .diagnostic => return diagnosticHandler(pp, start_idx + 2) catch |err| switch (err) {
+                error.UnknownPragma => {
+                    const tok = pp.tokens.get(start_idx + 2);
+                    return pp.comp.diag.add(.{
+                        .tag = .unknown_pragmas_directive,
+                        .loc = tok.loc,
+                    }, tok.expansionSlice());
+                },
+                else => |e| return e,
```
Vexu, line 138: Why is this here and not in `diagnosticHandler`?

iddev5: Because doing it inside ``diagnosticHandler`` prints the warning message twice. I think that's because the function is also used in ``parserHandler`` when pragma syntax error do not need to be checked.

ehaas: You could add a bool parameter to `diagnosticHandler` called `diagnose_unknown` (or a 2-element enum with `diagnose_unknown` & `ignore_unknown`) that is true when called from `preprocessorHandler` and false when called from `parserHandler`, which controls whether `unknown_pragmas_directive` gets issued

PR #180, opened 2021-12-08
Parser: add support for statement expressions

### LINKED ISSUE #171: Add statement expression support
A statement expression is a compound statement enclosed in parentheses. The value of the statement expression is the value of the last semicolon-terminated expression within the compound statement.

Simple example
```c
int foo(void) {
    int y = ({
        int x = 5;
        x++;
        x - 5;
    });
    return y;
}
```
(This would return `1`)

See full description: https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html

In clang this construct is available in all language standards

ehaas:
Closes #171

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -106,6 +106,7 @@ pub const Options = struct {
     @"predefined-identifier-outside-function": ?Kind = null,
     @"many-braces-around-scalar-init": ?Kind = null,
     uninitialized: ?Kind = null,
+    @"gnu-statement-expression": ?Kind = null,
 };
 
 const messages = struct {
@@ -1345,6 +1346,16 @@ const messages = struct {
         const extra = .str;
         const opt = "uninitialized";
         const kind = .off;
+    };
+    const gnu_statement_expression = struct {
+        const msg = "use of GNU statement expression extension";
+        const opt = "gnu-statement-expression";
+        const kind = .off;
+        const suppress_gnu = true;
+    };
+    const stmt_expr_not_allowed_file_scope = struct {
+        const msg = "statement expression not allowed at file scope";
+        const kind = .@"error";
     };
 };
 
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -911,7 +911,7 @@ fn decl(p: *Parser) Error!bool {
             }
         }
 
-        const body = (try p.compoundStmt(true)) orelse {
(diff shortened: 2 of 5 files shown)


ehaas:
There are (at least) two issues currently:

1. How to handle unused expressions - if a statement expression is unused, clang warns by pointing to the last statement in the statement expression and says that it's unused. `maybeWarnUnused` has no way of knowing if the result is the last statement in a statement expression, as far as I can tell. One idea I had was that it could push append to an ArrayList if currently parsing a statement expression, and then clean things up after the full statement expression had been parsed.
2. If a statement expression is used as an initializer, the statement expression is allowed to refer to the variable it's initializing. I didn't see any non-invasive ways to support this, since it seems like we need a declref to a node that doesn't exist yet.

Vexu:
>1. How to handle unused expressions 

I would solve this by adding another parameter to `compoundStmt`:
```zig
stmt_expr_state: ?*struct {
    last_expr_loc: TokenIndex,
    last_expr_res: Result,
},
```
Then just tweak the mechanism for unused value error a bit. This would also help get rid of `compoundStmtLastNode`.
 
>2. If a statement expression is used as an initializer, the statement expression is allowed to refer to the variable it's initializing. I didn't see any non-invasive ways to support this, since it seems like we need a declref to a node that doesn't exist yet.

`decl_ref` currently doesn't need the variable to have any node so this shouldn't be an issue as long as you can limit the visibility to inside the statement expression.

ehaas:
I might be missing something obvious, but I'm not sure how to get the `unused result` warning to point to the relevant expression. My current strategy is to unconditionally remove the `unused result` warning for the final expression in a statement expression, since it might actually be used. The problem is that `maybeWarnUnused` gets the `expr_start` of the whole statement expression. In theory it would be possible to get the location of the final expression but it seems like it would be really messy.

The alternative would be to *not* unconditionally remove the unused result warning; but then we would need to remove it later and I think we're back at the same place of needing to figure out the location of the final expression (since we wouldn't want to remove an unused result warning pointing to an earlier expression in the stmt expr.

Vexu:
> I might be missing something obvious, but I'm not sure how to get the `unused result` warning to point to the relevant expression.

I don't think it is possible in any reasonable way currently but once we start attaching a source token to the nodes in some way we should be able to get much closer.

PR #182, opened 2021-12-09
Parser: allow self-referential initializers

ehaas:


ehaas:
What do you think of this approach for self-referential initializers? I'm not currently issuing the warning because I'm not totally sure how to approach it - it should only be issued if the *value* is used in a calculation (so plain self-assignment is fine, and using your own address is fine), and it should only be issued once per init (so `int foo = foo + foo + foo` only issues one warning). Do you think we should have a step after the initializer is parsed, where we go through its AST and look for an illegal self-use? Or try to issue the warning while parsing the initializer?

ehaas:
Actually, just realized this won't work with nested statement expression initializers if the inner one refers to the outer.

Vexu:
> I'm not currently issuing the warning because I'm not totally sure how to approach it.

`Scope.Symbol` currently has 32 unused bits that can be used for flags. That combined with a new union field in `Result.val` should be enough to implement uninitialized value errors. But since they are off by default they do not block this pr and it's enough to just open an issue as a reminder.

Also I fixed the array initializer bug in a12673b0a8893067a4d0c44408d06c4647d5e578.

PR #201, opened 2021-12-14
Tokenizer: ignore multiline comments with no intervening whitespace

### LINKED ISSUE #199: Crash: integer literal immediately after end of multiline comment
I think this is just a matter of `Tokenizer` not updating `start` after seeing `'/'` when in the `.multi_line_comment_asterisk` state (line 1373)

```c
void foo(void) {
    /* */1;
}
```

ehaas:
If a multiline comment is immediately followed by a non-whitespace token,
and we are in preprocess-only mode, replace the comment with a single space.
Otherwise completely ignore the comment.

Fixes #199


ehaas:
Note: we could repurpose `.macro_ws` for what `.multi_line_comment_whitespace` is being used to accomplish, if we wanted to save a `Token.Id`, at a small cost of clarity

Vexu:
> Note: we could repurpose `.macro_ws` for what `.multi_line_comment_whitespace` is being used to accomplish, if we wanted to save a `Token.Id`, at a small cost of clarity

While a more descriptive name would be nice, I fear this will reintroduce most of the issues I resolved by adding `macro_ws`.

ehaas:
Does that mean you just don't want to re-use `.macro_ws` or that you would prefer a different approach to solving #199

Vexu:
I meant that we should either sacrifice the clarity or carefully check that `.multi_line_comment_whitespace` is always converted to `.macro_ws` when necessary.

ehaas:
Reusing the `.macro_ws` definitely sounds better. I added a few more tests and verified the output matches clang.

Vexu:
I changed it to return `.whitespace` instead of `.macro_ws` since the rest of the compiler doesn't expect the tokenizer to return special tokens and it was causing issues such as the `multiline comment.c` test cases not compiling due to a special token slipping by:
```
./a.c:26:4: error: expected identifier or '('
int/**/a;
   ^
./a.c:26:4: error: expected ';', found ' '
int/**/a;
   ^
./a.c:26:8: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
int/**/a;
       ^
```

The change in `prettyPrintTokens` would need to happen anyways since it was causing comments to be printed.

PR #203, opened 2021-12-16
Parser: warn if attributes placed before struct/enum/union

ehaas:
Also fixes a crash that occurred if an invalid attribute occurred at the end
of a file, since no attribute is added to the buffer in that case.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -281,11 +281,22 @@ pub fn todo(p: *Parser, msg: []const u8) Error {
     return error.ParsingFailed;
 }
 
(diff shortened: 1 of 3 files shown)


ehaas:
This is probably not the right place for this check; I'm also adding some verification for attribute args. I'll update this attribute check while I'm at it.

Vexu:
> I'm also adding some verification for attribute args.

I was also planning on doing this and documented the argument types fo each attribute in 398446e296cff82dd73d4723cc6721144748da53, maybe it'll be useful to you?

ehaas:
Oh nice, definitely. The four attributes which are directly used by `translate-c` are `section` (vars and functions), `packed` (records), `aligned` (vars, fields, and functions), and `cleanup` (vars); so I thought I'd start with those.

PR #204, opened 2021-12-17
Parser: handle the possibility of invalid attribute at eof

ehaas:
Invalid attributes are not added to `attr_buf`, so a different strategy
is needed to check whether any attributes were parsed.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -797,12 +797,10 @@ fn decl(p: *Parser) Error!bool {
         }
         switch (p.tok_ids[first_tok]) {
             .asterisk, .l_paren, .identifier, .extended_identifier => {},
-            else => if (attr_buf_top == p.attr_buf.items.len) {
-                return false;
-            } else {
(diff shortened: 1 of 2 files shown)


src/Parser.zig
```diff
@@ -797,12 +797,10 @@ fn decl(p: *Parser) Error!bool {
         }
         switch (p.tok_ids[first_tok]) {
             .asterisk, .l_paren, .identifier, .extended_identifier => {},
-            else => if (attr_buf_top == p.attr_buf.items.len) {
-                return false;
-            } else {
+            else => if (attr_tokens_seen) {
```
Vexu, line 800: Do you think this check will be needed in the other places? Otherwise this check can just be `p.tok_i != first_tok`.

ehaas: I wasn't sure if `declSpec` could consume tokens without being successful, but it looks like it can't. Updated it to just do that comparison.

Issue #40, opened 2021-10-11
Don't always use nodes to store attribute arguments


ehaas:
Using nodes is wasteful since we usually don't need to explicitly store the tag or type (it's known from the attribute name + arg position)

Attribute arguments can be of the following types:
1. Identifier (store using `TokenIndex`)
2. string literal (question: store with index+len or `[]const u8`?)
3. signed or unsigned integer (some may have further restrictions, e.g. `aligned` requires a power of 2)
4. expression (this is used by the `copy` attribute https://gcc.gnu.org/onlinedocs/gcc/Common-Type-Attributes.html#index-copy-type-attribute) - in this case we'll probably need a `NodeIndex`. We may also need to special-case the parser so it doesn't try to evaluate the expression.

Issue #43, opened 2021-10-11
Validate attribute context in `decl()` after parsing


ehaas:
`decl()` allows any leading attribute since we don't know if we're parsing a function or variable declaration. Once we know what we've parsed we should retroactively validate those attributes for the relevant context.

PR #207, opened 2021-12-18
Basic validation for `aligned` attribute

ehaas:
Based on doing this, I think we should probably do #40 first to make token locations better and make it easier to get the values we'll be working with.

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -134,6 +134,7 @@ pub const Options = packed struct {
     @"variadic-macros": Kind = .default,
     varargs: Kind = .default,
     @"#warnings": Kind = .default,
+    @"deprecated-declarations": Kind = .default,
 };
 
 const messages = struct {
@@ -1568,6 +1569,28 @@ const messages = struct {
         const extra = .str;
         const opt = "ignored-attributes";
         const kind = .warning;
+    };
+    const deprecated_declarations = struct {
+        const msg = "{s}";
+        const extra = .str;
+        const opt = "deprecated-declarations";
+        const kind = .warning;
+    };
+    const deprecated_note = struct {
+        const msg = "'{s}' has been explicitly marked deprecated here";
+        const extra = .str;
+        const opt = "deprecated-declarations";
+        const kind = .note;
+    };
+    const unavailable = struct {
+        const msg = "{s}";
+        const extra = .str;
+        const kind = .@"error";
+    };
+    const unavailable_note = struct {
+        const msg = "'{s}' has been explicitly marked unavailable here";
+        const extra = .str;
+        const kind = .note;
     };
 };
 
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -41,7 +41,6 @@ pub const Message = struct {
         },
         attr_enum: struct {
             tag: Attribute.Tag,
-            actual: []const u8,
         },
         ignored_record_attr: struct {
             tag: Attribute.Tag,
@@ -1555,7 +1554,7 @@ const messages = struct {
         const extra = .attr_arg_type;
     };
     const unknown_attr_enum = struct {
-        const msg = "Unknown `{s}` argument {s}{s}{s}. Possible values are: {s}";
+        const msg = "Unknown `{s}` argument. Possible values are: {s}";
         const kind = .@"error";
         const extra = .attr_enum;
     };
@@ -1797,9 +1796,6 @@ pub fn renderExtra(comp: *Compilation, m: anytype) void {
                         .signed => m.print(info.msg, .{msg.extra.signed}),
                         .attr_enum => m.print(info.msg, .{
                             @tagName(msg.extra.attr_enum.tag),
-                            Attribute.Formatting.quoteChar(msg.extra.attr_enum.tag),
-                            msg.extra.attr_enum.actual,
-                            Attribute.Formatting.quoteChar(msg.extra.attr_enum.tag),
                             Attribute.Formatting.choices(msg.extra.attr_enum.tag),
                         }),
                         .ignored_record_attr => m.print(info.msg, .{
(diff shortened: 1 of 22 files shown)


Vexu:
This looks to be going in a pre #147 diagnostics direction, while we should probably be using a more declarative approach. Attribute needs to support GNU style `__attribute__`s, C2x `[[attributes]]` and likely also MSVC `__declspec`.

Argument parsing should be quite easy to do based on list of types, the only one I can think of that would actually require the node to be stored is `aligned` which could be represented as `struct { node: NodeIndex, alignment: u29 }`.

Then the issue with getting the tokens and nodes and #43 could be solved by handling attribute parsing in this order:
1. Parse attribute and arguments, validating them at the same time.
2. Store the result in a attributes list along with token for source location.
3. Once needed, check that attribute is valid in current context before creating the attributed type.

Further improvements include:
* Combining the two places where alignment is validated and removing the alignment field on `Type` in favor of storing it in attributes.
* Adding a `getAttribute(comptime: tag) ?tag.Args` to `Type` for easier consumption of attributes.

ehaas:
Declarative sounds good - after doing `aligned` this way I can see how it's not going to scale in terms of effort. The only reason I started out this way is that clang seems to do it all imperatively so I figured they knew something I didn't.

> Parse attribute and arguments, validating them at the same time.

I'm not sure we can always 100% validate the arguments while parsing:
```c
int __attribute__((aligned(64))) x;
int __attribute__((aligned(64))) _Thread_local y;
```
```
➜  ~ clang -target x86_64-scei-ps4 test.c
test.c:2:48: error: alignment (64) of thread-local variable 'y' is greater than the maximum supported alignment (32) for a thread-local variable on this target
int __attribute__((aligned(64))) _Thread_local y;
                                               ^
1 error generated.
```
Granted this is probably an uncommon edge case but it does indicate that attribute validity will sometimes depend on future tokens. For something like that I imagine we'd need some attribute validation in `DeclSpec.validate`?

In terms of the data structures, is this something like what you were thinking?

```zig
const aligned = struct {
    const namespace = "gnu";
    const Args = struct {
        alignment: ?struct {
            node: NodeIndex,
            requested: u29,
        }
    }
}

const alloc_size = struct {
    const namespace = "gnu";
    const Args = struct {
        position_1: u32,
        position_2: ?u32,
    }
}
// etc
```

And then at comptime we make a big union of all the `Args` which (along with a `TokenIndex`) will be able to hold any attribute value?

ehaas:
Alternatively we could have something like 
```zig
const AttributeArg = struct {
    tok: TokenIndex,
    val: union {
        none: void,
        str: []const u8,
        unsigned: u32,
        ident: void, // `tok` contains token index of identifier
        // separate fields below for each unique string enum and int enum?
    }
}
```
And we store a slice of these, one for each argument (active field determined by attribute name + position)

Vexu:
> I'm not sure we can always 100% validate the arguments while parsing:

We can't validate that `aligned(64)` is a valid attribute for `y` but we can validate that `64` is a valid argument to `aligned`.

> For something like that I imagine we'd need some attribute validation in `DeclSpec.validate`?

Sure, it should happen in the same place where the context is validated.

> In terms of the data structures, is this something like what you were thinking?

Something like that:
 ```zig
const aligned = struct {
    const gnu = "aligned";
    const std_gnu = "aligned"; // [[gnu::aligned]]
    const declspec = "aligned";
    const Args = struct {
        // put this type in a separate declaration since we need to reference it
        alignment: ?struct {
            node: NodeIndex,
            requested: u29,
        }
     }
};
```

> And then at comptime we make a big union of all the `Args` which (along with a `TokenIndex`) will be able to hold any ~attribute value~ set of attribute arguments?

Yes, though we probably won't need the token after validation.

**Comment by ehaas** [comment]
Cool, I think this all makes sense. I probably won't be able to work on it very much until January but I'd be excited to tackle it - I don't have much experience doing metaprogramming with Zig so this sounds like a fun learning experience.

**Comment by Vexu** [comment]
This is looking good so far, I'll try to fix that macos test failure later this week.

**Comment by ehaas** [comment]
> This is looking good so far, I'll try to fix that macos test failure later this week.

Not sure if this is useful info but I develop on MacOS using Zig master and don't get any test failures; however I haven't upgraded to Monterey yet.

**Comment by Vexu** [comment]
I fixed the value thingy and this should now also pass CI on MacOS. If you want we could probably easily get this to a mergeable state so that the remaining issues could be worked by others too.

**Comment by ehaas** [comment]
Cool, I have some local updates for alignment parsing that I can push later today

File: src/Diagnostics.zig [hunk 1]
```diff
@@ -30,6 +31,22 @@ pub const Message = struct {
...
         },
+        attr_arg_count: struct {
+            attribute: Attribute.Tag,
+            expected: u32,
+        },
+        attr_arg_type: struct {
+            expected: Attribute.ArgumentType,
+            actual: Attribute.ArgumentType,
+        },
+        attr_enum: struct {
```
ehaas, line 42: This union field is too big - I think it would be tricky to format the string at the callsite, so how would you feel about excluding it from the error message? We would still retain the correct caret position e.g. the difference between:

```
./test.c:1:25: error: Unknown `mode` argument 'bad'. Possible values are: 'byte', '__byte__', 'word', '__word__', 'pointer', '__pointer__'
int __attribute__((mode(bad))) foo;
                        ^
```

and

```
./test.c:1:25: error: Unknown `mode` argument. Possible values are: 'byte', '__byte__', 'word', '__word__', 'pointer', '__pointer__'
int __attribute__((mode(bad))) foo;
                        ^
```

Vexu: Fine by me, Clang/GCC only give the name of the bad argument. Also why are the normalized values also listed as enum members?

ehaas: You're right, the double underscore versions should not be in the enum and `diagnoseIdent` should normalize the name. There are a bunch of other values that should be in the enum: https://gcc.gnu.org/onlinedocs/gccint/Machine-Modes.html (for some reason that page lists them as e.g. `QImode ` but the actual value is `QI`)

Vexu:
This is good to go but I messed up rebasing and dropped one commit which I now can't find in git reflog. Could you force push your local changes back here (hopefully you still have them)?

ehaas:
Rebased against latest master and pushed

PR #223, opened 2022-01-26
Compilation: perform newline splicing on input files

### LINKED ISSUE #210: Crash: stringify multi-line string
I believe this is different from #135 (which I couldn't reproduce on latest master) 

The following code causes a crash during compilation; it stringifies incorrectly in the preprocessor
which is probably why:

```c
#define str(s) #s
void foo(void) {
   str("\
");
}
```

`clang -E` output:
```
void foo(void) {
   "\"\"";

}
```

`arocc -E` output:
```
void foo(void) {
   "\"\\;
}
```

During compilation it crashes:
```
thread 38947802 panic: index out of bounds
/Users/ehaas/source/arocc/src/Parser.zig:5919:34: 0x10effc574 in Parser.stringLiteral (arocc)
                    switch (slice[i]) {
                                 ^
/Users/ehaas/source/arocc/src/Parser.zig:5793:34: 0x10f12c94c in Parser.primaryExpr (arocc)
        => return p.stringLiteral(),
```

ehaas:
Newline splicing happens in "phase 2" of translation, before everything
except character set mapping and trigraph replacement

Fixes #210


ehaas:
Note: [This stackoverflow answer](https://stackoverflow.com/a/65107803) indicates that this approach will not match the behavior of clang/gcc with regard to `__LINE__` if a file has line splicings (gcc/clang preserve the original pre-splice line number; this approach will produce the post-splice line number)

Vexu (approved):
Only real issue with this is that it doesn't support `\r\n` (and just `\r`) newlines. It would also be nice to allow whitespace before the newline and give a warning about it like gcc and clang.

A future improvement would be for the Tokenizer to keep track of the spliced lines since it feels pretty weird to store a line that isn't actually meaningful in any way.

ehaas:
> Only real issue with this is that it doesn't support \r\n (and just \r) newlines. It would also be nice to allow whitespace before the newline and give a warning about it like gcc and clang.
> Also the relevant part from the Tokenizer should now be deleted as dead code.


I can make those changes. I could also incorporate the utf8 validation in the loop if you think that's a good idea.

> A future improvement would be for the Tokenizer to keep track of the spliced lines since it feels pretty weird to store a line that isn't actually meaningful in any way.

Maybe I'm misunderstanding but that seems like a much more invasive change, since anything that uses token slicing would need to account for embedded spliced newlines, right?

Vexu:
> > Only real issue with this is that it doesn't support \r\n (and just \r) newlines. It would also be nice to allow whitespace before the newline and give a warning about it like gcc and clang.
> > Also the relevant part from the Tokenizer should now be deleted as dead code.
> 
> I can make those changes. I could also incorporate the utf8 validation in the loop if you think that's a good idea.

That does sound like a good idea, less looping over the same bytes is always good.

> > A future improvement would be for the Tokenizer to keep track of the spliced lines since it feels pretty weird to store a line that isn't actually meaningful in any way.
> 
> Maybe I'm misunderstanding but that seems like a much more invasive change, since anything that uses token slicing would need to account for embedded spliced newlines, right?

Slicing uses `Token.byte_offset`, `Token.line` is pretty much only used for diagnostics and `__LINE__`. But this is more of an enhancement to the API and not necessary for this pr since I expect this isn't going to be quite as straightforward to implement.

Alternatively `Token.line` could be renamed to `Token._line` with a comment explaining that it's not the actual line as in the file and to use `Token.line()` for that.

ehaas:
Currently the line number and context for the `backslash_newline_escape` warning are incorrect; I think this warning will have to be handled as a special case, since it needs to print the buffer as it existed before splicing occurred.

UTF-8 validation is still a separate loop over the bytes; I figured I could add that as an enhancement if this looks good so far.

PR #225, opened 2022-01-31
Preprocessor: ensure expansion_locs of trimmed tokens are freed.

### LINKED ISSUE #159: Memory leak in preprocessor
The test `macro unbalanced parens call` was disabled in #151 due to one of the tokens being leaked.

ehaas:
Fixes #159


Vexu:
Awesome, we now have zero known memory leaks/crashes!

PR #232, opened 2022-02-04
Parser: handle casting of string literals

### LINKED ISSUE #228: Crash: cast string literal to char before integer promotion
```c
void foo(void) {
    int x = (char)"abc";
}
```
```
thread 7710351 panic: access of inactive union field
/Users/ehaas/source/arocc/src/Value.zig:163:42: 0x10858d3ae in Value.getInt (arocc)
        @truncate(T, @bitCast(i64, v.data.int));
                                         ^
/Users/ehaas/source/arocc/src/Value.zig:111:50: 0x10863af56 in Value.intCast (arocc)
            4 => v.* = int(@bitCast(u32, v.getInt(i32))),
                                                 ^
/Users/ehaas/source/arocc/src/Parser.zig:4217:28: 0x108626c48 in Parser.Result.intCast (arocc)
            res.val.intCast(res.ty, int_ty, p.pp.comp);
```

ehaas:
Fixes #228
Print correct type in an error message about casting

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -4860,7 +4860,7 @@ fn castExpr(p: *Parser) Error!Result {
             const new_float = ty.isFloat();
 
             if (new_float and operand.ty.isPtr()) {
-                try p.errStr(.invalid_cast_to_float, l_paren, try p.typeStr(operand.ty));
+                try p.errStr(.invalid_cast_to_float, l_paren, try p.typeStr(ty));
                 return error.ParsingFailed;
             } else if (old_float and ty.isPtr()) {
                 try p.errStr(.invalid_cast_to_pointer, l_paren, try p.typeStr(operand.ty));
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -4856,6 +4856,8 @@ fn castExpr(p: *Parser) Error!Result {
             // everything can cast to void
             operand.val.tag = .unavailable;
         } else if (ty.isInt() or ty.isFloat() or ty.isPtr()) cast: {
+            try operand.lvalConversion(p);
+
             const old_float = operand.ty.isFloat();
             const new_float = ty.isFloat();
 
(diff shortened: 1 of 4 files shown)


ehaas:
Are we missing an `.pointer_to_int` operation when explicitly casting a pointer to an integral type? It's sort of implied by the operation but I wasn't sure if we should include an explicit node for it.

Vexu:
> Are we missing a `.pointer_to_int` operation when explicitly casting a pointer to an integral type? It's sort of implied by the operation but I wasn't sure if we should include an explicit node for it.

As in when doing `(long)a_ptr`? I don't think we need to add any implicit nodes here but we could try to include the kind of the cast so users of the AST do not need to determine it.

ehaas:
Are you thinking something like a `kind` field for `cast_expr` nodes, similar to how clang does it? We could have `implicit_cast` and `explicit_cast` nodes, each of which stores a kind e.g. `int_to_pointer`, `pointer_to_bool`, etc.

Vexu:
> Are you thinking something like a `kind` field for `cast_expr` nodes, similar to how clang does it? We could have `implicit_cast` and `explicit_cast` nodes, each of which stores a kind e.g. `int_to_pointer`, `pointer_to_bool`, etc.

That sounds good, should simplify the backend too.

src/Parser.zig
```diff
@@ -4858,9 +4858,14 @@ fn castExpr(p: *Parser) Error!Result {
         } else if (ty.isInt() or ty.isFloat() or ty.isPtr()) cast: {
             const old_float = operand.ty.isFloat();
             const new_float = ty.isFloat();
+            if (operand.ty.isArray()) {
+                operand.ty.decayArray();
+                try operand.un(p, .array_to_pointer);
+                operand.val.tag = .unavailable;
+            }
```
Vexu, line 4865: I think this should be just `try operand.lvalConversion(p);`.

PR #241, opened 2022-02-09
Parser: set results to unavailable if used as part of an invalid binary expression

### LINKED ISSUE #229: Crash: chained comparison with illegal value
```c
void foo(void) {
    1.0 == &1 == 2;
}
```

```
thread 7715096 panic: reached unreachable code
/Users/ehaas/source/zig/build/lib/zig/std/debug.zig:234:14: 0x10d8e44d8 in std.debug.assert (arocc)
    if (!ok) unreachable; // assertion failure
             ^
/Users/ehaas/source/arocc/src/Value.zig:386:11: 0x10d8efe20 in Value.compare (arocc)
    assert(a.tag == b.tag);
          ^
/Users/ehaas/source/arocc/src/Parser.zig:4689:40: 0x10dab2490 in Parser.eqExpr (arocc)
            const res = lhs.val.compare(op, rhs.val, lhs.ty, p.pp.comp);
                                       ^
/Users/ehaas/source/arocc/src/Parser.zig:4662:27: 0x10dab02e4 in Parser.andExpr (arocc)
    var lhs = try p.eqExpr();
```

ehaas:
Fixes #229

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -4293,6 +4293,8 @@ const Result = struct {
 
     fn invalidBinTy(a: *Result, tok: TokenIndex, b: *Result, p: *Parser) Error!bool {
         try p.errStr(.invalid_bin_types, tok, try p.typePairStr(a.ty, b.ty));
+        a.val.tag = .unavailable;
+        b.val.tag = .unavailable;
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


ehaas:
I'm not totally sure if this is the correct solution; what was happening is that `adjustTypes` was returning false due to an error condition and not actually adjusting the types, which allowed the `ty` and `val` fields of the `Result` to get out of sync. I think another solution would be to return `error.ParsingFailed` from `invalidBinTy`

PR #244, opened 2022-02-16
Tree: store cast kind as part of implicit/explicit cast nodes

ehaas:
Briefly discussed here: https://github.com/Vexu/arocc/pull/232#issuecomment-1030530859

If you like this approach I can fill out the `fromExplicitCast` function.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 *.zig text eol=lf
 test/cases/expanded/* text eol=lf
+test/cases/ast/* text eol=lf 
 
 deps/** linguist-vendored
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -3965,14 +3965,6 @@ const Result = struct {
         });
     }
 
-    fn qualCast(res: *Result, p: *Parser, elem_ty: *Type) Error!void {
-        res.ty = .{
-            .data = .{ .sub_type = elem_ty },
-            .specifier = .pointer,
-        };
-        try res.implicitCast(p, .qual_cast);
-    }
-
     fn adjustCondExprPtrs(a: *Result, tok: TokenIndex, b: *Result, p: *Parser) !bool {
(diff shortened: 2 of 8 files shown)


Vexu:
Looks good. I'd prefer if explicit casts were dumped the same way as implicit ones (cast kind, no "operand:").

>If you like this approach I can fill out the `fromExplicitCast` function.

It'd probably be best to unify that, the various if's in `castExpr` and all the `T1ToT2` functions in `Result` into one `castType` function.

ehaas:
I was thinking it might make sense to add some testing around the AST layout before I try to have `implicitCast` and the initialization code use `castType` - what would you think of something like a directory called `ast` in `test/cases` that contains the expected AST for the corresponding file of the same name in `test/cases` (basically, like `expanded` but for the AST). If the file exists, the test runner would check that the parsed AST matches what is expected.

Vexu:
> what would you think of something like a directory called `ast` in `test/cases` that contains the expected AST for the corresponding file of the same name in `test/cases`

Sounds good to me. I'm guessing the plan is to do a string comparison on the output of `Tree.dump`?

ehaas:
I was thinking just strings compare initially; perhaps something more sophisticated if minor formatting changes make the strings too annoying

Vexu:
Test failure caused by Git messing with the line ends, fixed by adding `test/cases/ast/* text eol=lf` to `.gitattributes`.

Vexu:
Is this blocked on something?

ehaas:
I think I set it aside to work on some other things I needed to complete it and then forgot about it - I can look into resolving the conflicts

ehaas:
There's still the potential cleanup for `castExpr` / `T1ToT2` but I think that can be done as a separate PR if this one looks good; it should be easier to test now that we can make sure the AST is correct

PR #245, opened 2022-03-01
Type: handle decayed attributed arrays

### LINKED ISSUE #237: Crash: decayed attributed array
I believe we might need a `decayed_attributed` type specifier - currently, decaying an attributed array turns it into a `.special_va_start`

```c
__attribute__((aligned)) int foo[1] = {0};
int *bar = foo;
```

```
thread 9698834 panic: reached unreachable code
/Users/ehaas/source/arocc/src/Type.zig:1427:21: 0x10399d5df in Type.Builder.fromType (arocc)
            else => unreachable,
                    ^
/Users/ehaas/source/arocc/src/Type.zig:1515:48: 0x10396060c in Type.printPrologue (arocc)
        else => try w.writeAll(Builder.fromType(ty).str().?),
                                               ^
/Users/ehaas/source/arocc/src/Type.zig:1448:29: 0x10395f8fa in Type.print (arocc)
    _ = try ty.printPrologue(w);
```

ehaas:
Fixes #237

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -883,7 +883,7 @@ fn decl(p: *Parser) Error!bool {
         // Collect old style parameter declarations.
         if (init_d.d.old_style_func != null) {
             const attrs = init_d.d.ty.getAttributes();
-            var base_ty = if (init_d.d.ty.specifier == .attributed) init_d.d.ty.elemType() else init_d.d.ty;
+            var base_ty = if (init_d.d.ty.specifier == .attributed) init_d.d.ty.data.attributed.base else init_d.d.ty;
             base_ty.specifier = .func;
             init_d.d.ty = try base_ty.withAttributes(p.arena, attrs);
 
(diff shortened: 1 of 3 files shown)


ehaas:
To avoid the proliferation of `decayed_` types - should we just decay these to regular pointers?

Vexu:
I think I'd prefer that, doesn't this also apply to `typeof_type` and `typeof_expr`?

ehaas:
`typeof_expr` with an array decays before it gets to the `typeof`; `typeof_type` is currently parsing as "array-of-attributed-base-type" instead of "attributed-array of base-type" which makes the problem not manifest; the current behavior is wrong anyway since attributes should be ignored when parsing types

e.g. in clang:

```c
typeof(__attribute__((aligned))char []) x = {1};
```

produces:
```
test.c:1:23: warning: 'aligned' attribute ignored when parsing type [-Wignored-attributes]
typeof(__attribute__((aligned))char []) x = {1};
                      ^~~~~~~
```

ehaas:
Closing this one for now, going to consider an alternate approach

PR #249, opened 2022-03-03
Preprocessor: Handle potential integer underflow in macro expansion

### LINKED ISSUE #247: Crash: integer overflow due to nested unterminated function-like macro invocation
```c
#define str(s) # s
#define xstr(s) str(s)
#define INCFILE(n) str(strcmp(
xstr(INCFILE(2)) INCFILE(2))
```

```
thread 22572550 panic: integer overflow
/Users/ehaas/source/arocc/src/Preprocessor.zig:1286:69: 0x1053ec56a in Preprocessor.expandMacroExhaustive (arocc)
                    moving_end_idx = moving_end_idx + res.items.len - count;
                                                                    ^
/Users/ehaas/source/arocc/src/Preprocessor.zig:1332:33: 0x105386834 in Preprocessor.expandMacro (arocc)
    try pp.expandMacroExhaustive(tokenizer, &pp.top_expansion_buf, 0, 1, true);
                                ^
/Users/ehaas/source/arocc/src/Preprocessor.zig:393:35: 0x105351354 in Preprocessor.preprocessExtra (arocc)
                try pp.expandMacro(&tokenizer, tok);
                                  ^
/Users/ehaas/source/arocc/src/Preprocessor.zig:184:30: 0x10534c232 in Preprocessor.preprocess (arocc)
    return pp.preprocessExtra(source) catch |err| switch (err) {
```

ehaas:
Nested unterminated macro expansions could underflow `moving_end_idx`

Also addresses a memory leak encountered when eof occurs during
macro argument collection.

Fixes #247

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -1167,6 +1167,7 @@ fn collectMacroFuncArguments(
                 }
             },
             .eof => {
+                try args.append(curArgument.toOwnedSlice());
                 deinitMacroArguments(pp.comp.gpa, &args);
                 tokenizer.* = saved_tokenizer;
                 end_idx.* = old_end;
@@ -1281,9 +1282,8 @@ fn expandMacroExhaustive(
                     const count = macro_scan_idx - idx + 1;
                     for (buf.items[idx .. idx + count]) |tok| Token.free(tok.expansion_locs, pp.comp.gpa);
                     try buf.replaceRange(idx, count, res.items);
-                    // TODO: moving_end_idx += res.items.len - (macro_scan_idx-idx+1)
-                    // doesn't work when the RHS is negative (unsigned!)
-                    moving_end_idx = moving_end_idx + res.items.len - count;
+
+                    moving_end_idx = (moving_end_idx + res.items.len) -| count;
                     idx += res.items.len;
                     do_rescan = true;
                 } else {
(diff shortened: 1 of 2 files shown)


ehaas:
Two quick observations from fixing this:

1. The saturating arithmetic syntax is easy to remember and, when it makes sense to use it, it's nicer than an explicit `if` or `@xWithOverflow` 
2. The GPA in debug mode made it pretty straightforward to find the leak

src/Preprocessor.zig
```diff
@@ -1281,9 +1282,8 @@ fn expandMacroExhaustive(
                     const count = macro_scan_idx - idx + 1;
                     for (buf.items[idx .. idx + count]) |tok| Token.free(tok.expansion_locs, pp.comp.gpa);
                     try buf.replaceRange(idx, count, res.items);
-                    // TODO: moving_end_idx += res.items.len - (macro_scan_idx-idx+1)
-                    // doesn't work when the RHS is negative (unsigned!)
-                    moving_end_idx = moving_end_idx + res.items.len - count;
+
+                    moving_end_idx = (moving_end_idx + res.items.len) -| count;
```
Vexu, line 1286: This looks good but I want to document it so that I can understand it in the future as well.

My understanding is that `count` should be called `tokens_removed` and `res.items.len` should be called `tokens_added`.  So the easiest to understand version would be
```suggestion
                    moving_end_idx += tokens_added;
                    // Overflow here means that we encountered an unterminated argument list
                    // while expanding the body of this macro.
                    moving_end_idx -|= tokens_removed;
```

Vexu (approved):

PR #254, opened 2022-03-05
Tree: don't try to print values of type void

### LINKED ISSUE #252: Crash: AST dump of void-casted conditional operator branch
This seems to parse fine but crashes with `--verbose-ast`:
```c
void foo(void) {
    int x = 1;
    x ? (void)1 : 1;
}
```

```
thread 24492263 panic: reached unreachable code
/Users/ehaas/source/arocc/src/Value.zig:62:17: 0x10a133d52 in Value.signExtend (arocc)
        else => unreachable,
                ^
/Users/ehaas/source/arocc/src/Value.zig:439:46: 0x10a0cee9b in Value.dump (arocc)
            try w.print("{d}", .{v.signExtend(ty, comp)});
                                             ^
/Users/ehaas/source/arocc/src/Tree.zig:657:21: 0x10a085d51 in Tree.dumpNode (arocc)
        try val.dump(ty, tree.comp, w);
                    ^
/Users/ehaas/source/arocc/src/Tree.zig:849:30: 0x10a08d517 in Tree.dumpNode (arocc)
            try tree.dumpNode(tree.data[data.if3.body + 1], level + delta, w);
                             ^
```

ehaas:
Values can be implicitly cast to void if the other branch of a
conditional operator is void.

Fixes #252

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -4242,6 +4242,7 @@ const Result = struct {
                 .ty = res.ty,
                 .data = .{ .un = res.node },
             });
+            res.val.tag = .unavailable;
         }
     }
 
--- a/test/cases/casts.c
+++ b/test/cases/casts.c
@@ -10,6 +10,9 @@ void foo(void) {
     int a = (char)"foo";
     int b = (float)"foo";
     unsigned long long d = (unsigned long long)"foo";
+
+    int x = 1;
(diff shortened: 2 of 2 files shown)


Vexu:
It looks like the actual issue is `res.val.tag` not being set to `.unavailable` in `Result.toVoid`.

PR #260, opened 2022-03-12
Value: handle float to int overflow

### LINKED ISSUE #259: Crash: float overflow
```C
void foo(void) {
    int b = 10000000000.0f;
}
```

```
thread 27767647 panic: integer part of floating point value out of bounds
/Users/ehaas/source/arocc/src/Value.zig:80:14: 0x1005ba813 in Value.floatToInt (arocc)
        4 => @floatToInt(i32, v.getFloat(f32)),
             ^
/Users/ehaas/source/arocc/src/Parser.zig:4209:31: 0x1005bf3df in Parser.Result.intCast (arocc)
            res.val.floatToInt(res.ty, int_ty, p.pp.comp);
                              ^
/Users/ehaas/source/arocc/src/Parser.zig:3014:29: 0x1005b2a87 in Parser.coerceInit (arocc)
            try item.intCast(p, unqual_ty);
                            ^
/Users/ehaas/source/arocc/src/Parser.zig:2604:25: 0x10054f114 in Parser.initializer (arocc)
        try p.coerceInit(&res, tok, init_ty);
                        ^
```

ehaas:
Fixes #259

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -125,6 +125,9 @@ pub const Options = packed struct {
     @"gnu-case-range": Kind = .default,
     @"c++-compat": Kind = .default,
     vla: Kind = .default,
+    @"float-overflow-conversion": Kind = .default,
+    @"float-zero-conversion": Kind = .default,
+    @"float-conversion": Kind = .default,
 };
 
 const messages = struct {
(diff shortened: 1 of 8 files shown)


ehaas:
Blocked on https://github.com/ziglang/zig/pull/11116

ehaas:
As a second phase of this I was thinking I could add the various warnings related to float-to-int conversions. I think I'll need a `TokenIndex` in `Result.intCast` (in case `res.val.floatToInt` needs to issue a warning) - do you think `Result` should be augmented to include a `TokenIndex` of the first token of the result? Or should we pass the token into `intCast`?

Vexu:
> do you think `Result` should be augmented to include a `TokenIndex` of the first token of the result? Or should we pass the token into `intCast`?

If you think it might be needed in some other operation where it is hard to get any other way then it should be added to `Result` otherwise passing it to `intCast` directly is preferable.

ehaas:
Some of the warnings / options on them for some case aren't quite right; clang's behavior seems to vary depending on whether a standalone literal or an expression is used, and whether it's an initializer or an assignment.

PR #264, opened 2022-03-29
Const decl folding

ehaas:
Add supports for folding const int decls in certain situations where GCC/clang allow them.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1921,6 +1921,7 @@ fn recordDeclarator(p: *Parser) Error!bool {
         ty = try p.withAttributes(ty, attr_buf_top);
 
         if (p.eatToken(.colon)) |_| bits: {
+            const bits_tok = p.tok_i;
             const res = try p.constExpr();
             if (!ty.isInt()) {
                 try p.errStr(.non_int_bitfield, first_tok, try p.typeStr(ty));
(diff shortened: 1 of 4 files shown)


src/Parser.zig
```diff
@@ -5740,7 +5776,15 @@ fn primaryExpr(p: *Parser) Error!Result {
...
                     try p.checkDeprecatedUnavailable(s.ty, name_tok, s.name_tok);
+                    if (s.val.tag == .int) {
+                        switch (p.const_decl_folding) {
+                            .gnu_folding_extension => try p.errTok(.const_decl_folded, name_tok),
+                            .gnu_vla_folding_extension => try p.errTok(.const_decl_folded_vla, name_tok),
+                            else => {},
+                        }
+                    }
                     return Result{
+                        .val = if (p.const_decl_folding == .no_const_decl_folding) Value{} else s.val,
```
Vexu, line 5787: Shouldn't this also check `s.val.tag == .int` or is that implied by `.no_const_decl_folding`?

ehaas: `.no_const_decl_folding` is used for `_Static_assert`, `_Alignas`, and `__builtin_choose_expr`, and for all 3 of those the parser will subsequently issue an error if `val.tag != .int`. Regardless of the type of the decl, a `.unavailable` value should be safe, right?

Vexu: I phrased that question poorly, I meant that in the previous lines an error can be added depending on `p.const_decl_folding` but only if `s.val.tag == .int` and then as long as `p.const_decl_folding != .no_const_decl_folding` the value is used even if it's not `.int`.

ehaas: I see - with non-ints, the value will be used (by which I mean the parser will fold it into whatever expression it's currently parsing), but then the check for `if (!res.ty.isInt())` in `constExpr` will return `error.ParsingFailed` and prevent the parser from actually using the value in the larger expression being parsed.  So we don't want to issue the gnu extension warning in that case.

Example:
```c
const double d = 2.0;
enum E {
    VALUE = d,
};
```

```
➜  ~ arocc -Wgnu-folding-constant test.c
./test.c:3:13: error: expression is not an integer constant expression
    VALUE = d + 1,
            ^
```

Issue #74, opened 2021-11-03
Have test runner check for leaks caused by `error.OutOfMemory`


ehaas:
The [Zig parser test](https://github.com/ziglang/zig/blob/08dc840247536cffb970b9fa3974017db44ed373/lib/std/zig/parser_test.zig#L5306) uses a clever technique to check for correct memory handling during error conditions. Parse once using unlimited memory, measure how much was actually used, then parse again N times using `FailingAllocator`, once for each size up to the actual amount of memory used. Each of those should fail (if not, memory usage is unexpectedly non-deterministic) and should free the same number of bytes as were allocated.

PR #276, opened 2022-04-07
Test all allocation failures

ehaas:
Initial support for #74

Run with `zig build test -Dtest-all-allocation-failures=true` (currently finds some leaks)

Eventually `checkAllAllocationFailures` should be used per test-case, but for now was easier to just apply it to the whole test suite

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -33,6 +33,8 @@ pub fn build(b: *Builder) !void {
     // for restricting supported target set are available.
     const target = b.standardTargetOptions(.{});
 
+    const test_all_allocation_failures = b.option(bool, "test-all-allocation-failures", "Test all allocation failures") orelse false;
+
     // Standard release options allow the person running `zig build` to select
     // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall.
     const mode = b.standardReleaseOptions();
@@ -71,6 +73,9 @@ pub fn build(b: *Builder) !void {
         .path = .{ .path = "src/lib.zig" },
         .dependencies = &.{zig_pkg},
     });
+    const test_runner_options = b.addOptions();
+    integration_tests.addOptions("build_options", test_runner_options);
+    test_runner_options.addOption(bool, "test_all_allocation_failures", test_all_allocation_failures);
 
     const integration_test_runner = integration_tests.run();
     integration_test_runner.addArg(b.pathFromRoot("test/cases"));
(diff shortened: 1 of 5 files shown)


src/Compilation.zig
```diff
@@ -509,13 +509,14 @@ pub fn getSource(comp: *Compilation, id: Source.Id) Source {
...
 /// line-ending normalization (convert line endings to \n), and UTF-8 validation.
+/// `contents` will be reallocated if newline-splicing occurs
 /// Creates a Source with `contents` as the buf and adds it to the Compilation.
 /// `contents` is assumed to be large enough to hold the entire content of `reader`.
 /// `contents` must have been allocated by `comp`'s allocator since it will be reallocated
 /// if splicing occurred.
 /// Compilation owns `contents` if and only if this call succeeds; caller always retains
 /// ownership of `path`.
-pub fn addSourceFromReader(comp: *Compilation, reader: anytype, path: []const u8, contents: []u8) !Source {
+pub fn addSourceFromReader(comp: *Compilation, reader: anytype, path: []const u8, contents: *[]u8) !Source {
```
Vexu, line 519: I feel like this change could be simplified by passing in the expected size and letting `addSourceFromReader` handle allocations. Is there any use case that couldn't handle?

ehaas: I like that solution better, I've never been too satisfied with the signature of this function or the extensive doc comment to begin with. Should I do that as a separate PR against master? Not sure how long it will take to get all the allocation failure leaks ironed out.

Vexu: A separate pr would be fine.

ehaas:
The last OutOfMemory-related leak is that sometimes Tokens can leak `expansion_locs` if an intermediate allocation fails.

ehaas:
I'm having trouble tracking down the last set of out-of-memory leaks related to the Token expansion locations. Seems like we have a couple options:

1. Merge the existing low-hanging fruit and continue trying to figure out the remaining leaks
2. Abandon the whole project
3. Use an arena for expansion tokens (downside is always using more memory for something that is only relevant for diagnostics, and the leak only happens in out-of-memory situations to begin with)

Vexu:
1 sounds good to me, I'll try to look into the leak too at some point. If it ever becomes an issue in a real world situation then we can use 3 as a temporary solution.

ehaas:
I think at least one set of issues is `for` loops that operate on a slice of tokens (e.g. `bufCopyTokens`) and add expansion locations, and partially succeed. In that case I don't think we free the expansion locations of tokens that were successfully duped.

PR #277, opened 2022-04-08
Parser: don't save returnStmt result

### LINKED ISSUE #272: Crash: invalid type specifier combination in function definition
The following crashes when using `--verbose-ast`:

```c
char int foo(void) {
    return 0;
}
```

```
thread 41735095 panic: reached unreachable code
/Users/ehaas/source/arocc/src/Value.zig:62:17: 0x100de85b2 in Value.signExtend (arocc)
        else => unreachable,
                ^
/Users/ehaas/source/arocc/src/Value.zig:489:46: 0x100d8685b in Value.dump (arocc)
            try w.print("{d}", .{v.signExtend(ty, comp)});
                                             ^
/Users/ehaas/source/arocc/src/Tree.zig:657:21: 0x100d37d81 in Tree.dumpNode (arocc)
        try val.dump(ty, tree.comp, w);
                    ^
/Users/ehaas/source/arocc/src/Tree.zig:950:34: 0x100d401e0 in Tree.dumpNode (arocc)
                try tree.dumpNode(data.un, level + delta, w);
                                 ^
```

ehaas:
Fixes #272


Vexu:
It would be nice to have that `saveValue` there for return statements with more complex constant expressions. One way to archive that would be to copy the result as is done in `constExpr` or add a new `saveValueExtra` which doesn't invalidate the result value and implement `saveValue` with that.

ehaas:
Did you see my follow-up comment here https://github.com/Vexu/arocc/issues/272#issuecomment-1093076303

The `signExtend` function in Value.zig assumes that integer values will be promoted to int, which doesn't happen for char return values. So if `signExtend` is allowed to operate on 1 and 2 byte types that would also solve the problem without having to change the parser.

Vexu:
I did see that but didn't realize `signExtend` was the problem. Thinking about it again, copying the value doesn't make any sense since it isn't used after the return statement.

>So if `signExtend` is allowed to operate on 1 and 2 byte types that would also solve the problem without having to change the parser.

That seems reasonable, the function was written with mostly expressions in mind in which the assumption holds true.

ehaas:
Just to clarify - should I add 1 and 2-byte integer support to Value.signExtend? And keep the change that removes the `saveValue` for return statements? Or should I revert that part.

Vexu:
Just the `signExtend` part, saving the return value is desired.

PR #278, opened 2022-04-15
Compilation: refactor addSourceFromReader

ehaas:
Refactor addSourceFromReader to take a size argument instead of a
pre-allocated buffer. This makes it easier to clean up resources in
the case of a memory allocation failure. Adds a test to ensure that
an allocation failure in addSourceFromReader will not leak memory.


Vexu:
Thanks, this feels like a much better API.

PR #279, opened 2022-04-20
Add -Wundef support

ehaas:


src/Preprocessor.zig
```diff
@@ -560,14 +559,19 @@ fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) MacroError!bool {
...
                     .tag = .invalid_preproc_operator,
                     .loc = tok.loc,
                 }, tok.expansionSlice());
                 return false;
             },
             else => if (id.isMacroIdentifier()) {
+                try pp.comp.diag.add(.{
+                    .tag = .undefined_macro,
+                    .loc = tok.loc,
+                    .extra = .{ .str = pp.expandedSlice(tok) },
```
ehaas, line 572: Is it safe to use `pp.expandedSlice(tok)` here, or do I need to dupe it into comp.diag's arena allocator?

Vexu: `generated_buf` was moved to `Compilation` so it should live as long as `Diagnostics`.

PR #280, opened 2022-04-24
Parser: Do not allow incomplete-size fields in anonymous records.

### LINKED ISSUE #275: Segfault: incomplete-size field in nested anonymous struct
```c
struct Foo {
    struct {
        struct Bar bar;
    };
};
```

```
Segmentation fault at address 0x0
???:?:?: 0x7fff204c404d in ??? (???)
/Users/ehaas/source/arocc/src/Parser.zig:147:38: 0x10931259c in Parser.struct:126:9.addFieldsFromAnonymous (arocc)
        for (ty.data.record.fields) |f| {
                                     ^
/Users/ehaas/source/arocc/src/Parser.zig:1815:52: 0x109311339 in Parser.recordDeclarator (arocc)
                try p.record.addFieldsFromAnonymous(p, ty);
                                                   ^
/Users/ehaas/source/arocc/src/Parser.zig:598:16: 0x1092fe002 in Parser.parseOrNextDecl (arocc)
    return func(p) catch |er| switch (er) {
               ^
```

ehaas:
Fixes #275

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1798,7 +1798,7 @@ fn recordDeclarator(p: *Parser) Error!bool {
         }
 
         if (name_tok == 0 and bits_node == .none) unnamed: {
-            if (ty.is(.@"enum")) break :unnamed;
+            if (ty.is(.@"enum") or ty.hasIncompleteSize()) break :unnamed;
             if (ty.isAnonymousRecord()) {
                 // An anonymous record appears as indirect fields on the parent
                 try p.record_buf.append(.{


ehaas:
I'm not sure the best way to test this one - the enclosing anonymous record in #275 has an absolute path in its name, so it will be different on each system.

Vexu:
> I'm not sure the best way to test this one

It's pretty tricky, best I could come up with would be to trim the path away but that would only leave something like `.c:69:5)'`.

PR #282, opened 2022-05-10
Digraphs

ehaas:
Add digraph support
Add user macro support to test runner for command line args specified via `//aro-args`


ehaas:
When using the `-E` flag with gcc and clang, digraphs are not converted into the corresponding token.  I don't think we can support this unless we add new tokens for each digraph (`.l_brace_digraph` etc). Would you want me to do that?

Vexu:
> When using the -E flag with gcc and clang, digraphs are not converted into the corresponding token. I don't think we can support this unless we add new tokens for each digraph (.l_brace_digraph etc). Would you want me to do that?

This should be doable without adding any new tags by just adding a little more logic in `Preprocessor.prettyPrintTokens` to always find the slice in source for these tokens when digraphs are enabled.

PR #284, opened 2022-05-15
Preprocessor: Add __has_include support

### LINKED ISSUE #41: Preprocessor: implement __has_include
GCC docs: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

Clang docs: https://clang.llvm.org/docs/LanguageExtensions.html#has-include

ehaas:
Closes #41


ehaas:
You're right; I think during my testing I accidentally left in the 0-argument placeholder version and I was getting errors about invalid binary operator.

src/Preprocessor.zig
```diff
@@ -1785,11 +1865,15 @@ fn pragma(pp: *Preprocessor, tokenizer: *Tokenizer, pragma_tok: RawToken, operat
     }, pragma_name_tok.expansionSlice());
 }
 
-fn findIncludeSource(pp: *Preprocessor, tokenizer: *Tokenizer) !Source {
+fn findIncludeFilenameToken(
+    pp: *Preprocessor,
+    first: *RawToken,
```
Vexu, line 1870: Why is `first` passed as a pointer?

src/Preprocessor.zig
```diff
@@ -851,6 +856,66 @@ fn stringify(pp: *Preprocessor, tokens: []const Token) !void {
...
 }
 
+fn reconstructIncludeString(pp: *Preprocessor, param_toks: []const Token) !?[]const u8 {
+    const char_top = pp.char_buf.items.len;
+    defer pp.char_buf.items.len = char_top;
+    var params = param_toks;
+
+    // Trim leading/trailing whitespace
+    if (params[0].id == .macro_ws) params = params[1..];
+    if (params.len > 0 and params[params.len - 1].id == .macro_ws) params = params[0 .. params.len - 1];
```
Vexu, line 866: `macro_ws` being non consecutive is not guaranteed, it might be true but I haven't checked it, better to just ignore them here and trim at the end.

src/Preprocessor.zig
ehaas: That's probably a holdover from how the code was original constructed; it looks like I can just pass the RawToken in and make a copy of it in the function since the `macro_string` mutation doesn't seem like it needs to escape this function.

src/Preprocessor.zig
ehaas: I ended up adjusting the trimming code to operate on tokens instead of the string since that makes it easier to detect multiple arguments outside of `<>` (e.g. pasted strings) and issue an error.

PR #290, opened 2022-06-02
Compilation: fix use-after-free in path->source map

tomc1998:


Vexu:
Thanks! Surprised this didn't show up earlier, I remember fixing this same bug and it was causing files to be loaded multiple times.

tomc1998:
Haha, yes that's what clued me into it. Possibly the `pragma once.c` test case is insufficient, since it only tests when files are included twice from the *same* file, i'm guessing this doesn't cause the test to fail for some reason (?)

In my case, 2 *different* files were including stddef.h

PR #291, opened 2022-06-03
Parser: Compute sizeof / alignof struct and union types

tomc1998:
Compute sizeof/alignof records.

This isn't a perfect solution - I'm unsure if it works identically to other compilers in all cases & it definitely doesn't factor in packed attributes. But it's good enough - I wanted to load `netinet/in.h`, which fails with the following error otherwise:

```
Compiler error:
/usr/include/netinet/in.h:244:53: warning: overflow in expression; result is '18446744073709551615' [-Winteger-overflow]
    unsigned char sin_zero[sizeof (struct sockaddr) -
                                                    ^
/usr/include/netinet/in.h:244:27: error: array is too large
    unsigned char sin_zero[sizeof (struct sockaddr) -


// Because of this expression:
unsigned char sin_zero[sizeof (struct sockaddr) -
                        __SOCKADDR_COMMON_SIZE -
                        sizeof (in_port_t) -
                        sizeof (struct in_addr)];
```

Two questions: 
1. How can i report a compiler error? see line 1711. I poked around Compilation.zig but nothing jumped out.
2. The test is obviously flawed, I just matched whatever GCC produced on my system. I can either remove the test, or maybe there's a way to skip it on non-x64-linux systems?

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1717,6 +1717,13 @@ fn recordSpec(p: *Parser) Error!Type {
         }
     }
 
+    if (p.record_buf.items.len == record_buf_top) {
+        try p.errStr(.empty_record, kind_tok, p.tokSlice(kind_tok));
+        try p.errStr(.empty_record_size, kind_tok, p.tokSlice(kind_tok));
+    }
+    try p.expectClosing(l_brace, .r_brace);
+    try p.attributeSpecifier(); // .record
+
     for (p.record_buf.items[record_buf_top..]) |field| {
         if (field.ty.hasIncompleteSize()) break;
     } else {
@@ -1736,7 +1743,7 @@ fn recordSpec(p: *Parser) Error!Type {
             for (record_ty.fields) |f| {
                 const field_size_opt = f.ty.sizeof(p.pp.comp);
                 if (field_size_opt) |field_size| {
-                    const field_align = f.ty.alignof(p.pp.comp);
+                    const field_align = f.ty.fieldAlignof(p.pp.comp);
                     largest_field_size = std.math.max(field_size, largest_field_size);
                     largest_field_align = std.math.max(field_align, largest_field_align);
                     const offset_needed = @divTrunc(total_size + (field_align - 1), field_align) * field_align;
@@ -1765,13 +1772,6 @@ fn recordSpec(p: *Parser) Error!Type {
             record_ty.alignment = largest_field_align;
         }
     }
-
-    if (p.record_buf.items.len == record_buf_top) {
-        try p.errStr(.empty_record, kind_tok, p.tokSlice(kind_tok));
-        try p.errStr(.empty_record_size, kind_tok, p.tokSlice(kind_tok));
-    }
-    try p.expectClosing(l_brace, .r_brace);
-    try p.attributeSpecifier(); // .record
 
     ty = try p.withAttributes(.{
         .specifier = if (is_struct) .@"struct" else .@"union",
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1721,9 +1721,49 @@ fn recordSpec(p: *Parser) Error!Type {
         if (field.ty.hasIncompleteSize()) break;
     } else {
         record_ty.fields = try p.arena.dupe(Type.Record.Field, p.record_buf.items[record_buf_top..]);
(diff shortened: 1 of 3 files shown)


src/Parser.zig
```diff
@@ -1686,9 +1686,49 @@ fn recordSpec(p: *Parser) Error!*Type.Record {
         if (field.ty.hasIncompleteSize()) break;
     } else {
         record_ty.fields = try p.arena.dupe(Type.Record.Field, p.record_buf.items[record_buf_top..]);
-        // TODO actually calculate
-        record_ty.size = 1;
-        record_ty.alignment = 1;
+        // Compute the size of this record.
+        if (record_ty.fields.len == 0) {
+            // Alignment of 0 likely to cause division by 0 errors
+            // later, align of 1 is semantically the same i think
```
Vexu, line 1692: GCC and Clang say that these are the correct values:
```c
struct Foo {};
_Static_assert(_Alignof(struct Foo)==1,"");
_Static_assert(sizeof(struct Foo)==0,"");
```

Vexu (commented):
>1. How can i report a compiler error? see line 1711. I poked around Compilation.zig but nothing jumped out.

See b9e81fde724b1d50e5fe92890b558fe88a0c65d8 for a simple example on how to add a new diagnostics message.

>2. The test is obviously flawed, I just matched whatever GCC produced on my system. I can either remove the test, or maybe there's a way to skip it on non-x64-linux systems?

You can use `#if/#else/#endif` in the test cases just like you would in regular C code.

ehaas:
> How can i report a compiler error? see line 1711. I poked around Compilation.zig but nothing jumped out.

For this particular error - the `for` loop that this `else` belongs to verifies that every field in the record has a non-null size. So I think it's safe to use `.?` here.

ehaas:
@tomc1998 if you rebase this against latest master I fixed a couple bugs with attributes. If you move the entire `for` loop that computes the size/alignment to right before the node creation (https://github.com/Vexu/arocc/blob/master/src/Parser.zig#L1717=) you should also be able to grab the value of the  `aligned` attribute of the record with `ty.requestedAlignment(p.pp.comp)` and update the values if necessary (keep in mind the `aligned` attribute can only increase alignment, not decrease it: https://gcc.gnu.org/onlinedocs/gcc-3.3/gcc/Type-Attributes.html). 

Another wrinkle is that `aligned` attributes on record fields themselves, unlike on variables, follow the "may not decrease alignment" rule. `field.ty.alignof` will return the requested alignment, even if it's lower than the natural alignment of the type. So we probably need a `fieldAlignment` function on `Type` that returns the max of the natural alignment and the requested alignment.

```c
struct S1 {
    __attribute__((aligned(64))) int x;
};
_Static_assert(_Alignof(struct S1) == 64, "bad alignment");

struct S2 {
    __attribute__((aligned(1))) int y;
};
_Static_assert(_Alignof(struct S2) == _Alignof(int), "bad alignment");

int __attribute__((aligned(1))) z;
_Static_assert(_Alignof(z) == 1, "bad alignment");
```

Finally, for platform specific testing you can wrap your test in `#if defined(__linux__) && defined(__x86_64__)`

tomc1998:
Hello, sorry i've been busy, i'll fix these up tonight or tomorrow night + maybe have a look at alignment :)

TwoClocks:
I have a port of mankoh's rust repr-c (https://github.com/mahkoh/repr-c) going on this branch here (https://github.com/TwoClocks/arocc/blob/records/src/RecordLayout.zig). Something I said I'd work on 6 months ago, but just came back to a few weeks ago.
 
 I'm happy to abandon this work, or integrate it w/ this PR, or whatever is best.

Vexu:
This PR is a naive solution and not enough for replacing but it can unblock progress on other things while waiting for a complete solution.

tomc1998:
Hello, i didn't abandon this - promise :)

I've rebased with master & made the changes @ehaas suggested - so now we take into account struct field attribute alignment!

ehaas:
Superseded by #364 - @tomc1998 can you see if `netinet/in.h` parses correctly for you now?

Vexu:
With 7c179f6b9402b41e03bf60211bdecd183441d8f8 `#include "netinet/in.h"` now compiles without errors.

PR #298, opened 2022-06-16
Basic implementation of #include_next

ehaas:
This is a very basic implementation of `#include_next` that just uses exact string matches to advance the iterator past the directory of the current file.

The clang code for looking up includes has a lot going on: https://github.com/llvm/llvm-project/blob/264c09b732ab61936a499f133abdfc67df422a86/clang/lib/Lex/HeaderSearch.cpp#L836= which could be a fruitful source of ideas of how to improve things.

Not sure if these should be separate issues, but possible enhancements include: caching so that we're not repeatedly looking up nonexistent files, Apple Framework support, `O(1)` advancement of the `include_next` iterator, MSVC-style "include stacks", tracking whether the current file was included by quoted include / angled include / absolute path

### DIFF
--- a/test/cases/include/my_include.h
+++ b/test/cases/include/my_include.h
@@ -1,3 +1,4 @@
 #define FOO 1
 
 #include_next <my_include.h> // test/cases/include/next/my_include.h
+#include_next <other.h>  // test/cases/include/next/other.h 
--- a/test/cases/include/my_include.h
+++ b/test/cases/include/my_include.h
@@ -1,0 +1,3 @@
+#define FOO 1
+
+#include_next <my_include.h> // test/cases/include/next/my_include.h
--- a/test/cases/include/next/my_include.h
+++ b/test/cases/include/next/my_include.h
@@ -1,0 +1,1 @@
+#define BAR 2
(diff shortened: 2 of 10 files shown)


test/cases/include/next/my_include.h
```diff
@@ -0,0 +1 @@
+#define BAR 2
```
Vexu, line 1: If I understand correctly you should be able to add this snippet here
```c
#ifdef BAR
#error bad
#endif
```

test/cases/include/my_include.h
```diff
@@ -0,0 +1,3 @@
+#define FOO 1
+
```
Vexu, line 2: And this with a static assert in the main file here.
```c
#ifdef BAR
#define BAZ 1
#endif
```

Vexu (approved):
> Not sure if these should be separate issues, but possible enhancements include:

Either works for me, do what you think is best.

test/cases/include_next.c
```diff
@@ -3,6 +3,7 @@
 #include_next <stdalign.h>
```
Vexu, line 3: Oh I read that this was also including `my_include.h`, that suggestion doesn't make much sense without it, it was supposed to test that `#include` and `#include_next` do the same thing when used from the main file.

ehaas: I see, I can add a test for that behavior too

PR #308, opened 2022-06-20
Enum sizes

ehaas:
Calculate enum sizes, accounting for packed attributes and overflow

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -2002,7 +2002,7 @@ fn enumSpec(p: *Parser) Error!Type {
         p.enum_buf.items.len = enum_buf_top;
     }
 
-    var e = Enumerator.init(p);
+    var e = Enumerator{};
     while (try p.enumerator(&e)) |field_and_node| {
         try p.enum_buf.append(field_and_node.field);
         try p.list_buf.append(field_and_node.node);
@@ -2019,6 +2019,7 @@ fn enumSpec(p: *Parser) Error!Type {
         .{ .specifier = .@"enum", .data = .{ .@"enum" = enum_ty } },
         attr_buf_top,
     );
+    enum_ty.tag_ty = try e.tagType(p, ty.hasAttribute(.@"packed") or p.comp.langopts.short_enums, maybe_ident orelse enum_tok);
 
     // declare a symbol for the type
     if (maybe_ident != null and !defined) {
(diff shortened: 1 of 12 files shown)


src/Parser.zig
```diff
@@ -2076,6 +2087,40 @@ const Enumerator = struct {
         e.res = res;
         // TODO adjust res type to try to fit with the previous type
     }
+
+    fn tagSpec(e: *const Enumerator, p: *Parser, is_packed: bool, tok: TokenIndex) !Type.Specifier {
```
Vexu, line 2091: The `spec` in `enumSpec` and `recordSpec` comes from the official grammar, maybe it would be better to call this `getTagType` or even just `tagType`?

src/Parser.zig
```diff
@@ -2076,6 +2087,40 @@ const Enumerator = struct {
...
+        const char_width = (Type{ .specifier = .schar }).sizeof(p.comp).? * 8;
+        const short_width = (Type{ .specifier = .short }).sizeof(p.comp).? * 8;
+        const int_width = (Type{ .specifier = .int }).sizeof(p.comp).? * 8;
+        if (e.num_negative_bits > 0) {
```
Vexu, line 2095: I'm not sure if it matters currently but this should also probably include `or p.comp.langopts.emulate == .msvc or (p.comp.langopts.emulate == .clang and p.comp.target.os.tag == .windows)`. This might not be complete either but I'm assuming that mingw at least behaves like GCC.

Vexu (approved):

src/Parser.zig
ehaas: That's a good point; I'm not sure how clang currently handles that. This is what I currently get with `zig translate-c`:

```c
enum E { A };
```

**zig translate-c -target x86_64-windows-msvc**
```zig
pub const enum_E = c_int;
```

**zig translate-c -target x86_64-windows-gnu**
```zig
pub const enum_E = c_uint;
```

I can dig into the clang code a little bit and see if I can figure out how it decides that msvc should use an `int` even if there are no negative values. And if it's not obvious, we can work around it for now in that if statement?

ehaas: Ok so it seems like MSVC strictly follows the standard, so enums declared using the standard syntax are *always* `int`, even if enumerator constants are too large or `__attribute__((packed))` is set (can be verified using `zig translate-c`).

However MSVC (and clang) also supports a custom syntax for specifying the underlying int type of an enum:
```c
enum E : unsigned char { A };
```

**zig translate-c -target x86_64-linux-gnu test.c**
```zig
pub const enum_E = u8;
```

**zig translate-c -target x86_64-windows-msvc**
```zig
pub const enum_E = u8;
```

Vexu: > Ok so it seems like MSVC strictly follows the standard, so enums declared using the standard syntax are _always_ `int`, even if enumerator constants are too large or `__attribute__((packed))` is set (can be verified using `zig translate-c`).

Huh, didn't expect that.
 
> However MSVC (and clang) also supports a custom syntax for specifying the underlying int type of an enum:

That's a C++ thing, but based on quick testing on https://c.godbolt.org clang also supports it for C while MSVC and GCC only allow it for C++.

Vexu:
Thanks for this! This might still need some tweaking to match clang, on my machine clang says the enum in `enum overflow.c` has type `long long` instead of `unsigned long` and clang generally seems to prefer `long long` over `long` even if they're the same size.

Also sorry for hijacking your branch at the end. I assumed you were done with this thought to add the two missing parts so that this PR would get enums to a (mostly) complete state but they turned out to be more complex than I assumed.

ehaas:
Sounds good! I can take a look at the overflowed enums

PR #313, opened 2022-06-26
Parser:Parse record field attributes

ehaas:


ehaas:
Needs tests and some cleanup. 

My goal was to not do any allocations if none of the fields have attributes but maybe that makes the logic more complicated for not much gain. Also, do you think the `field_attributes` ArrayList should be a member of `Parser.record` or directly on `Parser`?

src/Parser.zig
```diff
@@ -1796,12 +1804,10 @@ fn recordDecls(p: *Parser) Error!void {
 fn recordDeclarator(p: *Parser) Error!bool {
     const attr_buf_top = p.attr_buf.len;
     defer p.attr_buf.len = attr_buf_top;
+    try p.attributeSpecifier(); // .record
     const base_ty = (try p.specQual()) orelse return false;
 
     while (true) {
-        const this_decl_top = p.attr_buf.len;
-        defer p.attr_buf.len = this_decl_top;
-
```
Vexu, line 1804: These are here on purpose, multiple declarators on the same declaration specifier do not share attributes:
```c
struct Foo {
    int a __attribute__((packed)), b; // only 'a' is packed
};
```

Vexu (changes requested):
> Also, do you think the `field_attributes` ArrayList should be a member of `Parser.record` or directly on `Parser`?

It seems to me like the memory management would be easier if it was directly on `Parser` since it would be used in much the same way as `attr_buf`.

ehaas:
Just need to add validation back in for field attributes.

Issue #316, opened 2022-06-27
MSVC typedefs on arrays errors.


TwoClocks:
I think following code is valid for MSVC
```c
typedef int unnamed_type_1[];
unnamed_type_1 var2;
```
for GCC/Clang it would need to be
```c
typedef int unnamed_type_1[0];
unnamed_type_1 var2;
```

the MSVC code gives this error:
```
arocc/test/records/0043_test.c:12:16: error: variable has incomplete type 'int []'
unnamed_type_1 var2;
               ^
```

`comp.langopts.enableMSExtensions();` is called before compiling the MSVC version.

I just sync'd w/ master. I don't think I was getting this error a few days ago.

PR #317, opened 2022-06-27
test suite for record layout

TwoClocks:
Just the tests for record layout.
At the moment, some issues are causing the MSVC version to fail.  #316
The GCC/Clang versions work.
At the moment, only the parser is being tested. No static asserts for layout are being called.
The code can test layout w/ a small change.
A `README.md` is in the `/test/records` directory that explains the `.c` usage files and license.


TwoClocks:
this has drifted too much from records branch.

Issue #315, opened 2022-06-27
Use string interning for `[]const u8` fields in Type sub-types (`Record`, `Enum`, etc)


ehaas:
The string interning from [AstGen.zig](https://github.com/ziglang/zig/blob/master/src/AstGen.zig) (`string_table` + `string_bytes`) seems like it would be a good fit for all the `name: []const u8` fields in Type.zig

PR #320, opened 2022-06-29
WIP: String interning for types

ehaas:
Initial pass at #315

If this approach looks good I can apply it to the other `[]const u8` fields in `Type.zig`.

One slightly annoying thing is juggling the `string_interner` in the test runner, to make sure it doesn't get de-inited. I could make the string interner not be owned by the Compilation, but I'm not sure that feels right either.

### DIFF
--- a/src/Type.zig
+++ b/src/Type.zig
@@ -99,8 +99,8 @@ pub const Func = struct {
     params: []Param,
 
     pub const Param = struct {
-        name: StringId,
         ty: Type,
+        name: StringId,
         name_tok: TokenIndex,
     };
 };
@@ -137,14 +137,14 @@ pub const Attributed = struct {
 
 // TODO improve memory usage
 pub const Enum = struct {
-    name: StringId,
-    tag_ty: Type,
     fields: []Field,
+    tag_ty: Type,
+    name: StringId,
     fixed: bool,
 
     pub const Field = struct {
-        name: StringId,
         ty: Type,
+        name: StringId,
         name_tok: TokenIndex,
         node: NodeIndex,
     };
@@ -165,19 +165,18 @@ pub const Enum = struct {
 
 // TODO improve memory usage
 pub const Record = struct {
-    name: StringId,
     fields: []Field,
     size: u64,
-    alignment: u29,
-
     /// If this is null, none of the fields have attributes
     /// Otherwise, it's a pointer to N items (where N == number of fields)
     /// and the item at index i is the attributes for the field at index i
     field_attributes: ?[*][]const Attribute,
+    name: StringId,
+    alignment: u29,
 
     pub const Field = struct {
-        name: StringId,
         ty: Type,
+        name: StringId,
         /// zero for anonymous fields
         name_tok: TokenIndex = 0,
         bit_width: u32 = 0,
--- a/src/Type.zig
+++ b/src/Type.zig
@@ -5,6 +5,8 @@ const NodeIndex = Tree.NodeIndex;
 const Parser = @import("Parser.zig");
 const Compilation = @import("Compilation.zig");
 const Attribute = @import("Attribute.zig");
+const StringInterner = @import("StringInterner.zig");
+const StringId = StringInterner.StringId;
 
 const Type = @This();
 
@@ -97,7 +99,7 @@ pub const Func = struct {
     params: []Param,
 
     pub const Param = struct {
-        name: []const u8,
+        name: StringId,
         ty: Type,
         name_tok: TokenIndex,
     };
@@ -135,13 +137,13 @@ pub const Attributed = struct {
 
 // TODO improve memory usage
 pub const Enum = struct {
-    name: []const u8,
+    name: StringId,
     tag_ty: Type,
     fields: []Field,
     fixed: bool,
 
     pub const Field = struct {
-        name: []const u8,
+        name: StringId,
         ty: Type,
         name_tok: TokenIndex,
         node: NodeIndex,
@@ -151,7 +153,7 @@ pub const Enum = struct {
         return e.fields.len == std.math.maxInt(usize);
     }
 
-    pub fn create(allocator: std.mem.Allocator, name: []const u8, fixed_ty: ?Type) !*Enum {
+    pub fn create(allocator: std.mem.Allocator, name: StringId, fixed_ty: ?Type) !*Enum {
         var e = try allocator.create(Enum);
         e.name = name;
         e.fields.len = std.math.maxInt(usize);
@@ -163,7 +165,7 @@ pub const Enum = struct {
 
 // TODO improve memory usage
 pub const Record = struct {
-    name: []const u8,
+    name: StringId,
     fields: []Field,
     size: u64,
     alignment: u29,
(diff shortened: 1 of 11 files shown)


ehaas:
Some quick benchmarking suggests that this is actually slower than status quo. I suspect it may be due to encoding the length as a variable-length int; I'm going to switch it back to null termination and see if it makes a difference.

ehaas:
Here are some quick benchmark results of the runtime of the `Parser.parse` function, and overall memory usage, when parsing the sqlite3 source. Compiled with `-Drelease-fast=true`. Runtime was measured by calling `std.time.nanoTimestamp` before and after `var tree = try Parser.parse(&pp);` in `src/main.c`.  Memory usage was measured by using `GeneralPurposeAllocator`'s `total_requested_bytes` field. All three parses had identical textual AST representations. Test machine was an Intel i5 @ 2.40GHz on Ubuntu x86_64 GNU/Linux. Time measurements are the average of six runs with the slowest result dropped.

Master branch:
Bytes allocated: `8,969,028`
Runtime: `261 ms` (`260,662,139.8 ns`)

string-intern branch, null-terminated interned strings:
Bytes allocated: `9,211,470`
Runtime: `232 ms` (`232,072,750.8 ns`)

string-intern branch, LEB-encoded length:
Bytes allocated: `9,211,470`
Runtime: `229 ms` (`229,075,553.4 ns`)

So it looks like the additional storage needed for the interned strings array exceeds the savings from having smaller fields in `Record`/`Enum`/etc, but it does provide a 12-14% speedup. If we also wanted to reduce memory usage I think we'd have to figure out a way to use `TokenIndex` in the type fields (maybe the first TokenIndex that expands to the desired string). We'd also need a solution for anonymous records/enums if we went that route.

ehaas:
Directly interning the `getAnonymousName` result reduces the runtime by another 7ms, so it's consistently `222 ms` now with the C allocator.

Vexu:
I don't think it's a good idea to store the string bytes. For most identifiers the data is already stored in the loaded source files and the only generated ones, anonymous field and container names, are always unique. The only way that could be useful would be to store tokens for the source files instead of the raw bytes which could speed up preprocessing but would introduce other complexities.

It would however be nice to have a way to convert `[]const u8 -> u32` since `std.mem.eql` is likely one of the hottest functions in the entire compiler.

ehaas:
Agreed it should be possible to do this without duplicating the string bytes. I have some other ideas I'm going to test out.

ehaas:
Ok, tried a new approach here. Still a work in progress; sorry it's just one big commit - had a couple false starts and a bunch of abandoned approaches after doing some profiling.

1. The only reason `string_intern.zig` is its own file (and `TypeMapper` exists) was to avoid creating a compile-time dependency of `Type` on `Source` (since one didn't exist before). If that doesn't matter then I can remove that bit of complexity.
2. I still have some more cleanup / documentation / better commit message to do
3. I plan to do some memory usage benchmarks tomorrow; hopefully usage will be lower. Speed is the same as the previous version (about 12-14% faster than before for large files)

ehaas:
One thing I don't love about this current implementation is that it feels pretty un-ergonomic to have to pass the location slice around on every call to `intern`. In this model it's also a problem having the interned strings be owned by the Compilation, because if you attempt to compile 2 different files the location slices aren't going to match up (the hash table relies on the location slice being stable)

I was wondering if it would make more sense to move it into Parser. The only problem is the types that get generated in `generateVaListType`, so that would have to be moved into Parser somehow.

Vexu:
I don't understand why the locations are being used in the first place. Is it because you want to avoid storing slices?

ehaas:
Yeah that was the idea but it might not be worth the hassle. I'll adjust it to just be a regular hashmap that uses slices and take some more measurements.

ehaas:
Ok, here are some updated numbers. The memory usage numbers don't match up with what I recorded last time; I probably did something wrong back then. The memory numbers are measuring the difference between `general_purpose_allocator.total_requested_bytes` before and after running `Parser.parse` in `main.zig` on the sqlite3 source code.

Master branch:
`Parser.parse` runtime: `261.7ms`
Memory usage: `15,866,380` bytes

Simple hash table + incrementing string id (this branch):
`Parser.parse` runtime: `212.3ms` (decrease of `49.4ms` or 18.9%)
Memory usage: `16,148,930` bytes (increase of `282,550` bytes or ~1.8%)

String pseudo-key with string id indexing into an array storing LEB-encoded length + string contents:
`Parser.parse` runtime: `222.8ms` (decrease of `38.9ms` or 14.9%)
Memory usage: `16,058,645` bytes (increase of `192,265` bytes or ~1.2%)

So it looks like current master branch actually has the lowest memory usage of the options I looked at. The string pseudo-key method saves a bit of memory compared to the simple hashtable method (90,285 bytes on an 8-meg source file) but it's also more complicated and slower.

Vexu:
Increased memory usage was always expected but the speedup is nice. Maybe it'll be worth trying to only save tokens instead of full sources at some point.

src/Type.zig
```diff
@@ -97,7 +99,7 @@ pub const Func = struct {
     params: []Param,
 
     pub const Param = struct {
-        name: []const u8,
+        name: StringId,
```
Vexu, line 102: You might want to try to reorganize these fields for better memory usage now that the name fields take up less space.

src/main.zig
```diff
@@ -370,7 +370,10 @@ fn processSource(comp: *Compilation, source: Source, builtin: Source, user_macro
             fatal(comp, "unable to write result: {s}", .{util.errorDescription(er)});
     }
 
+    const start = std.time.nanoTimestamp();
```
Vexu, line 373: These should be removed once you're done.

Vexu (approved):

ehaas:
I rearranged some fields and got the following savings (based on @sizeOf())

`Func.Param` 32 bytes -> 24 bytes
`Enum` 48 bytes -> 40 bytes
`Record` 48 bytes -> 40 bytes

Strangely enough though, the `total_requested_bytes` difference (between start of parse and end of parse) actually went up a bit. So I'm not sure if that's just not a good way to measure things, or if something else is going on.

Vexu:
> Strangely enough though, the `total_requested_bytes` difference (between start of parse and end of parse) actually went up a bit. So I'm not sure if that's just not a good way to measure things, or if something else is going on.

That may very well be a bug in `GeneralPurposeAllocator`, might be better to use the C allocator and using valgrind or something.

ehaas:
Here are the numbers from valgrind for total usage on the sqlite3 source, arocc built with `-Dlink-libc=true -Dcpu=baseline -Drelease-fast=true`:

String intern branch, before field rearrangement: `total heap usage: 256,417 allocs, 256,417 frees, 120,724,797 bytes allocated`

String intern branch, after field rearrangement: `total heap usage: 256,417 allocs, 256,417 frees, 120,727,009 bytes allocated`

So somehow it's using `2212` more bytes despite the records being smaller.

Vexu:
Weird, maybe it's causing some arraylist to grow differently?

PR #323, opened 2022-06-30
Add Linux multiarch include dir

ehaas:
With this change I'm able to parse the sqlite3 amalgamated source with no errors (just a bunch of spurious "unreachable code" warnings) on an x86_64 Ubuntu system without passing any additional command-line flags.


Vexu (approved):
We should probably look into trying to reuse `std.zig.system.NativePaths` at some point.

>just a bunch of spurious "unreachable code" warnings

Did they seem incorrect or correct but not emitted by clang/GCC by default?

PR #324, opened 2022-06-30
Parser: `if_then_else_stmt` node is only noreturn if both arms are noreturn

ehaas:


Vexu (approved):
Oops, looks like I messed up the logic when fixing another bug in this.

Issue #328, opened 2022-07-01
`-Wreturn-type` warning should be suppressed if implicit return is not reachable


ehaas:
arocc currently issues a warning for this; clang and gcc do not:
```c
int foo(int x) {
  while (1) {
    return x;
  }
}
```

Issue #325, opened 2022-06-30
pointer cast warnings too strict


ehaas:
This currently gives a warning when it shouldn't:

```c
void foo(const void *arg) {
  const char * const p1 = (const char * const)arg;
}
```

```
./test.c:2:27: warning: cast to type 'const char *const ' will not preserve qualifiers [-Wcast-qualifiers]
  const char * const p1 = (const char * const)arg;
                          ^
```

Vexu:
That error is actually talking about the second `const` and happens for other types as well:
```c
void foo() {
    const char p1 = (const char)1; // warning: cast to type 'char' will not preserve qualifiers [-Wcast-qualifiers]
}
```

This is an warning that I added so it should either be clarified or removed altogether.

PR #330, opened 2022-07-03
Parser: don't issue a bounds warning for trailing one-element array access

### LINKED ISSUE #327: Alternative syntax for flexible arrays
GCC and clang treat trailing one-element arrays at the end of structs as flexible arrays, because that's how people used to do it in C90. From my research it seems that it's one of those things that's technically undefined behavior according to the standard, but all (or most) vendors implement it the same way.

From https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html (emphasis mine):

> In the absence of the zero-length array extension, in ISO C90 the contents array in the example above would typically be declared to have a single element. Unlike a zero-length array which only contributes to the size of the enclosing structure for the purposes of alignment, a one-element array always occupies at least as much space as a single object of the type. **Although using one-element arrays this way is discouraged, GCC handles accesses to trailing one-element array members analogously to zero-length arrays.**

This construct is used in sqlite, so arocc currently issues some warnings that gcc & clang do not.

ehaas:
Closes #327

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -141,6 +141,7 @@ pub const Options = packed struct {
     @"designated-init": Kind = .default,
     @"attribute-warning": Kind = .default,
     @"invalid-noreturn": Kind = .default,
+    @"zero-length-array": Kind = .default,
 };
 
 const messages = struct {
@@ -1981,6 +1982,12 @@ const messages = struct {
         const msg = "invalid type '{s}' to __real operator";
         const extra = .str;
         const kind = .@"error";
+    };
+    const zero_length_array = struct {
+        const msg = "zero size arrays are an extension";
+        const kind = .off;
+        const pedantic = true;
+        const opt = "zero-length-array";
     };
 };
 
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -142,6 +142,7 @@ pub const Options = packed struct {
     @"attribute-warning": Kind = .default,
     @"invalid-noreturn": Kind = .default,
     @"zero-length-array": Kind = .default,
+    @"old-style-flexible-struct": Kind = .default,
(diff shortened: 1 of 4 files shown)


ehaas:
After this change we're down to 19 warnings in sqlite.

2 from overflow due to incorrect sizeof calculations
2 from #328
4 from #325
11 from unreachable code warnings

Vexu:
This does seem to work with everything I try but it feels a bit odd having it be part of the global state of the `Parser`. Is there some specific reason why you chose this approach instead of something more confined in `checkArrayBounds`?

ehaas:
I'd prefer to do it in `checkArrayBounds` but I couldn't think of a way to tell if I was dealing with a trailing one-element array field. GCC doesn't seem to do any bounds checking at all; clang will issue a warning if it's a standalone array or non-trailing one-element field, but no warning if it's a trailing one element field.

Vexu:
Seems to me that if you pass `checkArrayBounds` the array `Result` instead of just the type, then you could inline `Parser.getNode` and move the logic you currently have in `fieldAccessExtra` there.

Vexu:
Accidentally pushed unrelated changes to this branch, reverted now.

ehaas:
> Is there/will there be an option to enable a warning for this?

I don't think Clang has an option to enable a warning for this (didn't see any here https://clang.llvm.org/docs/DiagnosticsReference.html) so we could create our own. Any suggestions for the name?

ehaas:
Updated to keep the check local to `checkArrayBounds`, removed bounds checking for zero-element arrays, added a warning for zero-element arrays (at creation time) and added the `-Wold-style-flexible-struct` warning. I set it to default off since clang doesn't issue that warning by default but I can switch it to warn by default if you want.

PR #335, opened 2022-07-07
CLI: disable color output by default if NO_COLOR set or escape codes not supported

ehaas:


ehaas:
My intention here was to prevent `--verbose-ast` from using color when redirected to a file / `NO_COLOR` set, but I see now that there is also some logic in `Diagnostics.init` - since one is using stdout and the other stderr, it seems like we'd actually need two flags (since `Tree.dump` is just using `diag.color`)

Vexu:
> On second thought, maybe the right approach here is just to temporarily override `comp.diag.color` in the `comp.verbose_ast` check, when dumping the Tree (if stdout does not support color)?

Sounds good to me.

ehaas:
Hopefully not too big of a change - I wanted to make it so the `NO_COLOR` env var can be overridden by command-line flags. I also wanted the default value of `Diagnostics.color` in the library not to depend on the existence of file descriptors or environment variables

src/util.zig
```diff
@@ -12,6 +12,20 @@ pub const Color = enum {
     white,
 };
 
+pub fn environmentDisabledColor() bool {
```
Vexu, line 15: This does the same as `std.process.hasEnvVarConstant`.

src/Compilation.zig
```diff
@@ -30,6 +30,11 @@ pragma_handlers: std.StringArrayHashMap(*Pragma),
 only_preprocess: bool = false,
 only_compile: bool = false,
 verbose_ast: bool = false,
+color_diagnostics: enum {
```
Vexu, line 33: This should be a local variable in `main.zig`.

src/main.zig
```diff
@@ -358,8 +359,10 @@ fn processSource(comp: *Compilation, source: Source, builtin: Source, user_macro
     defer tree.deinit();
 
     if (comp.verbose_ast) {
-        var buf_writer = std.io.bufferedWriter(std.io.getStdOut().writer());
-        tree.dump(buf_writer.writer()) catch {};
+        const stdout = std.io.getStdOut();
+        var buf_writer = std.io.bufferedWriter(stdout.writer());
+        const color = comp.shouldUseColor(util.fileSupportsColor(stdout), util.environmentDisabledColor());
```
Vexu, line 364:
```suggestion
        const color = comp.diag.color and util.fileSupportsColor(stdout);
```

Vexu (changes requested):

ehaas:
For the variable shadowing error - would you rather rename the `err` function in various modules, or rename the error capture?

Vexu:
> For the variable shadowing error - would you rather rename the `err` function in various modules, or rename the error capture?

Oh right, its the bugs I filed getting fixed. I decided to fix it by renaming the captures since I've done that before with `if`s on error unions and there are fewer captures than calls to the `err` functions.

PR #340, opened 2022-07-15
Preprocessor: stricter argument validation for zero-argument macros

### LINKED ISSUE #338: Preprocessor: incorrect handling of zero-argument function-like macro called with arguments
```c
#define NO_ARGUMENT()
NO_ARGUMENT(1)
```
clang output (gcc's error message is different but it also does not allow this construct):
```
test.c:2:13: error: too many arguments provided to function-like macro
      invocation
NO_ARGUMENT(1)
            ^
test.c:1:9: note: macro 'NO_ARGUMENT' defined here
#define NO_ARGUMENT()
        ^
```

aro currently does not issue an error for this.

ehaas:
Check that the argument to a zero-argument macro is empty or only
consists of .macro_ws tokens

Fixes #338


test/cases/expanded/zero argument macro.c
```diff
@@ -0,0 +1,10 @@
+
+
+
+
+
+
+
+
+NO_ARGUMENTS(1)
```
Vexu, line 9: Both GCC and clang remove the argument list, implementing that behavior should be as easy as copying the logic from the `error.Unterminated` prong from the switch above.

PR #344, opened 2022-07-26
set default compiler, and other target queires

TwoClocks:
Some target-specific question FNs needed for layout.

The most controversial is setting the compiler emulation based on the target. It sets the emulation to the compiler the target system is typically built with. So for example, Linux is built w/ gcc, so that's the compiler to emulate for linux targets. the `--emulate` flag overrides.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -899,6 +899,65 @@ pub fn isTlsSupported(comp: *Compilation) bool {
     };
 }
 
+pub fn ignoreNonZeroSizedBitfieldTypeAlignment(comp: *const Compilation) bool {
+    switch (comp.target.cpu.arch) {
+        .avr => return true,
+        .arm => {
+            if (std.Target.arm.featureSetHas(comp.target.cpu.features, .has_v7)) {
+                switch (comp.target.os.tag) {
+                    .ios => return true,
+                    else => return false,
+                }
+            }
+        },
+        else => return false,
+    }
+    return false;
+}
+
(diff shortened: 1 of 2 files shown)


Vexu (changes requested):
This doesn't compile if you reference the new functions.

TwoClocks:
I was never calling that `fn` from the layout code either, which is a bug as well.

TwoClocks:
added tests for new functions

src/Compilation.zig
```diff
@@ -1078,3 +1078,37 @@ test "addSourceFromReader - exhaustive check for carriage return elimination" {
...
+
+    const arm = std.Target.Cpu.Arch.arm;
+    comp.target.cpu = std.Target.Cpu.baseline(arm);
+    comp.target.os = std.Target.Os.Tag.defaultVersionRange(.ios, arm);
+    comp.target.abi = std.Target.Abi.default(arm, comp.target.os);
+
+    try std.testing.expect(!comp.isTlsSupported());
+    try std.testing.expect(comp.ignoreNonZeroSizedBitfieldTypeAlignment());
+    // stage 0 can't compare a ?u29 with a ?u29
+    if (comp.minZeroWidthBitfieldAlignment()) |zero_algin| {
```
ehaas, line 1107: You can use `std.testing.expectEqual(expected, actual)` for this - it works with optionals of the same type.

TwoClocks: ah yeah. much better.
thanks.

PR #345, opened 2022-07-26
prep record struct for layout

TwoClocks:
Add `TypeLayout` to `Record` and `FieldLayout` to `Field` in `Type.zig`

`TypeLayout` is probably too fat. I think I can cut it in half, but I'd like to finish msvc and MinGW layout emulation before doing that. It helps w/ diff-ing the rust code.

Removed `size` and `alignment` from `Record` as they are duplicative of `TypeLayout`

Changed `bits` in `Field` to be nullable, as 0 and unset mean different things to the layout code.

### DIFF
--- a/src/Type.zig
+++ b/src/Type.zig
@@ -170,23 +170,23 @@ pub const TypeLayout = struct {
     ///
     /// This is the value returned by `sizeof` and C and `std::mem::size_of` in Rust
     /// (but in bits instead of bytes). This is a multiple of `pointer_alignment_bits`.
-    size_bits: u29,
+    size_bits: u64,
     /// The alignment of the type, in bits, when used as a field in a record.
     ///
     /// This is usually the value returned by `_Alignof` in C, but there are some edge
     /// cases in GCC where `_Alignof` returns a smaller value.
-    field_alignment_bits: u29,
+    field_alignment_bits: u64,
     /// The alignment, in bits, of valid pointers to this type.
     ///
     /// This is the value returned by `std::mem::align_of` in Rust
     /// (but in bits instead of bytes). `size_bits` is a multiple of this value.
-    pointer_alignment_bits: u29,
+    pointer_alignment_bits: u64,
     /// The required alignment of the type in bits.
     ///
     /// This value is only used by MSVC targets. It is 8 on all other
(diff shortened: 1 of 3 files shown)


src/Type.zig
```diff
@@ -162,24 +162,70 @@ pub const Enum = struct {
...
 };
+// might not need all 4 of these when finished,
+// but currently it helps havaing all 4 when diff-ing
+// the reust code.
+pub const TypeLayout = struct {
+    /// The size of the type in bits.
+    ///
+    /// This is the value returned by `sizeof` and C and `std::mem::size_of` in Rust
+    /// (but in bits instead of bytes). This is a multiple of `pointer_alignment_bits`.
+    size_bits: u29,
```
Vexu, line 173: It's weird that this uses `u29` when we allow array sizes up to `u64`. How does the struct layout code handle huge arrays in structs? How do GCC/clang handle it?

ehaas:
All of the size fields in the Rust code (in FieldLayout and TypeLayout) are u64 and not u29

src/Type.zig
TwoClocks: That's not wierd, it's wrong. Fixed.
Thanks!

TwoClocks:
Yep, they should all be u64.
Thanks

src/Type.zig
```diff
@@ -162,24 +162,70 @@ pub const Enum = struct {
...
+    ///
+    /// This is the value returned by `std::mem::align_of` in Rust
+    /// (but in bits instead of bytes). `size_bits` is a multiple of this value.
+    pointer_alignment_bits: u64,
+    /// The required alignment of the type in bits.
+    ///
+    /// This value is only used by MSVC targets. It is 8 on all other
+    /// targets. On MSVC targets, this value restricts the effects of `#pragma pack` except
+    /// in some cases involving bit-fields.
+    required_alignmnet_bits: u64,
```
ehaas, line 189: typo/spelling error - should be `required_alignment_bits`

TwoClocks: I am so dyslexic, I had to stare at your comment for 2min before I saw the difference :smile: 
I need to get neovim to spell check snake & camelCase.

PR #347, opened 2022-07-28
Zig stdlib: update format string for attribute printing

ehaas:


TwoClocks:
I don't think the recent changes to `std.fmt` are backward compatible. I've been reverting `Tree.zig` before committing to pass the tests.

TwoClocks:
this PR is a recent change `std.fmt` in zig:
https://github.com/ziglang/zig/commit/a0d3a87ce15a9f68047dc900109f5b76184d046f

This PR updates arocc for this change to zig.
But the github CI is using an older version of zig (AFAICT), so this PR fails (failed?) CI.

So I've been backing out a local change to `Tree.zig` so my PR's pass CI, while my local zig version is post- `std.fmt` change.
I don't think the change to `std.fmt` is backwards compatible. That's all I meant.

Vexu:
The CI failed because the fix was broken.

PR #348, opened 2022-07-28
Preprocessor: check for include guards

### LINKED ISSUE #341: Preprocessor optimization: Detect include guards
GCC and clang detect the include guard pattern:
```C
#ifndef FOO_H
#define FOO_H
// ...
#endif  // FOO_H
```

and use that to avoid re-lexing files that use the include guard. 

https://gcc.gnu.org/onlinedocs/cpp/Once-Only-Headers.html#Once-Only-Headers (CPP on that page means C Preprocessor, not C++)

https://clang.llvm.org/docs/InternalsManual.html#multipleincludeopt

ehaas:
We can skip preprocessing a file if it is entirely contained in
an #ifndef ... #endif guard and the guard macro is defined.

Closes #341

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -79,6 +79,9 @@ include_depth: u8 = 0,
 counter: u32 = 0,
 expansion_source_loc: Source.Location = undefined,
 poisoned_identifiers: std.StringHashMap(void),
+/// Map from Source.Id to macro name in the `#ifndef` condition which guards the source, if any
+include_guards: std.AutoHashMapUnmanaged(Source.Id, []const u8) = .{},
+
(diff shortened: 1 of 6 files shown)


Vexu:
It seems like it would be easier to just check for EOF after every undef, is there some case I'm no thinking of?

ehaas:
I think we still need to check for `#elif` / `#else` (and eventually `#elifdef` / `#elifndef`) which are paired with top-level `#ifndef` but I like your idea better for the other ones.

Vexu:
> I think we still need to check for `#elif` / `#else` (and eventually `#elifdef` / `#elifndef`) which are paired with top-level `#ifndef`

That would be something I didn't think of but it seems like it was still a nicer solution.

PR #355, opened 2022-07-29
test suite for record layout

TwoClocks:
Here are the tests from the Rust repr-c converted to C with a zig runner.

Since neither layout nor `_bifoffsetof` are merged, only the parser is enabled for testing. All the static asserts are skipped.

See `test/records.md` for details about the tests, and how to control what targets and tests are run.


Vexu:
I think this should be merged last.

PR #356, opened 2022-07-30
`__builtin_bitoffsetof` support

TwoClocks:
add support for `bitoffsetof`

I took a best guess. If you imaged something else let me know, or fix it up the way you like it.

### DIFF
--- a/include/stddef.h
+++ b/include/stddef.h
@@ -14,3 +14,4 @@ typedef struct {
 
 #define NULL ((void*)0)
 #define offsetof(T, member) __builtin_offsetof(T, member)
+#define bitoffsetof(T, member) __builtin_bitoffsetof(T, member)
--- a/include/stddef.h
+++ b/include/stddef.h
@@ -14,4 +14,3 @@ typedef struct {
 
 #define NULL ((void*)0)
 #define offsetof(T, member) __builtin_offsetof(T, member)
-#define bitoffsetof(T, member) __builtin_bitoffsetof(T, member)
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -6069,7 +6069,7 @@ fn fieldAccessExtra(p: *Parser, lhs: NodeIndex, record_ty: Type, field_name: Str
             return p.fieldAccessExtra(inner, f.ty, field_name, false, offset_bits);
         }
         if (field_name == f.name) {
-            offset_bits.* = f.layout.offset_bits;
+            offset_bits.* = @intCast(usize, f.layout.offset_bits);
             return Result{
                 .ty = f.ty,
                 .node = try p.addNode(.{
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -5569,7 +5569,7 @@ fn offsetofMemberDesignator(p: *Parser, base_ty: Type) Error!Result {
     try p.validateFieldAccess(base_ty, base_ty, base_field_name_tok, base_field_name);
     const base_node = try p.addNode(.{ .tag = .default_init_expr, .ty = base_ty, .data = undefined });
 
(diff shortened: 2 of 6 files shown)


src/Parser.zig
```diff
@@ -6061,16 +6066,19 @@ fn fieldAccessExtra(p: *Parser, lhs: NodeIndex, record_ty: Type, field_name: Str
                 .ty = f.ty,
                 .data = .{ .member = .{ .lhs = lhs, .index = @intCast(u32, i) } },
             });
-            return p.fieldAccessExtra(inner, f.ty, field_name, false);
+            return p.fieldAccessExtra(inner, f.ty, field_name, false, offset_bits);
```
Vexu, line 6069: Shouldn't this also take into account the offset of the anonymous struct field? As in:
```c
struct Foo {
    int a;
    struct {
        int b;
    };
};
_Static_assert(__builtin_offsetof(struct Foo, b) == 4, "");

```

src/Parser.zig
```diff
@@ -5583,7 +5586,8 @@ fn offsetofMemberDesignator(p: *Parser, base_ty: Type) Error!Result {
             }
             try p.validateFieldAccess(lhs.ty, lhs.ty, field_name_tok, field_name);
             const record_ty = lhs.ty.canonicalize(.standard);
-            lhs = try p.fieldAccessExtra(lhs.node, record_ty, field_name, false);
+            lhs = try p.fieldAccessExtra(lhs.node, record_ty, field_name, false, &offset_num);
+            bit_offset = Value.int(offset_num);
```
Vexu, line 5590: This should check if `bit_offset != .unavailable` and then add to it rather than setting it.

src/Parser.zig
```diff
@@ -5516,7 +5518,7 @@ fn builtinVaArg(p: *Parser) Error!Result {
     }) };
 }
 
-fn builtinOffsetof(p: *Parser) Error!Result {
+fn builtinOffsetof(p: *Parser, wantBits: bool) Error!Result {
```
Vexu, line 5521:
```suggestion
fn builtinOffsetof(p: *Parser, want_bits: bool) Error!Result {
```
Style nit.

src/Parser.zig
TwoClocks: changed.
I tested this case in a branch w/ layout code. It works now.

src/Parser.zig
TwoClocks: added check, and addation.

src/Parser.zig
TwoClocks: fixed

Vexu:
> [fix usize cast for 32-bit platforms](https://github.com/Vexu/arocc/pull/356/commits/2d401bb65950993b6be2b3a8a19586888b8e7248)

Shouldn't this be the opposite, i.e. changing the offset form `usize` to `u64` to match `FieldLayout.offset_bits`? Not that it'll actually matter since I doubt many people are making 4GiB structs.

TwoClocks:
If the size is too big, the failure is a crash. So might as well fix it. Also, 4G-bits is in the sorta/almost/maybe possible.

include/stddef.h
```diff
@@ -14,3 +14,4 @@ typedef struct {
 
 #define NULL ((void*)0)
 #define offsetof(T, member) __builtin_offsetof(T, member)
+#define bitoffsetof(T, member) __builtin_bitoffsetof(T, member)
```
ehaas, line 17: We probably shouldn't define this, because it's not a reserved identifier and not part of the C standard. User code might use `bitoffsetof` and it shouldn't break due to including `stddef.h`.

TwoClocks: nice catch.

PR #357, opened 2022-07-30
change logic in `Type.alignof()`

TwoClocks:
This is the last change needed for `record_layout.zig`.

I could put this in the PR w/ the layout code, but I wanted to draw your attention to it, as I'm not sure this is the proper way to handle these issues. There might be more of these types of exceptions in the future.

### DIFF
--- a/src/Type.zig
+++ b/src/Type.zig
@@ -821,7 +821,7 @@ pub fn sizeof(ty: Type, comp: *const Compilation) ?u64 {
     };
 }
 
-pub fn bitSizeof(ty: Type, comp: *Compilation) ?u64 {
+pub fn bitSizeof(ty: Type, comp: *const Compilation) ?u64 {
     return switch (ty.specifier) {
         .bool => 1,
         .typeof_type, .decayed_typeof_type => ty.data.sub_type.bitSizeof(comp),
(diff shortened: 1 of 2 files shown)


src/Type.zig
```diff
@@ -833,7 +833,20 @@ pub fn bitSizeof(ty: Type, comp: *Compilation) ?u64 {
...
 /// Get the alignment of a type
 pub fn alignof(ty: Type, comp: *const Compilation) u29 {
-    if (ty.requestedAlignment(comp)) |requested| return requested;
+
+    // don't return the attribute for records
+    // layout has already accounted for requested alignment
+    if (!ty.isRecord()) {
+        if (ty.requestedAlignment(comp)) |requested| {
+            // gcc does not respect alignment on enums
+            if (ty.get(.@"enum")) |ty_enum| {
```
Vexu, line 842: Enums are not considered records so this logic is unreachable here and should go somewhere else or you should use `isEnumOrRecord`.

TwoClocks: This is correct. There are two issues being addressed here.
1 - if it is a record, do not return the requested alignment. Return whatever layout decided is the alignment.
2 - If it is an enum, and emulating gcc do not return the requested alignment.

Vexu: My bad, I missed the `!`.

TwoClocks: no worries. I could break this out into different checks for different issues. It would make it clear what is happening. It would do a few more comparisons than needed.

PR #358, opened 2022-07-30
Compilation: add MSP430 target-specific types and macros

### LINKED ISSUE #346: compiler crashes setting non x86 tarrget.
running
`arocc --target=msp430-other-none 0001_test.c`

yields
```
thread 2516858 panic: reached unreachable code
/home/jonross/tinker/forks/arocc/src/Compilation.zig:382:17: 0x309053 in Compilation.generateBuiltinTypes
(arocc)
        else => unreachable,
                ^
/home/jonross/tinker/forks/arocc/src/Compilation.zig:125:34: 0x2e6b52 in Compilation.generateBuiltinMacros
 (arocc)
    try comp.generateBuiltinTypes();
                                 ^
/home/jonross/tinker/forks/arocc/src/main.zig:327:51: 0x2dff2f in mainExtra (arocc)
    const builtin = try comp.generateBuiltinMacros();
                                                  ^
/home/jonross/tinker/forks/arocc/src/main.zig:44:14: 0x2dc430 in main (arocc)
    mainExtra(&comp, args) catch |er| switch (er) {
             ^
/home/jonross/tinker/clone/zig/lib/std/start.zig:577:29: 0x2bb883 in std.start.posixCallMainAndExit (arocc
)
            return root.main();
                            ^

```

ehaas:
Macros taken from here: https://github.com/llvm/llvm-project/blob/main/clang/lib/Basic/Targets/MSP430.cpp

Size, alignment, va_list, char signedness info: https://www.ti.com/lit/an/slaa534a/slaa534a.pdf (page 12)

The MSP430 compiler also has the concept of code models. Different code models can adjust the size
of function pointers, data pointers, size_t, and ptrdiff_t. We currently only support the small model.

Closes #346

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -278,6 +278,11 @@ pub fn generateBuiltinMacros(comp: *Compilation) !Source {
             \\
         ),
         .aarch64, .aarch64_be => try w.writeAll("#define __aarch64__ 1\n"),
+        .msp430 => try w.writeAll(
+            \\#define MSP430 1
+            \\#define __MSP430__ 1
+            \\
+        ),
         else => {},
     }
 
@@ -362,7 +367,7 @@ fn generateBuiltinTypes(comp: *Compilation) !void {
     const os = comp.target.os.tag;
     const wchar: Type = switch (comp.target.cpu.arch) {
         .xcore => .{ .specifier = .uchar },
-        .ve => .{ .specifier = .uint },
+        .ve, .msp430 => .{ .specifier = .uint },
         .arm, .armeb, .thumb, .thumbeb => .{
             .specifier = if (os != .windows and os != .netbsd and os != .openbsd) .uint else .int,
         },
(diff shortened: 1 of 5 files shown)


src/Compilation.zig
```diff
@@ -278,6 +278,12 @@ pub fn generateBuiltinMacros(comp: *Compilation) !Source {
             \\
         ),
         .aarch64, .aarch64_be => try w.writeAll("#define __aarch64__ 1\n"),
+        .msp430 => try w.writeAll(
+            \\#define MSP430 1
+            \\#define __MSP430__ 1
+            \\#define __ELF__ 1
```
Vexu, line 284: This should be defined for many more targets.

src/Compilation.zig
```diff
@@ -375,6 +381,8 @@ fn generateBuiltinTypes(comp: *Compilation) !void {
 
     const ptrdiff = if (os == .windows and comp.target.cpu.arch.ptrBitWidth() == 64)
         Type{ .specifier = .long_long }
+    else if (comp.target.cpu.arch == .msp430)
+        Type{ .specifier = .int }
```
Vexu, line 385: Wouldn't a more general solution be adding a prong for 16 bit architectures? I'd assume most of those also define `int` to be 16 bits.

Vexu (approved):
These type sizes should also be contributed to Zig's `CType` and then reused here.

ehaas:
> These type sizes should also be contributed to Zig's `CType` and then reused here.

It looks like `CType` already has defines for MSP430 (as well as other architectures) - should I just change `Type.zig` to use that for the basic scalar types?

ehaas:
I took out `__ELF__` and will add that as a separate PR.

Vexu:
> I took out `__ELF__` and will add that as a separate PR.

`getObjectFormat` should make this easy.

Issue #178, opened 2021-12-08
Calculate record size and alignment


Vexu:
Currently both just default to `1`.

Note: the size calculation should happen after the attributes after the type are parsed since they may contain a packed attribute.

ehaas:
There's a rust crate that does this, which is MIT licensed, which could possibly be used as inspiration: https://github.com/mahkoh/repr-c/tree/master/repc/facade

Should this be its own project? Seems big enough to be, plus it could also be used by the Zig compiler when it needs to know C ABI layout independently of translate-c

Vexu:
> There's a rust crate that does this, which is MIT licensed, which could possibly be used as inspiration: https://github.com/mahkoh/repr-c/tree/master/repc/facade

That does seem like the best place to start.

> Should this be its own project? Seems big enough to be, plus it could also be used by the Zig compiler when it needs to know C ABI layout independently of translate-c

I guess it could be, though if the Zig compiler needs it it will have access to it via the Aro dependency anyways.

TwoClocks:
I started poking around w/ this (unsolicited. sorry if this is redundant). 

The rust repro keeps in tests in one DSL and the results in a different DSL. To generate the reference compiler version, it creates C from the input DSL then runs it through the compilers and converts the compilers debug output (DWARF) to a DSL. The resulting compiler DSL's are checked into the repro.

I modified the generation code to just output the C files. I put them in this repro https://github.com/TwoClocks/repr-c-test-files.

AROCC compiles about 1/2 of them. It currently doesn't parse `__attribute__`, which are the majority of the errors.

The bulk of the rest look compiler-specific. 

I'm not sure what the plan for AROCC is regarding the intended C compiler (this is different from ABI?).

Looking at the repr-c code, it seems like you can do a fair bit of layout correctly w/o knowing the intended compiler. But for anything non-vanilla, the intended compiler must be known.

At a casual glance, more info is needed to do layout correctly than arocc currently has in it's `Type`. Or I'm misreading the code (this is likely).

Vexu:
> I'm not sure what the plan for AROCC is regarding the intended C compiler (this is different from ABI?).

One of the main goals of Aro is to replace clang as the C frontend of Zig's `translate-c` feature and eventually even being an alternative frontend for `zig cc` so by default it should be clang compatible (which should be gcc compatible).

> At a casual glance, more info is needed to do layout correctly than arocc currently has in it's `Type`. Or I'm misreading the code (this is likely).

[...]

PR #364, opened 2022-08-02
record layout

TwoClocks:
here it is. Have at it.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -19,6 +19,7 @@ const CharInfo = @import("CharInfo.zig");
 const Value = @import("Value.zig");
 const SymbolStack = @import("SymbolStack.zig");
 const Symbol = SymbolStack.Symbol;
+const RecordLayout = @import("record_layout.zig");
 const StringId = @import("StringInterner.zig").StringId;
 
 const Parser = @This();
@@ -1755,8 +1756,6 @@ fn recordSpec(p: *Parser) Error!Type {
         if (field.ty.hasIncompleteSize()) break;
     } else {
         record_ty.fields = try p.arena.dupe(Type.Record.Field, p.record_buf.items[record_buf_top..]);
-        // TODO actually calculate
-        record_ty.type_layout = Type.TypeLayout.init(1, 1);
     }
     if (old_field_attr_start < p.field_attr_buf.items.len) {
         const field_attr_slice = p.field_attr_buf.items[old_field_attr_start..];
@@ -1778,6 +1777,10 @@ fn recordSpec(p: *Parser) Error!Type {
     }, attr_buf_top, null);
     if (ty.specifier == .attributed and symbol_index != null) {
         p.syms.syms.items(.ty)[symbol_index.?] = ty;
+    }
+
+    if (!ty.hasIncompleteSize()) {
+        RecordLayout.recordLayout(&ty, p.pp.comp, p);
     }
 
     // finish by creating a node
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1780,7 +1780,7 @@ fn recordSpec(p: *Parser) Error!Type {
     }
 
     if (!ty.hasIncompleteSize()) {
-        RecordLayout.recordLayout(&ty, p.pp.comp, p);
+        RecordLayout.compute(&ty, p.pp.comp, p.pragma_pack);
     }
 
     // finish by creating a node
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -19,7 +19,7 @@ const CharInfo = @import("CharInfo.zig");
 const Value = @import("Value.zig");
 const SymbolStack = @import("SymbolStack.zig");
 const Symbol = SymbolStack.Symbol;
-const RecordLayout = @import("record_layout.zig");
+const record_layout = @import("record_layout.zig");
 const StringId = @import("StringInterner.zig").StringId;
 
 const Parser = @This();
@@ -1780,7 +1780,7 @@ fn recordSpec(p: *Parser) Error!Type {
     }
 
     if (!ty.hasIncompleteSize()) {
-        RecordLayout.compute(&ty, p.pp.comp, p.pragma_pack);
+        record_layout.compute(&ty, p.pp.comp, p.pragma_pack);
     }
 
     // finish by creating a node
--- a/src/record_layout.zig
+++ b/src/record_layout.zig
@@ -18,15 +18,15 @@ const TypeLayout = Type.TypeLayout;
 const BITS_PER_BYTE = 8;
 
 const SysVContext = struct {
-    // Does the record has an __attribute__((packed)) annotation.
+    /// Does the record has an __attribute__((packed)) annotation.
     attr_packed: bool,
-    // The value of #pragma pack(N) at the type level if any.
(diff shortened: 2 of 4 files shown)


src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+        };
+        self.size_bits = std.math.max(self.size_bits, offset_bits + size_bits);
+        fld.layout = .{
+            .offset_bits = offset_bits,
+            .size_bits = size_bits,
+        };
+    }
+};
+
+pub fn recordLayout(ty: *Type, comp: *const Compilation, parser: ?*const Parser) void {
```
ehaas, line 342: Is `parser.pragma_pack` the only value we ever need from the parser? Would it be simpler to just pass in a `?u8` pragma_pack value instead of the whole parser?

src/Parser.zig
```diff
@@ -1780,6 +1779,10 @@ fn recordSpec(p: *Parser) Error!Type {
         p.syms.syms.items(.ty)[symbol_index.?] = ty;
     }
 
+    if (!ty.hasIncompleteSize()) {
+        RecordLayout.recordLayout(&ty, p.pp.comp, p);
```
Vexu, line 1783:
```suggestion
        record_layout.compute(&ty, p.pp.comp, p);
```
Style nit + avoiding repetition.

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+
+// almost all the code for record layout
+// was liberally copied from this repro
+// https://github.com/mahkoh/repr-c
+// specifically the code in "repc/impl/src/" directory.
+// At the time of this writing (Q3 2022)
+// that code marked as "SPDX-License-Identifier: MIT OR Apache-2.0"
+// so is compatible with arocc's MIT licence
+
+const BITS_PER_BYTE: u29 = 8;
```
Vexu, line 18:
```suggestion
const BITS_PER_BYTE = 8;
```
No types on constants. Also lower snake case would be more conventional but I don't care about that too much.

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+    attr_packed: bool,
+    // The value of #pragma pack(N) at the type level if any.
+    max_field_align_bits: ?u64,
+    // The alignment of this record.
+    aligned_bits: u32,
+    is_union: bool,
+    // The size of the record. This might not be a multiple of 8 if the record contains bit-fields.
+    // For structs, this is also the offset of the first bit after the last field.
+    size_bits: u64,
+    fn init(ty: *Type, comp: *const Compilation, parser: ?*const Parser) SysVContext {
```
Vexu, line 31:
```suggestion
    size_bits: u64,

    fn init(ty: *Type, comp: *const Compilation, parser: ?*const Parser) SysVContext {
```
There should be empty lines between functions and other functions/fields.

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+        var ty_fld_algn_bits: u32 = fld_layout.field_alignment_bits;
+
+        // Some targets ignore the alignment of the underlying type when laying out
+        // non-zero-sized bit-fields. See test case 0072. On such targets, bit-fields never
+        // cross a storage boundary. See test case 0081.
+        if (comp.ignoreNonZeroSizedBitfieldTypeAlignment()) {
+            ty_fld_algn_bits = 1;
+        }
+
+        if (bit_width <= 0) {
```
Vexu, line 127:
```suggestion
        if (bit_width == 0) {
```
`bit_width` is unsigned so using less than or equal is misleading.

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+
+            // On GCC, if there are no packing annotations and
+            // - the field would otherwise start at an offset such that it would cross a
+            //   storage boundary or
+            // - the alignment of the type is larger than its size,
+            // then it is aligned to the type's field alignment. See test case 0083.
+            if (!has_packing_annotation) {
+                var start_bit = std.mem.alignForwardGeneric(u64, first_unused_bit, field_align_bits);
+
+                var does_field_cross_boundary = start_bit % ty_fld_algn_bits + bit_width > ty_size_bits;
```
Vexu, line 163:
```suggestion
                const does_field_cross_boundary = start_bit % ty_fld_algn_bits + bit_width > ty_size_bits;
```
`var` instead of `const`, this appears in multiple places so please review usage of `var` everywhere.

src/record_layout.zig
TwoClocks: yeah. I used to pull other things out of the parser, but as you and Vexu implemented attributes, I guess just the pragma is left.

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+                .pointer_alignment_bits = context.pointer_align_bits,
+                .required_alignment_bits = context.req_align_bits,
+            };
+        },
+    }
+}
+
+pub fn computeLayout(ty: Type, comp: *const Compilation, type_layout: *TypeLayout) void {
+    if (ty.isRecord()) {
+        const rec = ty.getRecord().?;
```
Vexu, line 386:
```suggestion
    if (ty.getRecord()) |rec| {
```

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+        type_layout.* = rec.type_layout;
+    } else {
+        type_layout.size_bits = ty.bitSizeof(comp) orelse 0;
+        type_layout.pointer_alignment_bits = ty.alignof(comp) * BITS_PER_BYTE;
+        type_layout.field_alignment_bits = type_layout.pointer_alignment_bits;
+        type_layout.required_alignment_bits = BITS_PER_BYTE;
+    }
+}
+
+pub fn getMutableRecord(ty: *Type) ?*Type.Record {
```
Vexu, line 396:
```suggestion
pub fn getMutableRecord(ty: *Type) *Type.Record {
```
This whole function is a bit awkward but it should definitely not return an optional.

Vexu (changes requested):
Found some minor issues (mostly style), but otherwise this looks quite good. I'll take a quick look after these are fixed and then it should be good to go.

src/record_layout.zig
TwoClocks: It's a clone of `getRecord()` that only differs by `const` in the return type. I didn't want to change `getRecord` because it's used in other places.

src/record_layout.zig
TwoClocks: old habit c of always using ranges when possible.

src/record_layout.zig
```diff
@@ -0,0 +1,459 @@
...
+            }
+
+            if (fld.bit_width != null) {
+                self.layoutBitField(comp, fld, field_attrs, type_layout);
+            } else {
+                self.layoutRegluarField(fld, field_attrs, type_layout);
+            }
+        }
+    }
+    fn layoutRegluarField(
```
ehaas, line 68: Typo on this function name, should be `layoutRegularField`

ehaas:
Nice work! This is very exciting. The new code correctly handles the very edge case-y struct here https://github.com/Vexu/arocc/issues/178#issuecomment-1022838829 in both gcc and clang modes (size in clang is 1, in gcc is 2). Should we close that issue?

PR #369, opened 2022-08-04
test suite for record layout

TwoClocks:
`record_readme.md` explains how the files and runner work. Please read first.

I think the `XXXX_test.c` file are close to correct. I'm not sure about some of the algorithmic tests on some platforms. I need to dig into it more.

I hacked around on `record_runner.zig` tons while I was working on layout. It went from an inclusion set of targets to an exclusion set of targets. I won't be insulted if you think it's too janky, but some direction on how to address would be helpful.

Lots of tests work for lots of targets. I'm pleasantly surprised by how many targets work.
@ehaas your changes to for `msp430` make that target work on every test except one! nice!


Vexu:
Sorry for being so slow to review this, if this is pr still works I can try to get it merged by the end of the week.

TwoClocks:
I'm doing some work to clean it up. I'll put up a new PR "soon".

TwoClocks:
feedback on the readme would be helpful while I'm re-working:
https://github.com/TwoClocks/arocc/blob/record_tests/test/record_readme.md

Vexu:
It seems fine aside from some typos. Also might be better to use 'variable' or something instead of 'static structure'.

PR #373, opened 2022-08-06
Preprocessor: treat `defined` as a built-in function-like macro

### LINKED ISSUE #337: Macro expansion which produces defined(X)
This is undefined behavior according to the C standard but both gcc and clang treat it like `defined` is a function-like macro:

```c
#define FOO_IS_DEFINED defined(FOO)
#define FOO

#if FOO_IS_DEFINED
#error "FOO is defined"
#endif
```

clang warns `test.c:4:5: warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]`

Both will evaluate the `#if` condition to true and will display the "FOO is defined" error message.

aro currently says `error: function-like macro 'defined' is not defined`

ehaas:
This allows macro expansions which produce `defined` (undefined behavior)
to behave the same as they do in GCC/clang

Closes #337

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -145,6 +145,7 @@ pub const Options = packed struct {
     @"old-style-flexible-struct": Kind = .default,
     @"gnu-zero-variadic-macro-arguments": Kind = .default,
     @"main-return-type": Kind = .default,
+    @"expansion-to-defined": Kind = .default,
 };
 
 const messages = struct {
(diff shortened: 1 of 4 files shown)


ehaas:
This does not handle the case of `defined` used as a unary operator as part of an expansion; handling that I think will be a bit trickier.

Vexu:
> This does not handle the case of `defined` used as a unary operator as part of an expansion; handling that I think will be a bit trickier.

This also produces the wrong result for:
```c
#define FOO_IS_DEFINED defined(FOO)
#define FOO

FOO_IS_DEFINED
```
I think it would be better to handle this in `Preprocessor.expr`.

PR #379, opened 2022-08-14
Preprocessor: handle macros which expand to `defined`

### LINKED ISSUE #337: Macro expansion which produces defined(X)

ehaas:
Since `defined` needs to look at its argument before expansion
happens, it needs to be special cased in expandMacroExhaustive.

It can't be a special function-like macro because that will cause
incorrect expansion outside of preprocessor expression contexts.
It can't be handled solely in `Preprocessor.expr` because the
argument will have been expanded by that point.

Closes #337

### DIFF
--- a/src/Preprocessor.zig
+++ b/src/Preprocessor.zig
@@ -693,17 +693,16 @@ fn handleKeywordDefined(pp: *Preprocessor, macro_tok: *Token, tokens: []const To
         return it.i;
     };
     switch (first.id) {
-        .identifier, .extended_identifier => {
-            macro_tok.id = if (pp.defines.contains(pp.expandedSlice(first))) .one else .zero;
-            return it.i;
-        },
         .l_paren => {},
         else => {
-            try pp.comp.diag.add(.{
-                .tag = .macro_name_must_be_identifier,
-                .loc = first.loc,
-                .extra = .{ .str = pp.expandedSlice(first) },
-            }, first.expansionSlice());
+            if (!first.id.isMacroIdentifier()) {
+                try pp.comp.diag.add(.{
+                    .tag = .macro_name_must_be_identifier,
+                    .loc = first.loc,
+                    .extra = .{ .str = pp.expandedSlice(first) },
+                }, first.expansionSlice());
+            }
+            macro_tok.id = if (pp.defines.contains(pp.expandedSlice(first))) .one else .zero;
             return it.i;
         },
     }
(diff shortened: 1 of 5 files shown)


src/Preprocessor.zig
```diff
@@ -689,6 +683,59 @@ fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) MacroError!bool {
...
+/// Returns the number of tokens consumed
+fn handleKeywordDefined(pp: *Preprocessor, macro_tok: *Token, tokens: []const Token, eof: RawToken) !usize {
+    std.debug.assert(macro_tok.id == .keyword_defined);
+    var it = TokenIterator.init(tokens);
+    const first = it.nextNoWS() orelse {
+        try pp.err(eof, .macro_name_missing);
+        return it.i;
+    };
+    switch (first.id) {
+        .identifier, .extended_identifier => {
```
Vexu, line 696: This has a bug similar to `__has_feature` that I fixed recently 7c179f6b9402b41e03bf60211bdecd183441d8f8:
```c
#define __const__ a

#if !defined __const__
#error isMacroIdentifier not used
#endif
```

src/Preprocessor.zig
```diff
@@ -1443,13 +1490,44 @@ fn removeExpandedTokens(pp: *Preprocessor, buf: *ExpandBuf, start: usize, len: u
...
+/// expression context (#if or #elif) or not.
+/// In a non-expression context it's just an identifier. Within a preprocessor
+/// expression it is a unary operator or one-argument function.
+const EvalContext = enum {
+    expr,
+    non_expr,
+};
+
+/// Helper for safely iterating over a slice of tokens while skipping whitespace
+const TokenIterator = struct {
```
Vexu, line 1503: This is nice, there are probably other places that could be improved to use this in the future.

Vexu (approved):
One small bug but otherwise this looks good!

PR #385, opened 2022-08-18
msvc bitfield layout

TwoClocks:
Bitfield layout for msvc.
a few small changes to get more of the tests to pass.

I think this is about the end of what I can do w/ layout w/o some of the outstanding bugs/features being fixed.


Vexu:
> Bitfield layout for msvc. a few small changes to get more of the tests to pass.
> 
> I think this is about the end of what I can do w/ layout w/o some of the outstanding bugs/features being fixed.

Thanks for all the work so far! I'll try to get to those bugs at some point but I think we already have the most common cases handled.

PR #393, opened 2022-08-31
Tokenizer: tokenize preprocessor numbers instead of typed ints/floats

ehaas:
This also includes the changes from #390, I can rebase afterwards if you want to merge that first.


src/number_affixes.zig
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    pub fn radix(prefix: Prefix) u8 {
+        return switch (prefix) {
+            .binary => 2,
+            .octal => 8,
+            .decimal => 10,
+            .hex => 16,
+        };
+    }
```
Vexu, line 52: `Prefix` enum tags should have these values and calls to this function should be replaced with `@enumToInt`.

src/number_affixes.zig
```diff
@@ -0,0 +1,150 @@
...
+    F,
+
+    I,
+    IU,
+    IUL,
+    IULL,
+    IL,
+    ILL,
+
+    IF,
```
Vexu, line 83: This part could really benefit from 
```
// zig fmt: off
// zig fmt: on
```

Vexu (approved):

Issue #387, opened 2022-08-22
Record Layout meta tracking issues.


TwoClocks:
A collection of issues related correct record layout for all platforms/compilers.

- [ ] #331 
- [ ] #332
- [ ] #361
- [ ] #367
- [ ] #381 
- [ ]  Do we need/want MinG support?

Issue #381, opened 2022-08-15
`Type.alingof` does not take target into account


TwoClocks:
This works running `clang-14 -c --target=i386-linux-gnu test.c`
```c
typedef long long b;

_Static_assert(sizeof(b) == 8, "");
_Static_assert(_Alignof(b) == 4, "");
```

but arocc thinks the alignment is 8

ehaas:
It looks like the `alignof` function in `Type.zig` is returning the wrong value - it returns `8` for everything except `msp430` (should be `4` for `i386`). From a quick glance at Zig's `type.zig` (which we aren't using) it looks like it would compute the wrong value there as well

PR #394, opened 2022-09-03
record layout test suite

TwoClocks:
Adds a test suite for record layout.

- Runs permutations of targets and tests
- Has expected failures for some tests due to various zig/arocc issues, including parser failures.
- will error if a test passes when it currently expects to fail
- Can run a specific platform/test for local debugging
- readme for more info


Vexu:
@TwoClocks I made some adjustments, could you double check that everything looks OK?

Also could we encode the expected compile errors in the test cases somehow or are they expected to change?

TwoClocks:
> @TwoClocks I made some adjustments, could you double check that everything looks OK?

Looks fine. But I'm not sure why you want to ignore warnings? The output isn't/shouldn't be displayed unless the results are unexpected. Are you seeing warnings spitting out?

Also, both `pack(32)` and `pack(0)` are valid, but compiler support may differ. `pack(0)` should behave the same as `pack()`. The spec just says "small powers of two", but never defines what "small" is.

> Also could we encode the expected compile errors in the test cases somehow or are they expected to change?

I've researched most of the errors in the test cases. I believe are all because of issues in arocc / zig #387. As fixes are merged, I'd expect more tests to pass. I'd expect all tests to pass for all platforms error free at some point.
As issues are addressed, it seemed easier to alter the runner than to alter the individual tests, as a single fix might change many tests. But I can put the expected failures in the tests if you feel that's better.

Vexu:
> Looks fine. But I'm not sure why you want to ignore warnings? The output isn't/shouldn't be displayed unless the results are unexpected. Are you seeing warnings spitting out?

Are you talking about setting some diagnostics off? Those were there before but weren't specific to the test cases causing them.

> Also, both `pack(32)` and `pack(0)` are valid, but compiler support may differ. `pack(0)` should behave the same as `pack()`. The spec just says "small powers of two", but never defines what "small" is.

What compiler supports `pack(32)`? Neither clang, GCC or MSVC seem to accept it. Also what about `pack(5)`, `pack(3)`, etc. in test 64?

> I've researched most of the errors in the test cases. I believe are all because of issues in arocc / zig #387. As fixes are merged, I'd expect more tests to pass. All tests for all targets should be error-free at some point. As issues are addressed, it seemed easier to alter the runner than to alter the individual tests, as a single fix might change many tests. For example, #381 is causing all targets to fail on 4 tests. But I can put the expected failures in the tests if you feel that's better.

All 980? Must be a lot of duplicates then.

TwoClocks:
> Are you talking about setting some diagnostics off? Those were there before but weren't specific to the test cases causing them.

I was referring to the changes to the test files. I was concerned you were seeing warnings.

> What compiler supports `pack(32)`? Neither clang, GCC or MSVC seem to accept it. Also what about `pack(5)`, `pack(3)`, etc. in test 64?

Clang accepts it.
Clang compiles 0064 w/o error (lots of warnings). Clang is doing record layout w/ those `pack` values. So while the values might be "invalid" record layout is happening. 0064 is testing that arocc's layout matches Clang's. 

The tests do not opine on what correct is. They are just testing that the layout matches the target:compiler.
While working on this I've run many tests through clang/gcc/msvc because I thought the test was invalid. While I have not done this will all the tests/targets, I have done it with many. I have not found a test that clang/gcc/msvc fail.

> All 980? Must be a lot of duplicates then.

Yeah. lots of duplicates. I test every combination of test file and target that's valid for a given test. Some tests are specific to MSVC, or SysV-like targets.
You can see from groups of asserts that a large number of targets are expected to be the same for any individual test.
I did not reduce the tests to "indicative" targets for each test because the grouping of results differs greatly across tests. There are targets that are always together across all tests, but not enough to speed up the running time much.
So I  test every perrmutation. I can try to reduce the number of targets if you like.

#381 causes all targets to fail when running test 0064. When it's fixed, I'd expect some subset of targets to pass. Then we'll need to explore the rermaining failures to see if they are a new issue, or an existing one.
This is what I have been doing for months now. Opening bugs, and when you fix them, altering the test results then going to see if the "unmasked" failures are new bugs or dupes. Rinse and repeat.

I'm open to any other process you think would be better.

Vexu:
> I was referring to the changes to the test files. I was concerned you were seeing warnings.

There should not be any changes to the test files, I originally modified few but later reverted them.

> Clang accepts it. Clang compiles 0064 w/o error (lots of warnings). Clang is doing record layout w/ those `pack` values. So while the values might be "invalid" record layout is happening. 0064 is testing that arocc's layout matches Clang's.
> 
> The tests do not opine on what correct is. They are just testing that the layout matches the target:compiler. While working on this I've run many tests through clang/gcc/msvc because I thought the test was invalid. While I have not done this will all the tests/targets, I have done it with many. I have not found a test that clang/gcc/msvc fail.

What do you mean by accept? For me clang gives a warning about invalid pack parameter and then does not modify the pack value. Either the file should be generated with `#pragma GCC diagnostic ignored "-Wignored-pragmas"` or those invalid `#pragma pack`s should not be added.

TwoClocks:
> What do you mean by accept? For me clang gives a warning about invalid pack parameter and then does not modify the pack value. Either the file should be generated with `#pragma GCC diagnostic ignored "-Wignored-pragmas"` or those invalid `#pragma pack`s should not be added.

`clang-14 -DONE ./test/records/0064_test.c`
This compiles, creates an object file, and no `_Static_assert` fire.  It does not error. I'd call this acceptance. The test checks arocc produces whatever clang produces.

If you want to make arocc error, while clang just warns, then we can remove that part of the test.

Issue #332, opened 2022-07-07
Alignment for nested `typedef` incorrect


TwoClocks:
```c
typedef int I1 __attribute__((aligned(8)));
typedef I1 I2 __attribute__((aligned(1)));

_Static_assert(sizeof(I2) == 4,"record I2 wrong sizeof");
_Static_assert(_Alignof(I2) == 1, "record I2 wrong alignment");
```

the alignment of I2 fails. It returns 8. It should be one. 

It looks like I2 has I1 attributes as well as its own. Seems like typedefs "roll up" attributes of the types they refer to.

typedef: 'attributed(int)'
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 8 }
 name: I1

typedef: 'attributed(int)'
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 8 }
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 1 }
 name: I2

TwoClocks:
Some more info on this. It looks like `I2` has `I1` attributes as well as it's own. I think this is the main issue.

```
typedef: 'attributed(int)'
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 8 }
 name: I1

typedef: 'attributed(int)'
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 8 }
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 1 }
 name: I2
```

PR #396, opened 2022-09-12
Fix for nested typedefs #332

TwoClocks:
#332
If a typedef has attributes and so does decl it refers to, only take the attributes from the "top" typedef. 
This just ignores all attributes on the referred decl. That might not be correct. I can stip out just the aligned attribs if you think it's incorrect.

This fix allows 1K-ish tests to pass the layout tests, and reduces the number of tests that fail on all platforms from 4 to 1.

### DIFF
--- a/src/Attribute.zig
+++ b/src/Attribute.zig
@@ -1099,12 +1099,19 @@ pub fn applyTypeAttributes(p: *Parser, ty: Type, attr_buf_start: usize, tag: ?Di
         => std.debug.panic("apply type attribute {s}", .{@tagName(attr.tag)}),
         else => try ignoredAttrErr(p, toks[i], attr.tag, "types"),
     };
-    const existing = ty.getAttributes();
-    if (existing.len == 0 and p.attr_application_buf.items.len == 0) return base_ty;
-    if (existing.len == 0) return base_ty.withAttributes(p.arena, p.attr_application_buf.items);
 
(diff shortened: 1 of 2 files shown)


Vexu (approved):
This likely isn't correct either but it seems better than what we currently do.

Vexu:
Based on Clang AST (via `clang-check --ast-dump <file>`) it seems that Clang never copies the previous attributes but instead has a `TypedefType` so that it can find the correct alignment that way:
```c
typedef int I1 __attribute__((aligned(8), unused));
typedef I1 I2;
```
```
|-TypedefDecl 0x55a116043680 <a.c:32:1, col:13> col:13 referenced I1 'int'
| |-BuiltinType 0x55a115fd6ea0 'int'
| |-AlignedAttr 0x55a1160436f8 <col:31, col:40> aligned
| | `-ConstantExpr 0x55a1160436d8 <col:39> 'int'
| |   |-value: Int 8
| |   `-IntegerLiteral 0x55a115fd7d20 <col:39> 'int' 8
| `-UnusedAttr 0x55a116043758 <col:43> unused
`-TypedefDecl 0x55a1160437b0 <line:33:1, col:12> col:12 I2 'I1':'int'
  `-TypedefType 0x55a116043780 'I1' sugar
    |-Typedef 0x55a116043680 'I1'
    `-BuiltinType 0x55a115fd6ea0 'int'
```
I think this patch should result in same behavior, at least for `aligned`, I'm not sure if other attributes behave the same.

PR #400, opened 2022-09-22
Sizeof zero len array

ehaas:


ehaas:
@TwoClocks I think these two commits should unblock some more layout work. I'm seeing some errors in `0053_test.c` but godbolt shows some of them too: Compare https://godbolt.org/z/zjq9E9nYe to `arocc -DMSVC -DONE --target=x86_64-windows-msvc test/records/0053_test.c` - we're showing an error on line 53 where godbolt is not (`_Alignof(struct A_extra_packed)` is zero for us) and we're not showing an error on line 54 whereas godbolt is (it thinks `sizeof(struct A_extra_required_alignment)` should be 2)

TwoClocks:
Yes. I noticed after the `typedef` fix that a large number of MSVC tests were only failing on the "extra" checks. Those are algorithmically generated. It was the next thing I was going to look at.

The rust repro does not state what the correct size/align should be for the generated permutations. I set it to agree with clang/gcc and assumed that was correct for other compiles. That seems like it's not the case.

The code to set the size/align for the "extra" permutations is here

https://github.com/TwoClocks/repr-c/blob/ba0064c6643a9ff54980656f448a9f7be48e136f/repc/arocc/src/main.rs#L445-L516

I will start trying to figure out the correct numbers for MSVC, and update the test files. If it's obvious to you what they should be, let me know.

TwoClocks:
This makes progress for MSVC. A bunch of tests parse and do layout correctly now.

I fixed one exposed layout bug w/ array sizes for MSVC, and updated the test to reflect the new working code.

I did a PR request to your branch. I'm not sure what what the correct way to append to your code is.

The layout test runner now ignores the "extra" tests for MSVC. I do not think the `.c` test files are correct, and I'm still digging into that.

PR #401, opened 2022-09-30
Type: `long long` and `double` alignment is 4 bytes on i386

### LINKED ISSUE #381: `Type.alingof` does not take target into account

ehaas:
Closes #381

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1167,7 +1167,7 @@ test "target size/align tests" {
     };
 
     try std.testing.expectEqual(@as(u64, 8), tt.sizeof(&comp).?);
-    try std.testing.expectEqual(@as(u64, 8), tt.alignof(&comp)); // TODO should be 4
+    try std.testing.expectEqual(@as(u64, 4), tt.alignof(&comp));
 
     const arm = std.Target.Cpu.Arch.arm;
     comp.target.cpu = std.Target.Cpu.Model.toCpu(&std.Target.arm.cpu.cortex_r4, arm);
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -2186,10 +2186,12 @@ pub fn fatalNoSrc(diag: *Diagnostics, comptime fmt: []const u8, args: anytype) e
 
 pub fn render(comp: *Compilation) void {
     if (comp.diag.list.items.len == 0) return;
-    var m = MsgWriter.init(comp.diag.color);
+    var m = defaultMsgWriter(comp);
     defer m.deinit();
-
-    renderExtra(comp, &m);
+    renderMessages(comp, &m);
+}
+pub fn defaultMsgWriter(comp: *const Compilation) MsgWriter {
+    return MsgWriter.init(comp.diag.color);
 }
 
 /// This is a workaround for a stage1 bug when constructing an anonymous struct with a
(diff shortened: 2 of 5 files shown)


TwoClocks:
I think there are a few things going on here :

1) It looks like this fix breaks MSVC for i386/32-bit targets . This code works on master currently and in compiler explore for x86 MSVC
```c
typedef struct {
    long long c;
} X;
_Static_assert(sizeof(X) == 8, "");
```
https://godbolt.org/z/T6aqcro9M
It fails in the `i386-align` branch now w/ the cmd line `arocc --emulate=msvc --target=i386-windows-msvc`.

This fix does seem to make gcc/clang 32-bit platforms layout correctly.

2) The output for MSVC targets is noisy and misleading. All the `_extra_` records fail for MSVC generally, so they are excluded when checking the results until I can figure out why they are wrong. When displaying errors during test, they overwhelm the output. I should filter them out for MSVC errors, so the results show meaningful differences.

If you are interested, here is the conversation I'm havaing w/ mahkoh about the MSVC `_extra_` records : https://github.com/mahkoh/repr-c/issues/4

I will clean up the output and get rid of all the noise errors, and update the gcc/clang tests to reflect the new code.
At first blush, I don't think the i386 MSVC failures are related to layout. Seems like a sizeof or something. But I can look into that as well if it's not obvious to you what it is.

ehaas:
I don't get errors with the code you posted, on `master` or on `i386-align`:
`cat test.c`
```c
typedef struct {
    long long c;
} X;
_Static_assert(sizeof(X) == 8, "");
```

```
➜  ~ arocc --emulate=msvc --target=i386-windows-msvc test.c
fatal error: unsupported target i386-windows-msvc, currently only x86-64 elf is supported
```

TwoClocks:
hmm. It doesn't seem to work for me.

```
❯ git --no-pager log -1
commit 906a34ae6d32637b5b989570aa3c67be7ef52c0c (HEAD -> i386-align, ehass/i386-align)
Author: Evan Haas <evan@lagerdata.com>
Date:   Thu Sep 29 23:14:59 2022 -0700

    Type: `long long` and `double` alignment is 4 bytes on i386

    Closes #381
❯ zig build
❯ cat test.c
typedef struct {
    long long c;
} X;

_Static_assert(sizeof(X) == 8, "");
_Static_assert(_Alignof(X) == 8, "");


❯ ./zig-out/bin/arocc --emulate=msvc --target=i386-windows-msvc test.c
./test.c:6:1: error: static assertion failed ""
_Static_assert(_Alignof(X) == 8, "");
^
1 error generated.
❯ hostnamectl
 Static hostname: deskone
       Icon name: computer
      Machine ID: 5ee5d758161c4f5f87839cdb6a0885d8
         Boot ID: 2f47d82a5cca4fb8a40fdae48e9cdb15
Operating System: Arch Linux
          Kernel: Linux 5.19.12-zen1-1-zen
    Architecture: x86-64
 Hardware Vendor: ASUS
  Hardware Model: System Product Name
Firmware Version: 2604
❯ zig version
0.10.0-dev.4212+34835bbbc
```

TwoClocks:
it doesn't seem to work for the linux CI for this PR either

```
Layout [143] 0002 | i386-i586-windows-msvc:Msvc | ONE... 
No errors expected for i386-i586-windows-msvc:Msvc 0002 Found:record_runner.ExpectedFailure{ .parse = false, .layout = true, .extra = false, .offset = false }
/home/runner/work/arocc/arocc/test/records/0002_test.c:92:1: error: static assertion failed ""
_Static_assert(_Alignof(X) == 8, "");
^
```

ehaas:
I missed the alignment assert - I updated the code to use 8 byte alignment on windows and 4 byte on other platforms for `i386`

PR #402, opened 2022-10-18
Layout: size and alignment adjustments, primarily for .arm arch

ehaas:
Match Zig's logic by using the size of the type as its alignment, in most cases. Fix `long` and `unsigned long` size on arm-ios.

`long long` and `unsigned long long` have 4-byte alignment on arm-ios even though they are 8 bytes in size.


Vexu:
That expected failure list is shrinking at a very nice pace!

PR #405, opened 2022-10-19
Update alignment of built-in types for avr

ehaas:


Vexu:
This should probably wait for https://github.com/ziglang/zig/pull/13257

ehaas:
Looks like some of my values were wrong anyway...not sure what's going on with my local copy of clang

PR #408, opened 2022-11-02
Layout fixes for avr/msp430

ehaas:
Fixes signed integer compare, and layout on avr/msp430 (clang and gcc disagree on `short` alignment on avr; GCC seems to be the compiler of record since clang's support is experimental https://llvm.org/doxygen/md_lib_Target_AVR_README.html) - should we try to emulate clang? 

https://godbolt.org/z/45Ezjzh8T


Vexu:
> should we try to emulate clang?

We could emulate both but since clang support is experimental we probably shouldn't.

PR #411, opened 2022-11-05
Layout fixes for sparc and mips

ehaas:
Updates some default alignments and pointer sizes.

Changes some test triples that were erroneously using `eabi` instead of `none` (no apparent effect on the passing tests)


ehaas:
Seems like `ptrBitWidth` can depend on other details of the target than just the CPU arch - do you think stdlib should move that function onto `std.Target` instead of where it is now on `Target.Cpu.Arch`?

Vexu:
> Seems like `ptrBitWidth` can depend on other details of the target than just the CPU arch - do you think stdlib should move that function onto `std.Target` instead of where it is now on `Target.Cpu.Arch`?

Maybe, probably worth asking others too since pointer size can also depend on address space.

Issue #367, opened 2022-08-04
MSVC isn't parsing `__declspec`?


TwoClocks:
for the following struct
```c
__declspec(align(2)) typedef struct {
    int a;
#ifdef MSVC
} X;
```
`ty.requestedAlignment()` returns null.

Vexu:
```sh-session
$ arocc a.c --verbose-ast --target=x86_64-windows-msvc
struct_decl_two: 'struct (anonymous struct at a.c:1:30)'
  record_field_decl: 'int'
   name: a

typedef: 'attributed(struct (anonymous struct at a.c:1:30))'
 attr: aligned alignment: Attribute.Alignment{ .node = Tree.NodeIndex.none, .requested = 2 }
 name: X
```
Looks fine to me?

TwoClocks:
I see it in `--verbose-ast`, but it's not there by the time layout is called.

I altered the layout compute call like this:
```rs
pub fn compute(ty: *Type, comp: *const Compilation, pragma_pack: ?u8) void {
    const mapper = comp.string_interner.getSlowTypeMapper();
    const name = mapper.lookup(ty.getRecord().?.name);
    std.debug.print("struct {s}\n", .{name});
    const attr = ty.getAttributes();
    std.debug.print("{any}:{}\n", .{ attr, attr.len });
```

and run `arocc --emulate=msvc test.c` where `test.c` is
```c
__declspec(align(2)) typedef struct { int a; } X;
__declspec(align(2)) struct non_typedef { int a; };
```

I get this output:

```
struct __va_list_tag
{  }:0
struct (anonymous struct at test.c:1:30)
{  }:0
struct non_typedef
{  }:0
./test.c:2:12: warning: attribute 'aligned' is ignored, place it after "struct" to apply attribute to type dec
laration [-Wignored-attributes]
__declspec(align(2)) struct non_typedef { int a; };
           ^
1 warning generated.
```

TwoClocks:
it's in the ast, but it's not making it to layout.
Here is the msvc and gcc versions of the same test.

`test1.c`
```c
__declspec(align(2)) typedef struct { int a; } X;
```
`test2.c`
```c
typedef struct { int a; } __attribute__((aligned(2))) X;
```

I alter the the beginning of the layout code like this
```rs
pub fn compute(ty: *Type, comp: *const Compilation, pragma_pack: ?u8) void {
    const mapper = comp.string_interner.getSlowTypeMapper();
[...]

PR #415, opened 2022-11-08
Implement MinGW layout

ehaas:
Also a small fix for s390x-linux-gnu

All GCC layout tests (besides 0044) passing now :)


Vexu:
Great progress!
>All GCC layout tests (besides 0044) passing now :)

Is there an issue for its blocker? The meta issue only mentions an MSVC issue.

ehaas:
Looks like we don't but it might be similar to / related to #367. This fails because arocc thinks the alignment of `Bar` is 4:


```
/* --target=x86_64-linux-gnu */

struct Foo {
    int c;
};

typedef struct Foo Bar __attribute__((aligned(8)));

_Static_assert(_Alignof(Bar) == 8, "");
```

PR #418, opened 2022-11-15
Parser: don't allow __declspec attributes after declarator in initDeclarator

### LINKED ISSUE #412: MSVC does not allow attributes after non-records.
MSVC will not parse this:
```c
typedef int Int __declspec(align(8)) ;
```
but this is fine
```c
typedef __declspec(align(8))  int Int;
```

AROC is fine with either.

I think this is related to #367
MSVC seems to apply attributes differently than gcc/clang.

ehaas:
Closes #412


Vexu (approved):
We could add a nice helpful error message and skip all the tokens but that isn't strictly necessary.

src/Parser.zig
```diff
@@ -1394,12 +1394,24 @@ fn gnuAttribute(p: *Parser) !bool {
...
     while (true) {
         if (try p.gnuAttribute()) continue;
         if (try p.c2xAttribute()) continue;
-        if (try p.msvcAttribute()) continue;
+        const maybe_declspec_tok = p.tok_i;
+        if (try p.msvcAttribute()) {
+            if (declarator_name) |name_tok| {
+                try p.errTok(.declspec_not_allowed_after_declarator, maybe_declspec_tok);
+                try p.errTok(.declarator_name_tok, name_tok);
+                p.attr_buf.len = p.attr_buf.len - 1;
```
Vexu, line 1411: I don't think this is correct, `msvcAttribute` can add multiple attributes so it should save the len before calling it.

PR #428, opened 2022-11-21
layout tests: skip int128 tests for other unsupported platforms

### LINKED ISSUE #421: Record layout tests use `__int128` on unsupported targets
I added an error for using `__int128` on targets that don't support it in 4f675179b8f932623e1a5e2975fafc39c354d631 but it doesn't pass tests since `test/records/0050_test.c` uses `__int128` and runs it on all targets.

Since the file is auto generated this should be fixed in the tool that generates it.

ehaas:
Exclude:
	mipsisa64r6el-unknown-linux-gnuabi64
	mipsisa64r6-unknown-linux-gnuabi64
	sparcv9-sun-solaris

Closes #421


ehaas:
Sent PR upstream https://github.com/TwoClocks/repr-c/pull/5

PR #432, opened 2022-11-22
C23 keywords

ehaas:
Note: this is based on #431, I can rebase or edit as necessary 

Add `alignas`, `alignof`, `bool`, `static_assert`, and `thread_local` as keywords. `true` and `false` will be done separately because they add a new syntax term `predefined constant` that can also be used to implement `nullptr`; plus they'll need a bit more preprocessor support to ensure `true` works as intended.

### DIFF
--- a/include/stdalign.h
+++ b/include/stdalign.h
@@ -1,9 +1,11 @@
 /* <stdalign.h> for the Aro C compiler */
 
 #pragma once
+#if __STDC_VERSION__ < 202311L
 
 #define alignas _Alignas
 #define alignof _Alignof
 
 #define __alignas_is_defined 1
 #define __alignof_is_defined 1
+#endif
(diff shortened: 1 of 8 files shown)


ehaas:
From reading this: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2654.pdf it looks like we are not required to provide the new keywords as macros. It seems like clang15 does not when in c2x mode, but I don't know if that will change. So it seems like we should either allow macros with the new keyword names (e.g. allow `#define bool whatever`) or we should also provide them as self-defined macros (e.g. also provide `#define bool bool` as part of the startup code and not allow redef or undef)

Vexu:
Clang trunc doesn't seem to have any special error for it. I think a good enough solution would be to not add any error for now but add this pedantic warning that we're currently missing.
```
warning: keyword is hidden by macro definition [-Wkeyword-macro]
#define bool _Bool
        ^
```

PR #433, opened 2022-11-23
Attribute: handle typedef decreasing record alignment

ehaas:
This addresses the GCC/clang side of #367. It's also a pretty ugly hack so feel free to close it if you'd prefer to handle it some other way.  Short of reworking the entire Type system, another way I think it could be handled would be adding a `?u29` alignment field to `Type.Attributed` to store typedef alignments

The fundamental problem is that after parse time there's no way to distinguish "struct with attributes" from "typedef which adds attributes to a struct" . It's relevant because attributes directly on a struct have a different effect on alignment than attributes on a typedef.

```c
struct A {
    int x;
};
struct __attribute__((aligned(2))) B {
    int x;
};
typedef __attribute__((aligned(2))) struct A Aligned_A;

_Static_assert(_Alignof(struct A) == 4, "");
_Static_assert(_Alignof(struct B) == 4, "");
_Static_assert(_Alignof(Aligned_A) == 2, "");
```

Both `struct B` and `Aligned_A` are represented as "attributed record", where the record has a single int field, but they have different alignment because the alignment on a typedef is allowed to be lower than the natural alignment.

### DIFF
--- a/src/Attribute.zig
+++ b/src/Attribute.zig
@@ -1055,6 +1055,18 @@ pub fn applyFieldAttributes(p: *Parser, field_ty: *Type, attr_buf_start: usize) 
     return p.arena.dupe(Attribute, p.attr_application_buf.items);
 }
 
+/// Handles the special case of `typedef` being used to decrease the alignment of a record
+/// after its layout has been computed
+pub fn applyTypedefAttributes(p: *Parser, ty: Type, attr_buf_start: usize) !Type {
+    const attrs = p.attr_buf.items(.attr)[attr_buf_start..];
+    const attributed = try applyTypeAttributes(p, ty, attr_buf_start, null);
+    const typedef_align = Type.annotationAlignment(p.comp, attrs) orelse return attributed;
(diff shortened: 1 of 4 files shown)


Vexu:
> The fundamental problem is that after parse time there's no way to distinguish "struct with attributes" from "typedef which adds attributes to a struct" .

Sounds like this could be solved by adding a `typedef` type like Clang has.

Vexu:
It can't get much worse than it already is with attributed types and typeof types, working with them just needs to be made less painful somehow.

ehaas:
One option would be to remove typeof-type and typeof-expr types as separate things, and just resolve them at parse time. Would have a negative impact on user-facing error messages and obviously lose info that's in the AST, though.

Vexu:
> One option would be to remove typeof-type and typeof-expr types as separate things, and just resolve them at parse time. Would have a negative impact on user-facing error messages and obviously lose info that's in the AST, though.

But that would hardly reduce the complexity since we still need attributed types.

TwoClocks:
This might be related
#437

PR #434, opened 2022-11-23
driver: Add `-fsyntax-only` option

ehaas:
This will check for syntax errors and evaluate static asserts and then exit


src/main.zig
```diff
@@ -92,6 +92,7 @@ const usage =
     \\                          Set limit on how many macro expansion traces are shown in errors (default 6)
     \\  -fshort-enums           Use the narrowest possible integer type for enums
     \\  -fno-short-enums        Use "int" as the tag type for enums
+    \\  -fsyntax-only           Only run the preprocessor, parser, and semantic analysis stages
```
Vexu, line 95: GCC also accepts `-fno-syntax-only`

src/main.zig
```diff
@@ -466,9 +469,10 @@ fn processSource(
     const prev_errors = comp.diag.errors;
     comp.renderErrors();
 
-    if (comp.diag.errors != prev_errors) {
-        if (fast_exit) exitWithCleanup(link_objects.items, temp_file_count.*, 1);
-        return; // do not compile if there were errors
+    if (comp.diag.errors != prev_errors or comp.only_syntax) {
+        const has_errors = comp.diag.errors != prev_errors;
+        if (fast_exit) exitWithCleanup(link_objects.items, temp_file_count.*, @boolToInt(has_errors));
+        return; // do not compile if there were errors or we are only checking syntax
```
Vexu, line 475: I'd prefer to keep these branches separate for simplicity:
```suggestion
    if (comp.diag.errors != prev_errors) {
        if (fast_exit) exitWithCleanup(link_objects.items, temp_file_count.*, 1);
        return; // do not compile if there were errors
    }

    if (comp.only_syntax) {
        if (fast_exit) std.process.exit(0); // Not linking, no need for cleanup.
        return;
```

Vexu (approved):

PR #439, opened 2022-11-30
Parser: improve sizeof/alignof handling and errors

ehaas:
1. Issue a warning for sizeof/alignof `void`
2. Issue a warning for sizeof(decayed array)
3. Treat sizeof(incomplete_size_type) or alignof(non_alignable_type) as parse errors. This matches clang's behavior and lets us remove some hacky error-suppression code in static_assert handling

This is part of the groundwork for some other static_assert changes I'll be making (clang trunk no longer requires that static_assert expressions be *integer* constants: https://godbolt.org/z/oafvqTEjx) which is itself part of the work for adding C23 nullptr support.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -1004,7 +1004,9 @@ fn staticAssert(p: *Parser) Error!bool {
     if (str.node == .none) try p.errTok(.static_assert_missing_message, static_assert);
 
     if (res.val.tag == .unavailable) {
-        try p.errTok(.static_assert_not_constant, res_token);
+        if (res.ty.specifier != .invalid) {
+            try p.errTok(.static_assert_not_constant, res_token);
+        }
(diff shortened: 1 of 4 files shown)


src/Parser.zig
```diff
@@ -6016,8 +6021,12 @@ fn unExpr(p: *Parser) Error!Result {
                 try p.errTok(.alignof_expr, expected_paren);
             }
 
+            if (res.ty.is(.void)) {
+                try p.errStr(.pointer_arith_void, tok, "alignof");
+            }
             if (!res.ty.alignable()) {
                 try p.errStr(.invalid_alignof, expected_paren, try p.typeStr(res.ty));
+                return error.ParsingFailed;
```
Vexu, line 6029: I think this is a perfect place to use the new `invalid` type.

PR #440, opened 2022-11-30
Zig stdlib: update to changes in Allocator interface

Vexu:


ehaas:
I haven't had a chance to dig into the new allocator changes yet but will it have an impact on this: https://github.com/Vexu/arocc/blob/2280654d4bdff4c4e5bab4f2d2b651e2093784de/src/Tree.zig#L36

Vexu:
I don't think that should be affected since the array list still does a calculation for a more optimized capacity. The comment should be updated though.

PR #444, opened 2022-12-09
Builtin: Add C23 unreachable() support to frontend

ehaas:
Note: the `Type` returned by `Builtin.get` for `__builtin_unreachable` is currently incorrect since it doesn't include the `noreturn` attribute - do you want me to add that or will that be handled in by grabbing the associated `Builtin.Attributes` object?

### DIFF
--- a/include/stddef.h
+++ b/include/stddef.h
@@ -2,8 +2,7 @@
 
 #pragma once
 
-/* Todo: Set to 202311L once header is compliant with C23 */
-#define __STDC_VERSION_STDDEF_H__ 0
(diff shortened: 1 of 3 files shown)


Vexu:
> Note: the `Type` returned by `Builtin.get` for `__builtin_unreachable` is currently incorrect since it doesn't include the `noreturn` attribute

I don't think the `noreturn` attribute is read anywhere yet so it doesn't matter too much, the whole builtin function system will likely need to be redone since the current one doesn't feel too scalable so that can be fixed at the same time.

PR #447, opened 2022-12-27
Type: remove unused parameters to `combine`

ehaas:
These parameters don't seem to actually be used (just passed along to recursive calls to the same function). Not sure if they're there as a placeholder for future use or just a remnant of some earlier code


Vexu (approved):
> Not sure if they're there as a placeholder for future use or just a remnant of some earlier code

They were initially used for `todo` errors and then later for `func_cannot_return_array` errors which were moved to `validateCombinedType` in c28578b75bd052bb1f93c2dd9b11544817988516.

PR #449, opened 2022-12-30
WIP: Builtins frontend

ehaas:


ehaas:
Ok, I think this is finally ready for review! There's still lots of room for improvement and tons more work to do but I think it's in a reasonably usable state at this point.

src/Builtins.zig
```diff
@@ -1,93 +1,374 @@
...
+const Expanded = struct {
+    ty: Type,
+    builtin: BuiltinFunction,
 };
-const BuiltinMap = std.StringHashMapUnmanaged(Builtin);
 
-_builtins: BuiltinMap = .{},
-_params: []Type.Func.Param = &.{},
+/// A growable list with stable addresses for items
+const Cache = struct {
```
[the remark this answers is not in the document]
Vexu: That does sound much simpler, probably should make it just directly map to `Type`.

Vexu (approved):
Looks good, only issue I have with it is that it creates a small dependency on clang.

`BuiltinFunction.zig` should be marked as auto-generated in `.gitattributes`.

ehaas:
> Looks good, only issue I have with it is that it creates a small dependency on clang.

Do you think we should we vendor the .def files? Or not use them at all, and just define the builtins on an ad hoc basis as we need them?

If we strip out the comments they seem more like a "collection of facts" as opposed to a creative expression, but I'm not an expert on copyright law :)

Vexu:
I don't have a better solution in mind at the moment and doing it like this saves us a lot of work so it's fine for now.

src/Compilation.zig
```diff
@@ -400,12 +412,39 @@ fn generateBuiltinTypes(comp: *Compilation) !void {
 
     const va_list = try comp.generateVaListType();
 
+    const pid_t: Type = switch (os) {
+        .haiku => .{ .specifier = .long },
+        // Todo: pid_t is required to "a signed integer type"; are there any systems
```
matu3ba, line 417: This sounds like a 16 bit Kernel https://de.wikipedia.org/wiki/Embeddable_Linux_Kernel_Subset, where `sizeof(int)=16`. See also https://stackoverflow.com/questions/11438794/is-the-size-of-c-int-2-bytes-or-4-bytes.

ehaas: If `int` is 16 bits, then `short` will also be 16 bits since that's the minimum size and it can't be larger than `int`. So I think `short` and `int` would be equivalent in that case; the question is whether there are 32 bit systems with 16 bit pid_t since technically that seems to be allowed.

matu3ba: After searching for a while I did not find any system, who would do that.
However, there is https://stackoverflow.com/questions/7829058/how-to-run-16-bit-code-on-32-bit-linux, for example via https://github.com/lkundrak/dev86/tree/master/elksemu or vm86 (`man vm86`) or qemu /simulation.

Sounds very niche though for (very) old software.

PR #452, opened 2023-01-13
Parser: Treat compound literals as a suffix expression

### LINKED ISSUE #451: Parse error: Field access of compound literal
The following works in GCC and clang; it seems to be the only error preventing the frontend from parsing `zig2.c`:

```c
struct B {
   _Bool payload;
};

int foo(void) {
   return (struct B) {.payload = 1}.payload;
}
```

```
./test.c:6:36: error: expected ';', found '.'
   return (struct B) {.payload = 1}.payload;
                                   ^
```

ehaas:
The C grammar treats them as suffix expressions instead of cast expressions

Closes #451

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -5667,18 +5667,9 @@ fn castExpr(p: *Parser) Error!Result {
         try p.expectClosing(l_paren, .r_paren);
 
         if (p.tok_ids[p.tok_i] == .l_brace) {
-            // compound literal
-            if (ty.isFunc()) {
-                try p.err(.func_init);
-            } else if (ty.is(.variable_len_array)) {
-                try p.err(.vla_init);
-            } else if (ty.hasIncompleteSize() and !ty.is(.incomplete_array)) {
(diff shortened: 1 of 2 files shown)


Vexu:
Is there a reason to not make it a `primaryExpr`?

Vexu:
I guess this does match the official grammar better:

```
postfix-expression:
    primary-expression
    postfix-expression [ expression ]
    postfix-expression ( argument-expression-listopt )
    postfix-expression . identifier
    postfix-expression -> identifier
    postfix-expression ++
    postfix-expression -
    ( type-name ) { initializer-list }
    ( type-name ) { initializer-list , }
    
argument-expression-list:
    assignment-expression
    argument-expression-list , assignment-expression
```

Vexu (approved):

ehaas:
Functionally I think it would be the same if we put it there; the only potential problem would be if `primaryExpr` is ever called from anywhere else (currently it's only called from one place) and that new calling location doesn't allow the possibility of a compound literal.

PR #453, opened 2023-01-16
Compilation: add basic float.h support

ehaas:


include/float.h
```diff
@@ -0,0 +1,123 @@
...
+#undef DBL_MANT_DIG
+#define DBL_MANT_DIG __DBL_MANT_DIG__
+
+#undef LDBL_MANT_DIG
+#define LDBL_MANT_DIG __LDBL_MANT_DIG__
+
+#if __STDC_VERSION__ >= 199901L
+#undef DECIMAL_DIG
+#define DECIMAL_DIG __DECIMAL_DIG__
+#endif  // __STDC_VERSION__ >= 199901L
```
Vexu, line 20: The preprocessor should just ignore the extra tokens but it'd still be better to use C style `/* */` comments in these headers.

Vexu (approved):

PR #456, opened 2023-01-22
Fix some promotion edge cases

ehaas:
Promote char/short to int before converting to floating point if use in FP arithmetic.

Promote bitfields to int if int is large enough to contain all the values (keeping some gcc/clang errata)

Promote __fp16 to float for arithmetic, and disallow it as a parameter type

Use fp_eval_method setting to promote intermediate floating point results

Avoid extra implicit cast when converting between real and complex of same underlying type in an init expression.

### DIFF
--- a/src/Driver.zig
+++ b/src/Driver.zig
@@ -75,6 +75,8 @@ pub const usage =
     \\  -fmacro-backtrace-limit=<limit>
     \\                          Set limit on how many macro expansion traces are shown in errors (default 6)
     \\  -fnative-half-type      Use the native half type for __fp16 instead of promoting to float
+    \\  -fnative-half-arguments-and-returns
+    \\                          Allow half-precision function arguments and return values
     \\  -fshort-enums           Use the narrowest possible integer type for enums
     \\  -fno-short-enums        Use "int" as the tag type for enums
     \\  -fsyntax-only           Only run the preprocessor, parser, and semantic analysis stages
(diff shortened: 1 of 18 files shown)


src/Driver.zig
```diff
@@ -72,6 +74,7 @@ pub const usage =
     \\                          Disallow '$' in identifiers
     \\  -fmacro-backtrace-limit=<limit>
     \\                          Set limit on how many macro expansion traces are shown in errors (default 6)
+    \\  -fnative-half-type      Use the native half type for __fp16 instead of promoting to float
```
Vexu, line 77: Neither clang nor GCC seem to expose this option, should we?

src/SymbolStack.zig
```diff
@@ -287,6 +287,9 @@ pub fn defineParam(s: *SymbolStack, p: *Parser, name: StringId, ty: Type, tok: T
             else => {},
         }
     }
+    if (ty.is(.fp16)) {
+        try p.errStr(.suggest_pointer_for_invalid_parameter_type, tok, "__fp16");
+    }
```
Vexu, line 292: ARM and aarch64 allow `__fp16` returns and arguments.

src/Driver.zig
ehaas: It's hidden behind `-Xclang`  https://godbolt.org/z/x739v9z7v - do you think we should have a similar flag for non-standard options like that? I don't think GCC has an option for native half. There's also a similar `-Xclang` flag `-fnative-half-arguments-and-returns` which allows `__fp16` as an argument/return type on all platforms

Vexu: Ah, I didn't realize that's how they do it. I guess this is fine for now then?

Vexu (approved):

PR #460, opened 2023-02-16
Add C23 _BitInt suffix for integer literals

ehaas:
Currently only supports 64-bit or smaller literals.  We'll need a way to handle larger values - any thoughts on that? 128-bit `int` field in `Value.data`? make `Value.data.int` a union w/ 64 bit int or pointer to bigint?

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -24,7 +24,7 @@ pub const Error = error{
     FatalError,
 } || Allocator.Error;
 
-pub const BitIntMaxBits = 128;
+pub const bit_int_max_bits = 128;
 
 gpa: Allocator,
 sources: std.StringArrayHashMap(Source),
@@ -349,7 +349,7 @@ pub fn generateBuiltinMacros(comp: *Compilation) !Source {
     // try comp.generateIntMax(w, "UINTPTR", comp.types.size);
 
     // int widths
-    try w.print("#define __BITINT_MAXWIDTH__ {d}\n", .{BitIntMaxBits});
+    try w.print("#define __BITINT_MAXWIDTH__ {d}\n", .{bit_int_max_bits});
 
     // sizeof types
     try comp.generateSizeofType(w, "__SIZEOF_FLOAT__", .{ .specifier = .float });
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -24,6 +24,8 @@ pub const Error = error{
     FatalError,
 } || Allocator.Error;
 
+pub const BitIntMaxBits = 128;
+
 gpa: Allocator,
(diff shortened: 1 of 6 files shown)


src/Compilation.zig
```diff
@@ -24,6 +24,8 @@ pub const Error = error{
     FatalError,
 } || Allocator.Error;
 
+pub const BitIntMaxBits = 128;
```
Vexu, line 27: Should be snake case.

src/Parser.zig
```diff
@@ -7280,59 +7281,66 @@ fn parseInt(p: *Parser, prefix: NumberPrefix, buf: []const u8, suffix: NumberSuf
...
-            'a'...'z' => c - 'a' + 10,
-            '\'' => continue,
-            else => unreachable,
-        };
+    var res: Result = undefined;
+    if (suffix.isBitInt()) {
+        try p.errStr(.pre_c2x_compat, tok_i, "'_BitInt' suffix for literals");
+        try p.errTok(.bitint_suffix, tok_i);
+        res = try p.bitInt(base, buf, suffix, tok_i);
+    } else {
```
Vexu, line 7289: Could you move this else branch to a new function too?

src/Type.zig
```diff
@@ -683,21 +683,24 @@ pub fn integerPromotion(ty: Type, comp: *Compilation) Type {
...
-            .complex_uint, .complex_long, .complex_ulong, .complex_long_long, .complex_ulong_long,
-            .complex_int128, .complex_uint128, .bit_int, .complex_bit_int => specifier,
-            // zig fmt: on
-            .typeof_type => return ty.data.sub_type.integerPromotion(comp),
-            .typeof_expr => return ty.data.expr.ty.integerPromotion(comp),
-            .attributed => return ty.data.attributed.base.integerPromotion(comp),
-            .invalid => .invalid,
-            else => unreachable, // not an integer type
+    return switch (specifier) {
+        .bit_int, .complex_bit_int => .{ .specifier = .bit_int, .data = .{ .int = ty.data.int } },
```
Vexu, line 687: Is it correct to return `.bit_int` for `.complex_bit_int`? And could you move this extract this switch as a standalone statement like the if before it.

Vexu (commented):
> Currently only supports 64-bit or smaller literals. We'll need a way to handle larger values - any thoughts on that? 128-bit `int` field in `Value.data`? make `Value.data.int` a union w/ 64 bit int or pointer to bigint?

BigInt is definitely the way to go. Adding it as a filed would be the easiest but it would be better to change it more extensively. Not sure if it'd be better to use an extern union like the Zig compiler does or use the interner.

PR #465, opened 2023-03-09
Add __builtin_types_compatible support

ehaas:


ehaas:
I'm adding the lhs and rhs arguments to __builtin_types_compatible_p as nodes with `.invalid` tag - this is only for dumping the AST, since I only need the type info. Would it be better to add a new tag type so `.invalid` isn't overloaded? Just use `undefined`? Or keep as-is?

Vexu:
I think it's fine, it could also use `null_stmt` if you want to avoid overloading `invalid` as long as it is documented.

PR #466, opened 2023-03-26
build: Zig build system update - remove RunStep.expected_term usage

ehaas:
Not sure if this is the right fix because I don't actually remember what `expected_term` used to be, but since we were setting it to null maybe it's not needed? With latest Zig it complains `no field named 'expected_term' in struct 'Build.RunStep'`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -96,16 +96,6 @@ pub fn build(b: *Build) !void {
     }
     exe.install();
 
-    const run_cmd = exe.run();
-    run_cmd.expected_term = null;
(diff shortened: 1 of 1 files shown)


Vexu:
Null used to mean any exit code but that has now been removed seemingly on purpose according to an issue that has seemingly been removed. That means that now if you do `zig build run -- a.c` and compiling `a.c` fails, it will fail the build step and spam your console with a bunch of useless information.

It'd be better to remove the run step completely now that it is useless.

Issue #430, opened 2022-11-21
C23 feature support checklist


Vexu:
Based on https://en.cppreference.com/w/c/23

- [x] `_Static_asert` with no message
- [x] `[[nodiscard]]`
- [x] `[[maybe_unused]]`
- [x] `[[deprecated]]`
- [x] Attributes
~~- [ ] IEEE 754 decimal floating-point types~~
    - #552
- [x] `[[fallthrough]]`
- [x] u8 character constants
- [x] Removal of function definitions without prototype
- [x] `[[nodiscard]]` with message
- [x] Unnamed parameters in function definitions
- [x] Labels before declarations and end of blocks
- [x] Binary integer constants
- [x] `__has_c_attribute` in preprocessor conditionals
- [x] `__has_c_attribute(vendor::attribute)` 
- [x] Allow duplicate attributes
~~- [ ] IEEE 754 interchange and extended types~~
    - #552
- [x] Digit separators
- [x] #elifdef and #elifndef
- [x] Type change of u8 string literals
- [x] `[[maybe_unused]]` for labels
- [x] `#warning`
- [x] Bit-precise integer types (`_BitInt`)
- [x] `[[noreturn]]`
- [x] Suffixes for bit-precise integer constants
- [x] `__has_include` in preprocessor conditionals
- [x] Removal of function declarations without prototype
- [x] Empty initializers
- [x] typeof ...
- [x] ... and typeof_unqual
- [x] New spelling of keywords
- [x] Predefined true and false
- [x] `[[unsequenced]]`
- [x] `[[reproducible]]`
- [x] Relax requirements for variadic parameter list
- [x] Type inference in object definitions (`auto`)
- [x] constexpr objects
- [x] nullptr
- [x] `#embed`
- [x] `#embed` parameters `limit`, `suffix`, `prefix`, `if_empty`
- [x] `__has_embed`
- [x] Storage-class specifiers for compound literals 
- [x] `__VA_OPT__`
- [x] Identifier Syntax using Unicode Standard Annex 31
- [x] Check identifier NFC normalized
- [x] Type change of u8 string literals
- [x] u8 character constants (already implemented, should be an error for older versions)

ehaas:
This list may also be helpful for test cases: https://htmlpreview.github.io/?https://icube-forge.unistra.fr/icps/c23-library/-/raw/main/README.html

Vexu:
@ehaas Do you want to handle the Unicode identifier change?

[...]

PR #467, opened 2023-04-28
Implement #elifdef and #elifndef

danielsan901998:
Implement C23 feature #elifdef and #elifndef from #430.
For testing added two new Preproccesor tests [#elifdef.c](https://github.com/Vexu/arocc/compare/master...danielsan901998:arocc:master?expand=1#diff-d863a9d1369bd4db5960cf805446a9c5a247128c4780dfc6088d0d595d2e5ab2) and [#elifdefc2x.c](https://github.com/Vexu/arocc/compare/master...danielsan901998:arocc:master?expand=1#diff-b705cf5e912d8d27e52ba76e1113f7d680dd3795c2ecf374d633794483cecb29) to check it is proccessed correctly with the -std=c2x flag
Also changed build.zig to compile with current master branch of zig.


src/Preprocessor.zig
```diff
@@ -365,6 +365,72 @@ fn preprocessExtra(pp: *Preprocessor, source: Source) MacroError!Token {
...
+                        if (if_level == 0) {
+                            try pp.err(directive, .elifdef_without_if);
+                            if_level += 1;
+                            if_kind.set(if_level, until_else);
+                        } else if (if_level == 1) {
+                            guard_name = null;
+                        }
+                        switch (if_kind.get(if_level)) {
+                            until_else => {
+                                const macro_name = (try pp.expectMacroName(&tokenizer)) orelse continue;
```
Vexu, line 378: I think a missing macro name should be considered `false` similarly to a missing expression with `#if`.

Vexu (approved):
Looks good! And thanks for updating this for the latest master.

src/Preprocessor.zig
danielsan901998: I have tested it with clang and you are right, i have also added a test for checking the missing macro name in the commit ef844b664b3ff4f6a025ae415669c94f2665ac22

danielsan901998: Clang generate this output:
```cpp
test/cases/#elifdefc2x_error.c:7:9: error: macro name missing
#elifdef
        ^
# 10 "test/cases/#elifdefc2x_error.c"
inttest/cases/#elifdefc2x_error.c:16:10: error: macro name missing
#elifndef
         ^

# 19 "test/cases/#elifdefc2x_error.c"
int
2 errors generated.
```
And Gcc genete this:
```cpp
# 10 "test/cases/#elifdefc2x_error.c"
int
# 19 "test/cases/#elifdefc2x_error.c"
int
```
I think the behaviour of gcc looks more correct, i'm going to commit to the pull request to reflect this.

danielsan901998: If there are not more issues, I think that with commit e1567309b0af3e27dad70a22e083aecedc5c3183 it should be ready to merge.

Vexu: It looks like they have the same output with clang missing a newline after the `int`.

PR #482, opened 2023-06-15
Parser: allow tentative top-level records and enums

ehaas:
Allow top-level definitions of symbols which have incomplete record or enum types, and only issue an error if the type is never completed.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -113,6 +113,13 @@ record_buf: std.ArrayList(Type.Record.Field),
 attr_buf: std.MultiArrayList(TentativeAttribute) = .{},
 attr_application_buf: std.ArrayListUnmanaged(Attribute) = .{},
 field_attr_buf: std.ArrayList([]const Attribute),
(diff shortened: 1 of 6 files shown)


src/Parser.zig
```diff
@@ -514,6 +521,42 @@ fn pragma(p: *Parser) Compilation.Error!bool {
...
 
+/// Issue errors for top-level definitions whose type was never completed.
+fn diagnoseIncompleteDefinitions(p: *Parser) !void {
+    @setCold(true);
+
+    const node_slices = p.nodes.slice();
+    const tags = node_slices.items(.tag);
+
+    var it = p.tentative_defs.iterator();
+    while (it.next()) |item| {
```
Vexu, line 532: Could this be made more efficient by looping over the decls and checking if there is an entry in the hash map?

Vexu (approved):
Nice, this looks like it could also be helpful for solving that `tentative_array` case.

src/Parser.zig
ehaas: Definitely...that way we only have to loop over the decls once instead of N times

PR #487, opened 2023-06-27
Zig update: Update cast builtins to single-argument variants

ehaas:


src/Compilation.zig
```diff
@@ -124,7 +124,7 @@ fn generateDateAndTime(w: anytype) !void {
 
     const day_names = [_][]const u8{ "Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun" };
     // days since Thu Oct 1 1970
-    const day_name = day_names[@truncate(u8, (epoch_day.day + 3) % 7)];
+    const day_name = day_names[@as(u8, @truncate((epoch_day.day + 3) % 7))];
```
Vexu, line 127: Is this `@as` required?

src/Compilation.zig
```diff
@@ -1297,10 +1297,10 @@ test "addSourceFromReader" {
             defer comp.deinit();
 
             var buf_reader = std.io.fixedBufferStream(str);
-            const source = try comp.addSourceFromReader(buf_reader.reader(), "path", @intCast(u32, str.len));
+            const source = try comp.addSourceFromReader(buf_reader.reader(), "path", @intCast(str.len));
 
             try std.testing.expectEqualStrings(expected, source.buf);
-            try std.testing.expectEqual(warning_count, @intCast(u32, comp.diag.list.items.len));
+            try std.testing.expectEqual(warning_count, @as(u32, @intCast(comp.diag.list.items.len)));
```
Vexu, line 1303: Shouldn't this get the type from `warning_count`?

src/Parser.zig
```diff
@@ -3344,7 +3344,7 @@ fn findScalarInitializer(p: *Parser, il: **InitList, ty: *Type, actual_ty: Type,
         }
         const struct_il = il.*;
         while (index < fields.len) : (index += 1) {
-            const field = fields[@intCast(u32, index)];
+            const field = fields[@as(u32, @intCast(index))];
```
Vexu, line 3347: Why was there a cast here in the first place?

src/Parser.zig
```diff
@@ -3401,7 +3401,7 @@ fn findAggregateInitializer(p: *Parser, il: **InitList, ty: *Type, start_index:
 
         const field_count = struct_ty.data.record.fields.len;
         if (index < field_count) {
-            ty.* = struct_ty.data.record.fields[@intCast(usize, index)].ty;
+            ty.* = struct_ty.data.record.fields[@as(usize, @intCast(index))].ty;
```
Vexu, line 3404: Ditto.

src/Source.zig
```diff
@@ -45,15 +45,15 @@ pub fn lineCol(source: Source, loc: Location) LineCol {
     var start: usize = 0;
     // find the start of the line which is either a newline or a splice
     if (std.mem.lastIndexOfScalar(u8, source.buf[0..loc.byte_offset], '\n')) |some| start = some + 1;
-    const splice_index = for (source.splice_locs, 0..) |splice_offset, i| {
+    const splice_index: u32 = for (source.splice_locs, 0..) |splice_offset, i| {
         if (splice_offset > start) {
             if (splice_offset < loc.byte_offset) {
                 start = splice_offset;
-                break @intCast(u32, i) + 1;
+                break @as(u32, @intCast(i)) + 1;
```
Vexu, line 52: Is this `@as` needed?

test/fuzz/fuzz_lib.zig
```diff
@@ -12,7 +12,7 @@ const Parser = aro.Parser;
 var fixed_buffer_mem: [10 * 1024 * 1024]u8 = undefined;
 
 export fn compile_c_buf(buf: [*]const u8, len: c_int) void {
-    compileSlice(buf[0..@intCast(usize, len)]) catch unreachable;
+    compileSlice(buf[0..@as(usize, @intCast(len))]) catch unreachable;
```
Vexu, line 15: Doesn't seem like `@as` should be needed.

src/Compilation.zig
ehaas: It looks like it's not able to see through the `@TypeOf(expected)` where `expected` is `anytype` - without the cast I get:

```
src/Compilation.zig:1303:56: error: @intCast must have a known result type
            try std.testing.expectEqual(warning_count, @intCast(comp.diag.list.items.len));
                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/Compilation.zig:1303:56: note: result type is unknown due to anytype parameter
src/Compilation.zig:1303:56: note: use @as to provide explicit result type
```

Vexu: That seems like a bug.

src/Compilation.zig
ehaas: Removed the cast and the truncate

src/Parser.zig
ehaas: `index` is `u64` so it's needed for 32-bit builds

src/Parser.zig
ehaas: This one might not actually be needed; I'll push an update and see if it still builds

src/Source.zig
ehaas: Compilation fails with unknown result type without it

src/Compilation.zig
Vexu: Seems like the truncate is needed for 32-bit builds.

src/Parser.zig
Vexu: Right, forgot about that.

Vexu: `@as` should still be unnecessary.

test/fuzz/fuzz_lib.zig
ehaas: Removed. This file has bitrotted a bit; I haven't run the fuzzer in a while.

src/Parser.zig
ehaas: Removed this and the other one in `findScalarInitializer` and seems to have worked

src/Compilation.zig
Vexu: The `@as` should still be unnecessary.

ehaas: Without it, on a 64bit system I get

```
src/Compilation.zig:127:32: error: destination type 'usize' has more bits than source type 'u47'
    const day_name = day_names[@truncate((epoch_day.day + 3) % 7)];
                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/Compilation.zig:127:32: note: destination type has 64 bits
src/Compilation.zig:127:62: note: operand type has 47 bits
    const day_name = day_names[@truncate((epoch_day.day + 3) % 7)];
                                         ~~~~~~~~~~~~~~~~~~~~^~~
```

ehaas: Maybe we should just be using `@intCast` in the first place instead of truncate

Issue #490, opened 2023-07-08
Cannot build `ReleaseFast` or `ReleaseSmall` on linux


ehaas:
I suspect this is a Zig bug but I'm not sure how to figure out what's going on. Details:

```
ehaas@system76-pc:~/source/arocc$ zig version
0.11.0-dev.3947+89396ff02
ehaas@system76-pc:~/source/arocc$
ehaas@system76-pc:~/source/arocc$ uname -a
Linux system76-pc 5.18.10-76051810-generic #202207071639~1659403207~20.04~cb5f582~dev-Ubuntu SMP PREEMPT_DY x86_64 x86_64 x86_64 GNU/Linux
ehaas@system76-pc:~/source/arocc$
ehaas@system76-pc:~/source/arocc$ zig build -Doptimize=ReleaseFast
zig build-exe arocc ReleaseFast native: error: the following command terminated unexpectedly:
/home/ehaas/.local/zig-linux-x86_64-0.11.0-dev.3947+89396ff02/zig build-exe /home/ehaas/source/arocc/src/main.zig -lc -OReleaseFast --cache-dir /home/ehaas/source/arocc/zig-cache --global-cache-dir /home/ehaas/.cache/zig --name arocc --mod build_options::/home/ehaas/source/arocc/zig-cache/c/f27f2c9c0c5353d14b6819f0936d41a3/options.zig --mod zig::/home/ehaas/source/arocc/deps/zig/lib.zig --deps build_options,zig --listen=-
Build Summary: 1/4 steps succeeded; 1 failed (disable with --summary none)
install transitive failure
└─ install arocc transitive failure
   └─ zig build-exe arocc ReleaseFast native failure
error: the following build command failed with exit code 1:
/home/ehaas/source/arocc/zig-cache/o/a1cd27b47543b4cc16e8062fdc1652f8/build /home/ehaas/.local/zig-linux-x86_64-0.11.0-dev.3947+89396ff02/zig /home/ehaas/source/arocc /home/ehaas/source/arocc/zig-cache /home/ehaas/.cache/zig -Doptimize=ReleaseFast
```

Happens on this branch: https://github.com/ehaas/arocc/tree/coerce and not on master. Does not happen on MacOS with the same zig version, for any build modes. Does not happen on Linux in `Debug` or `ReleaseSafe` modes. I tried clearing the zig cache but that made no difference.

andrewrk:
Are you able to run the command with a debug build of the compiler?

[...]

PR #491, opened 2023-07-10
Improve pointer coercion

ehaas:
Note: With these changes, `ReleaseFast` and `ReleaseSmall` fail to build on Linux, see https://github.com/Vexu/arocc/issues/490

Fixes non-ints/pointers coercing to `void *`

Allows pointers to different child types to coerce to each other with a warning instead of error

Adds a special warning for coercing between pointers where only the signedness of the child type differs

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -12,6 +12,8 @@ const is_windows = @import("builtin").os.tag == .windows;
 
 const Diagnostics = @This();
 
+const PointerSignMessage = " converts between pointers to integer types with different sign";
+
 pub const Message = struct {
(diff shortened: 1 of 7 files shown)


Vexu:
> Note: With these changes, `ReleaseFast` and `ReleaseSmall` fail to build on Linux, see #490

I'm not sure I want to merge this before this gets fixed upstream since it blocks Aro based translate-c.

ehaas:
I'm no longer experiencing the `ReleaseFast` / `ReleaseSmall` compiler crash with this branch (Zig `0.11.0-dev.4271+1aacfa718`)

PR #492, opened 2023-07-14
Improve toolchain integration

ehaas:
Work in progress - Add some additional arguments to the linker + path discovery for file components

Next step will be figuring out where GCC is installed


ehaas:
@Vexu Ok I think this is finally ready for review. There's a lot here, would it be easier if I tried to break it up into a bunch of smaller PRs that build on each other? The logic is mostly adapted from clang; with the caveat that there are probably lots of bugs in the translation because I can't really test it on a wide variety of systems.

src/main.zig
```diff
@@ -3,12 +3,13 @@ const mem = std.mem;
 const Allocator = mem.Allocator;
 const process = std.process;
 const Compilation = @import("Compilation.zig");
-const Driver = @import("Driver.zig");
+const Driver = @import("driver/Driver.zig");
```
Vexu, line 6: I think the proper naming scheme would be for `Driver.zig` to remain in `src` and the other files related to it being in a `Driver` folder.

src/main.zig
```diff
@@ -3,12 +3,13 @@ const mem = std.mem;
...
 const target_util = @import("target.zig");
+const Toolchain = @import("Toolchain.zig");
 
 var general_purpose_allocator = std.heap.GeneralPurposeAllocator(.{}){};
 
-pub fn main() u8 {
+pub fn main() !u8 {
```
Vexu, line 12: I'd prefer main to not return an error union.

Vexu (commented):
> Ok I think this is finally ready for review. There's a lot here, would it be easier if I tried to break it up into a bunch of smaller PRs that build on each other? The logic is mostly adapted from clang; with the caveat that there are probably lots of bugs in the translation because I can't really test it on a wide variety of systems.

Skimming through I didn't spot anything obviously bad, I'm fine with merging it as is and fixing things as they're found unless you want more throughout reviews?

ehaas:
Ok, made the suggested changes. I'm ok with merging it as-is (assuming it works on your machine) and then fixing issues as they come up when more people start using it on a broader variety of systems.

PR #493, opened 2023-07-27
Zig stdib: DeclEnums now require pub decls

ehaas:
As of Zig 3c08fe931a10618950c6af9e89226d1d9b20bbb9 only pub decls are used for DeclEnums


src/Parser.zig
```diff
@@ -4357,7 +4357,7 @@ const CallExpr = union(enum) {
         return switch (self) {
             .standard => true,
             .builtin => |builtin| switch (builtin.tag) {
-                .__builtin_va_start, .__va_start, .va_start => arg_idx != 1,
+                .__builtin_va_start, .__va_start => arg_idx != 1,
```
Vexu, line 4360: What's this change?

ehaas: Oops, thank you for catching that. I initially updated `BuiltinFunction.zig` by regenerating it with the python script, against LLVM16 (I believe it was initially generated with LLVM15). But they changed how they represent `LIBBUILTIN` headers, and the script failed to render those. Before I realized what had happened I updated that switch, but then I figured it out and reverted the `BuiltinFunction.zig` change and just made the decls pub, planning to do the update as a separate commit. But then I forgot to revert the parser change. I've pushed an update that removes that change.

PR #494, opened 2023-07-31
Add environment variable abstraction

ehaas:
The `toolchain` branch uses the `PATH` environment variable, so it might be slightly easier to review if I get this in first.

The motivation is https://github.com/ziglang/zig/issues/4524 - the library code shouldn't read from the environment since it's not guaranteed to be available, plus this makes testing a bit easier.


Vexu (commented):
This seems a bit overcomplicated, are you sure you want to go with this over some fields in Compilation? I can't imagine there being that many places checking for environment variables.

ehaas:
That would work too. Here's the list of env vars GCC uses: https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html

It looks like the following might be relevant for aro:
```
TMPDIR
GCC_EXEC_PREFIX
COMPILER_PATH
LIBRARY_PATH
CPATH
C_INCLUDE_PATH
SOURCE_DATE_EPOCH
```

So maybe just a field called `env` that has a `?[]const u8` for each of those 7 variables, and let the driver/end user manage them and their lifetimes?

PR #496, opened 2023-08-06
Build: make aro module public

marler8997:
Exposes `src/lib.zig` as the `aro` module.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -49,7 +49,7 @@ pub fn build(b: *Build) !void {
     const zig_module = b.createModule(.{
         .source_file = .{ .path = "deps/zig/lib.zig" },
     });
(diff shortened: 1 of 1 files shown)


Vexu:
Not completely sure what this does but I'm guessing it's to do with pure Zig dependencies?

marler8997:
If `arocc` is in your `build.zig.zon`, it let's you access the module from another `build.zig` like this:

```zig
    const arocc_dep = b.dependency("arocc", .{
        .target = target,
        .optimize = optimize,
    });

    // call arocc_dep.module("aro") to get the module from aro's build.zig, i.e.
    exe.addModule("aro", arocc_dep.module("aro"));
```

PR #499, opened 2023-08-30
Assembly statement parsing 

ehaas:
Primary thing here is adding support for GNU inline assembly statements. Currently it just parses the statement, no attempt is made to validate it. There's also the question of how it should be added to the AST - any thought about how to structure that? There's a lot of data - the assembly template string, `volatile` flag, inputs, output, and label expressions, constraints, clobbers...

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -3638,6 +3638,154 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
     }
 }
 
+fn parseMSVCAsmStmt(p: *Parser) Error!?NodeIndex {
+    return p.todo("MSVC assembly statements");
+}
+
+fn parseAsmOperands(p: *Parser, names: *std.ArrayList(?TokenIndex), constraints: *NodeList, exprs: *NodeList) Error!void {
+    if (!p.tok_ids[p.tok_i].isStringLiteral() and p.tok_ids[p.tok_i] != .l_bracket) {
+        // Empty
+        return;
+    }
+    while (true) {
+        if (p.eatToken(.l_bracket)) |l_bracket| {
+            const ident = (try p.eatIdentifier()) orelse {
+                try p.err(.expected_identifier);
+                return error.ParsingFailed;
+            };
+            try names.append(ident);
+            try p.expectClosing(l_bracket, .r_bracket);
+        } else {
+            try names.append(null);
+        }
+        const constraint = try p.asmStr();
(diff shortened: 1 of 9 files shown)


src/Parser.zig
```diff
@@ -3635,50 +3638,219 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
...
+        });
+    }
+
+    const expected_items = 8; // arbitrarily chosen, most assembly will have fewer than 8 inputs/outputs/constraints/names
+    const bytes_needed = expected_items * @sizeOf(?TokenIndex) + expected_items * 3 * @sizeOf(NodeIndex);
+
+    var stack_fallback = std.heap.stackFallback(bytes_needed, p.comp.gpa);
+    const allocator = stack_fallback.get();
+
+    var names = std.ArrayList(?TokenIndex).initCapacity(allocator, expected_items) catch unreachable;
```
ehaas, line 3690: I'm using `null` `TokenIndex` to represent a non-existent name but I think we could also get away with using `0` since a name will never be the 0th token.

src/Parser.zig
```diff
@@ -3635,50 +3638,219 @@ fn convertInitList(p: *Parser, il: InitList, init_ty: Type) Error!NodeIndex {
...
+    defer names.deinit();
+    var constraints = NodeList.initCapacity(allocator, expected_items) catch unreachable;
```
ehaas, line 3692: Also, are these `catch unreachable` safe/worthwhile/a good idea? My reasoning is that they'll all allocate from the stack initially, given the stack space reserved + the number of items, and therefore can't fail.

src/Parser.zig
Vexu: Doesn't matter now while it's a temporary value but it might be good if it ends up being saved in the tree.

src/Parser.zig
Vexu: They're fine but could use a `stack allocation already succeeded` comment.

Vexu (approved):
Looks good, thanks! Could you add comments on the grammar? I always forget how these work.

ehaas:
Added grammar comments, let me know if they should be more detailed - right now it's just the syntax; would it be helpful to explain e.g. that the first asmOperand is inputs, second is outputs, etc.?

Vexu:
The grammar combined with reading the comments in the function is good enough for me and gnu has this documented online if needed.

PR #500, opened 2023-09-03
Improve complex number support

ehaas:
Add support for 2-element initlist for complex numbers + `__builtin_complex`

### DIFF
--- a/test/cases/ast/_Float16.c
+++ b/test/cases/ast/_Float16.c
@@ -27,7 +27,7 @@ fn_def: 'fn (x: int, ...) void'
     var: '[1]struct __va_list_tag'
      name: va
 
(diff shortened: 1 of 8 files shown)


test/cases/ast/_Float16.c
```diff
@@ -27,7 +27,7 @@ fn_def: 'fn (x: int, ...) void'
     var: '[1]struct __va_list_tag'
      name: va
 
-    builtin_call_expr: 'fn (d[1]struct __va_list_tag, ...) void'
+    builtin_call_expr: 'void'
```
ehaas, line 30: This was previously incorrect and showed the type of the builtin itself instead of the type of the return value

PR #503, opened 2023-09-11
Value: Use a start/end range instead of slice for byte values

ehaas:


Vexu:
Not currently but possibly sometime in the future. It can be removed for now.

ehaas:
On this branch with `value` removed I get `max mem used = 10.97MiB` for the struct layout tests vs `11.31MiB` on master

PR #504, opened 2023-09-12
Driver: don't exit with an error code if linker succeeds in non-debug modes

ehaas:


Vexu:
There should still be an early exit, just with 0.

PR #510, opened 2023-09-28
target: add liteos to toLLVMTriple

ehaas:


ehaas:
I'm getting a crash probably unrelated to this change when building the integration test runner (crash in ReleaseFast, seems to be an infinite loop in ReleaseSafe, running on Linux x86_64, zig `0.12.0-dev.627+de4d1ea25`, doesn't happen on MacOS for me). I'll have some time tomorrow to set up LLVM17 and a debug build of the zig compiler.

PR #511, opened 2023-09-28
Don't require utf-8 source files

ehaas:
This removes the requirement that source files always be UTF-8 encoded. Identifiers are still required to be UTF-8 encoded; and there is a todo item for char literals to match clang/gcc behavior when the source text of the literal is not UTF-8 encoded.

Some simple testing on my machine shows that skipping the up-front UTF-8 validation is significantly faster for large files - the 8.4 meg amalgamated sqlite3.c source parses in ~400ms on master; and ~340ms on this branch for me.


ehaas:
@squeek502 when you get a chance can you try out this PR and see if it fixes the encoding issues you were seeing? You'll need a fairly recent build of the Zig compiler. It should also be somewhat faster than before since we're no longer scanning the whole file up front to validate UTF8-ness.

squeek502:
Haven't investigated fully yet, but preliminary results are good. This branch brings the [win32-samples-rc-tests](https://github.com/squeek502/win32-samples-rc-tests) results to:

```
57 .rc files processed with discrepancies
different .res outputs:     0
unexpected compile errors:  57
missing compile errors:     0

428 .rc files processed without discrepancies
identical .res outputs:     403
expected compile errors:    25
```

up from

```
83 .rc files processed with discrepancies
different .res outputs:     0
unexpected compile errors:  83
missing compile errors:     0

402 .rc files processed without discrepancies
identical .res outputs:     377
expected compile errors:    25
```

src/Parser.zig
```diff
@@ -186,55 +186,104 @@ string_ids: struct {
...
+/// Return value indicates whether the token should be considered an identifier
+/// true means consider the token to actually be an identifier
+/// false means it is not
+fn validateExtendedIdentifier(p: *Parser) !bool {
+    assert(p.tok_ids[p.tok_i] == .extended_identifier);
+
+    const slice = p.tokSlice(p.tok_i);
+    const view = std.unicode.Utf8View.init(slice) catch {
+        try p.errTok(.invalid_utf8, p.tok_i);
+        return error.FatalError;
```
Vexu, line 229: This looks like a place for future improvements.

Vexu (approved):
These are some really nice changes!

PR #514, opened 2023-09-30
feat: implement `__has_declspec_attribute`

### LINKED ISSUE #419: Implement __has_declspec_attribute
This should be fairly straightforward to implement - the functionality is almost the same as for `.macro_param_has_attribute` in `Tokenizer.zig` and `Preprocessor.zig`

Additionally, `fromStringDeclspec` in `Attribute.zig` should operate on "normalized" attribute names (with leading and trailing double-underscores removed)

Tests - uncomment lines 11-17 in `test/cases/declspec.c`

bingcicle:
closes #419 

CI seems to be failing, not sure if it's related to [this](https://github.com/Vexu/arocc/pull/510#issuecomment-1738617220)?

### DIFF
--- a/src/target.zig
+++ b/src/target.zig
@@ -684,6 +684,7 @@ pub fn toLLVMTriple(target: std.Target, buf: []u8) []const u8 {
         .watchos => "watchos",
         .driverkit => "driverkit",
         .shadermodel => "shadermodel",
+        .liteos => "liteos",
         .opencl,
         .glsl450,
         .vulkan,
--- a/src/Attribute.zig
+++ b/src/Attribute.zig
@@ -977,10 +977,11 @@ fn fromStringC2X(namespace: ?[]const u8, name: []const u8) ?Tag {
 }
 
 fn fromStringDeclspec(name: []const u8) ?Tag {
+    const normalized = normalize(name);
     const decls = @typeInfo(attributes).Struct.decls;
     inline for (decls, 0..) |decl, i| {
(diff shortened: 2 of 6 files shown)


src/target.zig
```diff
@@ -684,6 +684,7 @@ pub fn toLLVMTriple(target: std.Target, buf: []u8) []const u8 {
         .watchos => "watchos",
         .driverkit => "driverkit",
         .shadermodel => "shadermodel",
+        .liteos => "liteos",
```
bingcicle, line 687: I added this in order to make it compile, but https://github.com/Vexu/arocc/pull/510 fixes this.

Vexu (changes requested):
`__has_declspec_attribute(<anything>)` should return `0` when `langopts.declspec_attrs == false`.

CI failure is definitely because of llvm-17, I'll try to figure it out.

bingcicle:
@Vexu Hmm, am I misundestanding the usage of

```
//aro-args -fdeclspec
```

at the top of the file? I was under the impression that including the args switches on the options in tests

Vexu:
No you understand correctly, I was talking about the case where `-fdeclspec` was not specified.

bingcicle:
Ah should I add a check there as well? I'm assuming you mean `no_declspec.c`

Vexu:
Something like that yes.

Vexu:
Thanks!

PR #519, opened 2023-10-05
add C23 error diagnostic: missing type specifier

### LINKED ISSUE #420: Implicit int diagnostics
Starting with C23, implicit-int type specifiers are no longer allowed:
```C
// test.c
static x = 5;
```

```
➜  ~/local/llvm15-release/bin/clang -std=c2x test.c
test.c:1:8: error: a type specifier is required for all declarations
static x = 5;
~~~~~~ ^
1 error generated.
```

This will need a new error diagnostic in `Diagnostics.zig` and an update to `Type.zig` where we issue `.missing_type_specifier`

danielsan901998:
resolve #420

### DIFF
--- a/src/Diagnostics.zig
+++ b/src/Diagnostics.zig
@@ -374,6 +374,10 @@ const messages = struct {
         const kind = .warning;
         const all = true;
     };
+    pub const missing_type_specifier_c2x = struct {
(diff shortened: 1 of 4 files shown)


Vexu:
Thanks!

PR #520, opened 2023-10-09
Compilation: read entire source into a buffer before processing it

ehaas:
Instead of reading one byte at a time via a reader, just read the whole thing into a []u8 and operate on it from there.

Not sure why I did it one byte at a time via a reader originally but this way is faster:

parsing sqlite3.c, master branch
```
real	0m0.358s
user	0m0.342s
```

parsing sqlite3.c, this branch,
```
real	0m0.333s
user	0m0.304s
```

Tests in ReleaseFast, master branch
```
real	0m3.036s
user	0m23.240s
```

Tests in ReleaseFast, this branch:
```
real	0m2.733s
user	0m20.778s
```

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1084,10 +1084,8 @@ pub fn addSourceFromOwnedBuffer(comp: *Compilation, buf: []u8, path: []const u8,
     const splice_locs = try splice_list.toOwnedSlice();
     errdefer comp.gpa.free(splice_locs);
 
-    // Important: do not perform any possibly-failing operations in this function after this realloc.
-    // Otherwise errdefers in callers will possibly free the realloced slice using the original len
-    // instead of the new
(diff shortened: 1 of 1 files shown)


src/Compilation.zig
```diff
@@ -1081,6 +1084,9 @@ pub fn addSourceFromReader(comp: *Compilation, reader: anytype, path: []const u8
     const splice_locs = try splice_list.toOwnedSlice();
     errdefer comp.gpa.free(splice_locs);
 
+    // Important: do not perform any possibly-failing operations in this function after this realloc.
+    // Otherwise errdefers in callers will possibly free the realloced slice using the original len
+    // instead of the new
```
andrewrk, line 1089: Check this out, you can put this on the next line:
```zig
errdefer @compileError("errdefers in callers would possibly free the realloced slice using the original len");
```

ehaas: oh wow, that's awesome!

PR #525, opened 2023-10-16
Add a build step to generate dafsa from a def file

Vexu:
This mostly works except for an ever so pesky "file exists in multiple modules" error.

Also give it up for the first DSL in Aro!


ehaas:
This fixes the extra space in the output for me:

```
diff --git a/src/Tree.zig b/src/Tree.zig
index 91d1035..c3bc67c 100644
--- a/src/Tree.zig
+++ b/src/Tree.zig
@@ -703,7 +703,7 @@ fn dumpAttribute(attr: Attribute, strings: []const u8, writer: anytype) !void {
     switch (attr.tag) {
         inline else => |tag| {
             const args = @field(attr.args, @tagName(tag));
-            if (@TypeOf(args) == void) {
+            if (@typeInfo(@TypeOf(args)).Struct.fields.len == 0) {
                 try writer.writeByte('\n');
                 return;
             }
```

PR #527, opened 2023-10-20
Add wide & unicode string literal support

ehaas:
Adds support for `L""` / `u""` / `U""` literals.

String contents are written to `p.retained_strings`

Also fixes 2 unrelated bugs:

1) Max diagnostics for a char or string literal is currently capped at 4; we drop any after that. So if 4 warnings happened first, followed by an error, the error would get dropped. New behavior is to replace the last warning if adding an error fails.
2) Fix a crash when passing a non-string to an attribute that expects a string (introduced in #503)


Vexu (approved):
Thanks for working on these! The `invalid` token is now almost exclusively used for incorrectly encoded bytes.

Would you mind taking a look at the remaining cases? I think some of them can simply be marked unreachable thanks to line splicing and the others could be changed to `empty_char_literal` and `unterminated_{char,string}_literal` with better error messages.

The split between `CharKind` and `StringKind` with nearly identical APIs seems a bit odd to me but if you think it's better then it's fine.

ehaas:
I think the only reason I made them separate is to prevent string and char literals from concatenating but you're right, they can probably be combined. I'll take a look at that and the `invalid` token.

PR #528, opened 2023-10-24
Type: improve type equality checking

ehaas:
I tried to figure these out by reading the standard but it's pretty dry. 

Functions in C11 6.7.6.3.15 (too long to paste)

Arrays in C11 6.7.6.2.6: "For two array types to be compatible, both shall have compatible element types, and if both size specifiers are present, and are integer constant expressions, then both size specifiers shall have the same constant value"

Briefly,

1) Arrays with same child type are compatible if at least one array is incomplete or if they have the same constant integer length
2) Functions must both have or both not have ellipses to be compatible
3) A function with no prototype is compatible with a function with a parameter list if none of the named parameter types undergo default argument promotion and the return type is compatible

### DIFF
--- a/test/cases/tentative array.c
+++ b/test/cases/tentative array.c
@@ -9,5 +9,4 @@ int arr[3]; // TODO should make the error go away
 #define EXPECTED_ERRORS \
     "tentative array.c:1:5: warning: tentative array definition assumed to have one element" \
     "tentative array.c:3:16: error: invalid application of 'sizeof' to an incomplete type 'int []'" \
-    "tentative array.c:5:5: error: redefinition of 'arr' with a different type" \
-    "tentative array.c:1:5: note: previous definition is here" \
+
--- a/test/cases/tentative array.c
+++ b/test/cases/tentative array.c
@@ -2,7 +2,7 @@ int arr[];
 
 _Static_assert(sizeof arr,"");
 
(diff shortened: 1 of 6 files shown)


test/cases/tentative array.c
```diff
@@ -9,5 +9,4 @@ int arr[3]; // TODO should make the error go away
 #define EXPECTED_ERRORS \
     "tentative array.c:1:5: warning: tentative array definition assumed to have one element" \
     "tentative array.c:3:16: error: invalid application of 'sizeof' to an incomplete type 'int []'" \
-    "tentative array.c:5:5: error: redefinition of 'arr' with a different type" \
-    "tentative array.c:1:5: note: previous definition is here" \
+
```
Vexu, line 12: Should also remove the `TODO` and `SKIPPED_TESTS`.

Vexu (approved):
This looks like it resolves my `TODO adjusted equality check` comments in `SymbolStack.zig`.

test/cases/tentative array.c
ehaas: For this one the `warning: tentative array definition assumed to have one element` should not happen since the definition is later completed. I think we'd have to do something similar to what we do for `tentative_defs` in the Parser since we don't want to issue the warning until we're done and have determined the definition is incomplete.

PR #532, opened 2023-11-01
Zig workaround: don't use packed structs larger than 128 bits.

ehaas:
The C backend currently can't handle them. This is the only thing preventing the CBE from being able to produce valid C code from aro (I have not verified that the produced code actually works, however). 

If the produced code works as expected, this will unblock https://github.com/ziglang/zig/pull/17771

Does `Relocation` even need to be packed? It looks like we aren't bitcasting it, and it's 32 bytes regardless of whether it's packed or not (the bitsize is 184 when packed vs 256 when not packed, but if we're only using it in `ArrayListUnmanaged` then that shouldn't matter, right?)

`Options` goes from 64 bytes -> 104 bytes when not packed, but we only have 1 of those per Compilation.

### DIFF
--- a/deps/zig/codegen/x86_64.zig
+++ b/deps/zig/codegen/x86_64.zig
@@ -176,19 +176,21 @@ pub const Encoder = struct {
 
     /// Directly write a number to the code array with big endianness
     pub fn writeIntBig(self: Self, comptime T: type, value: T) void {
-        mem.writeIntBig(
+        mem.writeInt(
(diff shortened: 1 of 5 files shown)


Vexu:
I removed the TODO comment since the options struct was originally created using the original packed struct semantics which didn't involve bit manipulation of 512 bit integers so the tiny memory increase should be negated by smaller code size.

PR #533, opened 2023-11-02
Parser: set kind to param in old style parameter declarations

### LINKED ISSUE #351: `static` keyword not allowed in array parameters for K&R style function definition
The following is not accepted but should be:

```C
void foo(x)
    int x[static 5];
{

}
```

```
./test.c:2:10: error: 'static' used outside of function parameters
    int x[static 5];
         ^
```

danielsan901998:
resolve #351
When parsing old style parameter declarations, declarator kind was not initialized as param but as normal, that generate an error for K&R style function definition when using static.
Added a test in cases/functions.c


Vexu:
Thanks!

PR #535, opened 2023-11-06
Implement more C23 changes

Vexu:


src/Diagnostics.zig
```diff
@@ -2546,6 +2547,21 @@ const messages = struct {
         const msg = "unterminated comment";
         const kind = .@"error";
     };
+    pub const def_no_proto_deprecated = struct {
+        const msg = "a function definition without a prototype is deprecated in all versions of C and is not supported in C2x";
+        const kind = .warning;
+        const opt = "deprecated-non-prototype";
+    };
+    pub const passing_args_to_kr = struct {
+        const msg = "passing arguments to without a prototype is deprecated in all versions of C and is not supported in C2x";
```
ehaas, line 2556: Looks like a small typo in the error message - `s/to without/to a function without/`

Vexu: Looks like I copied Clang's bugs too. Thanks, fixed.

PR #540, opened 2023-11-08
CharInfo: implement C23 rules for identifiers

ehaas:
This is the first part of the c23 unicode update; the second part will be adding identifier normalization.


Vexu:
> the second part will be adding identifier normalization.

There might already be a Zig library for this that we could use.

ehaas:
It looks like maybe we can get away with just rejecting non-normalized identifiers? 


https://open-std.org/JTC1/SC22/WG14/www/docs/n2836.pdf
> Therefore, normalization of identifiers must happen before object files are emitted. It is either the
responsibility of programmers to provide normalized identifiers, and compilers to either warn or error
on non-normalized ones, or for compilers to normalize Unicode input. For non-Unicode input, the
transformation to normalized form is the simple table based lookup used to translate now.

This project provides code that handles the normalization form we'll need https://codeberg.org/dude_the_builder/ziglyph

Vexu:
> It looks like maybe we can get away with just rejecting non-normalized identifiers

That would be nice. We should definitely detect non-normalized identifiers since there is a simple check for it and the tables for `NFC_Quick_Check={No,Maybe}` are quite small.

Issue #240, opened 2022-02-07
rework Aro type and value system to use this InternPool data structure


andrewrk:
I was thinking about sending a patch to Aro to test pilot this idea:

https://github.com/ziglang/zig/blob/21135387fb7c2dbaf70a72f2c97341e9c1307045/src/InternArena.zig

I have a hunch that it would provide the following improvements:
 * more convenient for Aro development
 * more convenient for Aro API user (zig)
 * improved performance, *especially* for a C compiler
 * reduced memory usage

Would such a patch be welcome, or would you recommend against it?

Vexu:
The type system especially is in a serious need of being redesigned but while using InternPool would give performance and memory usage improvements, it wouldn't directly solve the biggest pain point in the current system which is the complexity brought by `Type.attributed` and `Type.typeof`.

ehaas:
In case we do end up using this I started adding features on a branch https://github.com/ehaas/arocc/blob/interned-types/src/NewType.zig

I added variable-length encoding for `u64` and an encoding for slices (e.g. function parameters). Using this interned type would also make it easy to do something like clang does where the const/volatile/restrict qualifiers are stored in the top 3 bits (also one bit is reserved for "this has extended qualifiers, look for the data elsewhere", but I haven't checked to see if those are ObjC / C++ specific). That means we can only have about 2**28 unique unqualified types but that should be enough for anyone.

Would still need to figure out Attribute storage, storage for strings (if we include them in the type e.g. record field names), and of course port everything to the new system.

andrewrk:
Does C let you put const/volatile/restrict on non pointer types? If not, it would be possible to fold these flags into the enum tag, taking up 8 enum tags for pointer rather than 1.

ehaas:
I think it might be needed for `typeof` types e.g.

```c
typeof(const int) *foo = 0;
```

PR #541, opened 2023-11-08
Intern Values

Vexu:
This implements the `Value` part of #240. It'll probably make sense to rework `Type` some other way.


ehaas:
I'm getting a crash on division (linux x86_64):
```c
void foo(void) {
   int x = 128/8;
}
```

```
thread 424146 panic: index out of bounds: index 20, len 2
/home/ehaas/.local/zig-linux-x86_64-0.12.0-dev.1396+f6de3ec96/lib/std/math/big/int.zig:145:32: 0x4aecd4 in toConst (arocc)
            .limbs = self.limbs[0..self.len],
                               ^
/home/ehaas/source/arocc/src/Value.zig:484:79: 0x6dec03 in div (arocc)
        res.* = try intern(comp, .{ .int = .{ .big_int = result_bigint.toConst() } });
                                                                              ^
/home/ehaas/source/arocc/src/Parser.zig:6294:36: 0x6dce31 in mulExpr (arocc)
                if (try lhs.val.div(lhs.val, rhs.val, lhs.ty, p.comp)) try p.errOverflow(mul.?, lhs);
                                   ^
/home/ehaas/source/arocc/src/Parser.zig:6241:28: 0x6da907 in addExpr (arocc)
    var lhs = try p.mulExpr();
                           ^
/home/ehaas/source/arocc/src/Parser.zig:6218:28: 0x6d8687 in shiftExpr (arocc)
    var lhs = try p.addExpr();
                           ^
/home/ehaas/source/arocc/src/Parser.zig:6189:30: 0x6d71a7 in compExpr (arocc)
    var lhs = try p.shiftExpr();
                             ^
/home/ehaas/source/arocc/src/Parser.zig:6168:29: 0x6d4b47 in eqExpr (arocc)
    var lhs = try p.compExpr();
                            ^
```

src/Value.zig
```diff
@@ -691,6 +691,12 @@ pub fn shr(lhs: Value, rhs: Value, ty: Type, ctx: Context) !Value {
 }
 
 pub fn compare(lhs: Value, op: std.math.CompareOperator, rhs: Value, ctx: Context) bool {
+    if (op == .eq) {
+        return lhs.opt_ref == rhs.opt_ref;
```
ehaas, line 668: I think we'll also need to check for "runtime" 1 & 0 vs "comptime" (statically known). e.g. the following will fail:

`_Static_assert(sizeof(char) == 1, "");` since the `Value.one` isn't the same as the computed value of `sizeof(char)`

Vexu: > since the `Value.one` isn't the same as the computed value of `sizeof(char)`

It is for me locally after I fixed a bug.

Vexu:
All tests pass locally, CI is waiting for https://github.com/ziglang/zig/commit/e72049bc61eedd2b7381b87b530428a83581cafe

PR #545, opened 2023-11-09
Parser: Coerce array members properly from strings

### LINKED ISSUE #537: Compound literals do not initialize array members properly from strings
Looks like it tries to initialize the first member with a pointer and uses a filler for the rest.

```c
struct S {
    char bytes[32];
};

void foo(void) {
    struct S s = (struct S){"ABC"};
}
```

```
fn_def: 'fn () void'
 name: foo
 body:
  compound_stmt_two: 'void'
    var: 'struct S'
     name: s
     init:
      implicit_cast: (lval_to_rval) 'struct S'
        compound_literal_expr: 'struct S' lvalue
         struct_init_expr_two: 'struct S'
           array_init_expr_two: '[32]char'
             implicit_cast: (pointer_to_int) 'char'
               implicit_cast: (array_to_pointer) 'd[4]char'
                 string_literal_expr: '[4]char' lvalue (value: "ABC")
             array_filler_expr: 'char'
              count: 31
    implicit_return: 'void'

./test.c:7:29: warning: implicit pointer to integer conversion from 'char *' to 'char' [-Wint-conversion]
    struct S s = (struct S){"ABC"};
```

danielsan901998:
resolve #537
coerceArrayInit was not used in findScalarInitializer, using only a type equality, so instead of coercing the string it only checked if the type was equal, and because the length was different the array was rejected and istead used implicit conversion to char from pointer.

### DIFF
--- a/src/Parser.zig
+++ b/src/Parser.zig
@@ -3356,8 +3356,8 @@ fn initializerItem(p: *Parser, il: *InitList, init_ty: Type) Error!bool {
 
             excess: {
                 if (index_hint) |*hint| {
-                    if (try p.findScalarInitializerAt(&cur_il, &cur_ty, res.ty, first_tok, hint)) break :excess;
-                } else if (try p.findScalarInitializer(&cur_il, &cur_ty, res.ty, first_tok)) break :excess;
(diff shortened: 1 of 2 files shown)


Vexu:
Wonderful, Thanks!

PR #550, opened 2023-11-10
Parser: check that identifiers are in NFC

Vexu:


ehaas:
It looks like this issues an error for the "maybe" QuickCheck classification? I think that's fine - I have a local branch that does the actual NFC normalization for maybe's (based on https://codeberg.org/dude_the_builder/ziglyph) but it involves pulling in a surprising amount of code and unicode data files. We could always add it later as an enhancement if we want.

src/aro/Parser.zig
```diff
@@ -260,6 +262,26 @@ fn validateExtendedIdentifier(p: *Parser) !bool {
         if (!warned) {
             warned = try checkIdentifierCodepointWarnings(p.comp, codepoint, loc);
         }
+
+        // Check NFC normalization.
+        if (!normalized) continue;
+        if (codepoint >= 0x10000 and codepoint <= 0x10FFFF) continue;
```
ehaas, line 268: Is this right? I think I made the same mistake - if you adapted this from the example in the unicode docs, it's written for Java which is UTF-16:

```
    int ch = source.codepointAt(i);
        if (Character.isSupplementaryCodePoint(ch)) ++i;
   ...
```

So it's not skipping over supplementary codepoints; it's just skipping the low surrogate of the UTF-16 encoding.

Vexu: Oh right, thanks!

Vexu:
> It looks like this issues an error for the "maybe" QuickCheck classification?

That's what GCC does by default.

ehaas:
It's not done yet but I have most of it here: https://github.com/ehaas/arocc/tree/unicode-normalize

I rebased against latest master so there will be some duplication of char classification code.

Vexu:
That's definitely a lot (and also requires all of this PR?); clang doesn't even check for normalization. I'd also assume this to be rarely needed due to most inputs providing correctly NFC data.

Do you have any opinions on how this should behave?

ehaas:
I'm definitely on board with the less-involved solution. Seems a bit silly for a C compiler to need a relatively deep understanding of unicode IMO.

Vexu:
> Seems a bit silly for a C compiler to need a relatively deep understanding of unicode IMO.

Didn't stop the GNU people.

But I agree, I think this check is a good middle ground.

PR #556, opened 2023-11-12
Escape filename path in line markers

### LINKED ISSUE #553: Backslashes should be escaped in `#line` directive filenames
All of the below uses the following as `foo.c` (doesn't really matter what's in this file):
```c
foo
```

Current behavior:
```
> arocc -E .\foo.c
# 1 ".\foo.c" 1
# 1 "<builtin>" 1
# 186 "<builtin>"
# 1 "<command line>" 1
# 1 ".\foo.c" 2
foo
```

Expected behavior:

```
> arocc -E .\foo.c
# 1 ".\\foo.c" 1
# 1 "<builtin>" 1
# 186 "<builtin>"
# 1 "<command line>" 1
# 1 ".\\foo.c" 2
foo
```

The expected behavior matches clang:

```
> clang -E -xc .\foo.c
# 1 ".\\foo.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 368 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 ".\\foo.c" 2
foo
```

and `rc.exe`'s preprocessor, too (so it should always escape `\` even with `--emulate=msvc`:

```
> rc.exe /p .\foo.c
#line 1 "C:\\Users\\Ryan\\Programming\\Zig\\resinator\\tmp\\RCa03024"
#line 1 ".\\foo.c"
#line 1 ".\\foo.c"
foo
```

---

Unsure about other characters that should be escaped, might want to do something similar to `std.zig.fmtEscapes`

squeek502:
Closes #553. See https://github.com/Vexu/arocc/issues/553#issuecomment-1807069502 for some relevant details.

@Vexu unsure what you meant by inlining the loop btw, adding the `inline` keyword before the `for` results in

```
src\aro\Preprocessor.zig:3060:17: error: unable to resolve comptime value
    inline for (bytes) |byte| switch (byte) {
                ^~~~~
```

Let me know if I was misinterpreting that.

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -3059,7 +3059,23 @@ fn printLinemarker(
     try w.writeByte('#');
     if (pp.linemarkers == .line_directives) try w.writeAll("line");
     // line_no is 0 indexed
-    try w.print(" {d} \"{s}\"", .{ line_no + 1, source.path });
+    try w.print(" {d} \"", .{line_no + 1});
+    for (source.path) |byte| switch (byte) {
+        '\n' => try w.writeAll("\\n"),
+        '\r' => try w.writeAll("\\r"),
+        '\t' => try w.writeAll("\\t"),
+        '\\' => try w.writeAll("\\\\"),
+        '"' => try w.writeAll("\\\""),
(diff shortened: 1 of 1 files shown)


Vexu:
I meant it as in moving the for loop into the print function.

PR #559, opened 2023-11-14
Update fuzz step and instructions and fix a bug it caught

ehaas:
I was having trouble getting the old fuzzing instructions to work (having to do with opaque pointers in llvm; by default AFL uses clang 14 which defaults to typed pointers but I didn't investigate too deeply).

The new method just compiles aro to C and then runs that through `afl-clang-lto`. It seems to work pretty well - I'm just not sure if I'm handling `compiler-rt` and the `zig.h` location in the best way.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const Build = std.Build;
 const GenerateDef = @import("build/GenerateDef.zig");
+const ZigLibDirStep = @import("build/ZigLibDir.zig");
 
 const aro_version = std.SemanticVersion{
     .major = 0,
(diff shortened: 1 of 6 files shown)


build.zig
```diff
@@ -1,39 +1,59 @@
...
     .patch = 0,
 };
 
-fn addFuzzStep(b: *Build, target: std.zig.CrossTarget) !void {
+fn addFuzzStep(b: *Build, target: std.zig.CrossTarget, afl_clang_lto_path: []const u8, aro_module: *std.Build.Module) !void {
+    const fuzz_step = b.step("fuzz", "Build executable for fuzz testing.");
+    var fuzz_target = target;
+    fuzz_target.ofmt = .c;
+
+    const zig_lib_dir = try introspect.findZigLibDir(b.allocator, b.zig_exe);
```
Vexu, line 17: I think you should be able to use `b.zig_lib_dir` here.

src/aro/Parser.zig
```diff
@@ -323,7 +323,7 @@ fn expectIdentifier(p: *Parser) Error!TokenIndex {
         return p.errExpectedToken(.identifier, actual);
     }
 
-    return (try p.eatIdentifier()) orelse unreachable;
+    return (try p.eatIdentifier()) orelse error.ParsingFailed;
```
Vexu, line 326: Removing the `catch unreachables` makes sense but the reason this can't use `.?` is because `eatIdentifier` returns null when an extended identifier contains errors. Shouldn't it return `error.ParsingFailed` instead or just the token with errors? Seems like returning null leads to redundant diagnostics in many places.

build.zig
ehaas: Looks like it's null by default - the other option would be to require the user to pass `-Dzig-lib-dir=<whatever>` on the command line

Vexu: Didn't notice that, the path is also available via `zig env | jq .lib_dir` (but implemented in build.zig of course).

ehaas: Updated it to lazily read from `zig env` if a fuzz build is requested.

Vexu:
I wasn't sure about the changes to `validateExtendedIdentifier` so I dropped it for now.

PR #567, opened 2023-11-14
Parser: Coerce array from designated initializers

### LINKED ISSUE #565: Compound literals with designated initializers do not initialize arrays properly
Similar to #537 except it happens with designated initializers instead of positional ones. @danielsan901998 maybe a similar fix?

```c
struct S {
    char payload[32];
};
void foo(void) {
    struct S s = (struct S){.payload = "abc"};
}
```

```
test.c:5:40: warning: implicit pointer to integer conversion from 'char *' to 'char' [-Wint-conversion]
    struct S s = (struct S){.payload = "abc"};
                                       ^
```

danielsan901998:
resolve #565
The previous fix of using coerceArrayInit in #545  was wrong, since it generate an error when the array coercion fail instead of simply indicating failing to coerce the initialization, and since up the stack the function called coerceArrayInit again, it could potentially repeat the same error message.
So instead i created a new function called canCoerceArrayInit that do the same logic as coerceArrayInit, but it only return true when coercion is valid and does not generate errors, that allow in the array and complex case to continue processing the initializer when the array coercion fail.

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -3226,7 +3226,7 @@ fn initializer(p: *Parser, init_ty: Type) Error!Result {
         const tok = p.tok_i;
         var res = try p.assignExpr();
         try res.expect(p);
-        if (try p.canCoerceArrayInit(&res, tok, init_ty)) return res;
+        if (try p.coerceArrayInit(&res, tok, init_ty)) return res;
         try p.coerceInit(&res, tok, init_ty);
         return res;
     }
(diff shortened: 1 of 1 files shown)


Vexu (commented):
Nice work!

>So instead i created a new function called canCoerceArrayInit that do the same logic as coerceArrayInit, but it only return true when coercion is valid and does not generate errors,

Could you instead rename `canCoerceArrayInit` to `canCoerceArrayInitExtra` and add a `report_err: bool` parameter and then add a new `canCoerceArrayInit` for backwards compatibility?

Duplicating logic like this easily leads to new bugs.

PR #568, opened 2023-11-15
Type: use a separate flag for decayed arrays instead of encoding it in the specifier

ehaas:
It's not pretty but I view this as a stopgap measure until the Type system is rewritten. My main motivation is that it's currently not possible to use aligned arrays, which I suspect are fairly common in large real-world projects.


Vexu:
Can you put it in the qualifiers like `register`? Though maybe wait for me to write down an idea I had for improving types first.

ehaas:
It could go in quals, just wasn't sure if it belonged there. It seems like either way `Type` is 16 bytes in fast modes and 24 bytes in safe modes. No rush so I can definitely wait to hear your idea for improving types. My short-term goal is to being able to parse `zig2.c` but that's just because it's the biggest standalone C program I can find :)

Issue #546, opened 2023-11-09
Incorrect parse of function returning pointer to array


ehaas:
```c
int ARRAY[2];
int (*foo(void))[2] {
    return &ARRAY;
}
```

```
./test.c:2:21: error: expected ';', found '{'
int (*foo(void))[2] {
                    ^
1 error generated.
```

ehaas:
Introduced in 1cdf4d5380d72cd71d2a4c8dd552cd7eae4018c3

Issue #380, opened 2022-08-14
Preprocessor: expansion disabled when it shouldn't be


ehaas:
```c
#define foo(X) 1 bar
#define bar(X) 2 foo

foo(X)(Y)(Z)
```

gcc and clang expand this to:
`1 2 1 bar`

aro expands it to:
`1 2 foo(Z)`

PR #570, opened 2023-11-17
Builtins: Add frontend support for some builtin atomic operations

ehaas:
This adds all the builtin atomic functions to `Builtin.def` as well as some of the custom typechecking for some of the functions (specifically argument counts and return values; it doesn't implement checking that the types of the parameters are compatible).

With this change + once #546 is fixed, the aro frontend can parse a preprocessed version of `zig2.c` (134 megs) with no obvious errors, just some spurious warnings about functions not returning values.  Successful preprocessing I believe is blocked on #380 (and potentially other issues that may come up after that is resolved).


Vexu (approved):
Can you give a quick summary of what else is needed for atomics support?

ehaas:
1. Defining the `int_fastN_t` / `int_leastN_t` in `stdint.h` so we can define the rest of the atomic typedefs
2. A minimal frontend implementation can implement most of stdatomic.h with just macros e.g. https://raw.githubusercontent.com/rui314/chibicc/main/include/stdatomic.h - for the operations with an explicit memory order this just hardcodes it to always use `memory_order_seq_cst` which is the strongest one (because that's the ordering used by compound assignment to atomic types) . The `atomic_fetch_op` implementations there aren't actually correct since they return the new value but they can be fixed with `typeof`
3. backend support for atomics :)

PR #571, opened 2023-11-18
Add more int types

ehaas:
Adds fastN and leastN int types, and adds the rest of the atomic types to stdatomic.h


src/aro/Driver.zig
```diff
@@ -277,6 +278,8 @@ pub fn parseArgs(
                 d.comp.langopts.declspec_attrs = true;
             } else if (mem.eql(u8, arg, "-fno-declspec")) {
                 d.comp.langopts.declspec_attrs = false;
+            } else if (mem.eql(u8, arg, "-ffreestanding")) {
+                d.comp.langopts.hosted = false;
```
Vexu, line 282: This duplicates information in `comp.target.os.tag`, I think it would be better to move the `hosted` into the arg parsing function, add `-fhosted` and then set `os.tag` to freestanding at the end if `hosted == false`.

ehaas: Updated - how do you think we should handle `-fhosted` + a freestanding `--target`? Right now I've set it to issue a fatal error. Does clang have a command-line notion of a freestanding target?

Vexu: LLVM triples use `unknown` OS for freestanding targets, I think a fatal error or a warning is a good way to go.

PR #579, opened 2023-11-23
Use BoundedArray less

Vexu:


src/aro/Toolchain.zig
```diff
@@ -183,18 +183,15 @@ pub fn getLinkerPath(tc: *const Toolchain, buf: []u8) ![]const u8 {
...
 
-const TargetSpecificToolName = std.BoundedArray(u8, 64);
-
 /// If an explicit target is provided, also check the prefixed tool-specific name
 /// TODO: this isn't exactly right since our target names don't necessarily match up
 /// with GCC's.
 /// For example the Zig target `arm-freestanding-eabi` would need the `arm-none-eabi` tools
-fn possibleProgramNames(raw_triple: ?[]const u8, name: []const u8, target_specific: *TargetSpecificToolName) std.BoundedArray([]const u8, 2) {
+fn possibleProgramNames(raw_triple: ?[]const u8, name: []const u8, buf: *[64]u8) std.BoundedArray([]const u8, 2) {
     var possible_names: std.BoundedArray([]const u8, 2) = .{};
```
ehaas, line 191: We could probably get rid of this usage too without too much effort.  I don't mind doing it but I won't be able to build until the download page updates to latest. Once I do should I just push that change to this branch?

Vexu: I think that is a reasonable place to use `BoundedArray` so no need to remove it unless you want to.

> Once I do should I just push that change to this branch?

Yes, feel free to push to any of my branches when collaborating on a PR.

PR #583, opened 2023-11-26
update README with zig run

bingcicle:
run step is no longer in `build.zig`, so the README example no longer works. We just use `zig run` instead.


Vexu (changes requested):
`zig run` uses clang, this should use `zig-out/bin/arocc`.

bingcicle:
right, oops

I'll just update the instruction to do a `zig build` and running `arocc` after instead

PR #586, opened 2023-11-27
Compilation: check for embedded null bytes in path for has_include

ehaas:
Prevents a crash caused by embedded NUL bytes

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -1349,6 +1349,10 @@ pub fn hasInclude(
     /// __has_include vs __has_include_next
     which: WhichInclude,
 ) !bool {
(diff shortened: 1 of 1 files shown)


ehaas:
This makes it think that `__has_include.c` is a binary file - I can move it into its own file or just leave the test out if you want.

Vexu:
Should this also be done for `__has_embed`?

> This makes it think that `__has_include.c` is a binary file

You can change that with `.gitattributes`.

ehaas:
`__has_embed` uses a different path, I wasn't able to get it to crash on latest master.

Vexu:
Hmm, locally git shows a regular diff but GitHub wont. This can probably go without a test...

PR #589, opened 2023-11-28
Intern types

Vexu:
Some design ideas for improving working with types:
* `Parser.Result` can store both a `Type` and `Type.Interner.Key`
* `get` by default recurses wrapper types
* move named types out of `Compilation`
* ...more?

### DIFF
--- a/src/aro/Type.zig
+++ b/src/aro/Type.zig
@@ -83,16 +83,6 @@ pub const Signedness = enum {
     signed,
     unsigned,
     unspecified,
-    platform_dependant,
-
-    pub fn unwrap(s: Signedness, comp: *const Compilation) std.builtin.Signedness {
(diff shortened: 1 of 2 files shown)


Vexu:
On further thought using an interner here doesn't make much sense (can't do fast equality check; only identity, simple types are as big as the `Ref`, no plans of incremental compilation). I'm going to instead look into using tagged pointers which also enable variable size and different storage/unwrapped representation.

ehaas:
I'm not very familiar with the clang codebase but I was curious how they handle types and it sounds like they might do something similar (with tagged pointers) https://github.com/llvm/llvm-project/blob/ce731770f869e8e61afa2dccb772e8367105a1a6/clang/include/clang/AST/Type.h#L740

PR #590, opened 2023-11-29
fix: spurious unsigned add and mul overflows

bingcicle:
Add signedness checks prior to doing an `errOverflow`.

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -6443,9 +6443,12 @@ fn addExpr(p: *Parser) Error!Result {
         try rhs.expect(p);
 
         const lhs_ty = lhs.ty;
+        const rhs_ty = rhs.ty;
         if (try lhs.adjustTypes(minus.?, &rhs, p, if (plus != null) .add else .sub)) {
             if (plus != null) {
-                if (try lhs.val.add(lhs.val, rhs.val, lhs.ty, p.comp)) try p.errOverflow(plus.?, lhs);
+                if (try lhs.val.add(lhs.val, rhs.val, lhs.ty, p.comp) and
+                    rhs_ty.signedness(p.comp) != .unsigned and
+                    lhs_ty.signedness(p.comp) != .unsigned) try p.errOverflow(plus.?, lhs);
             } else {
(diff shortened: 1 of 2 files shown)


src/aro/Parser.zig
```diff
@@ -6443,9 +6443,12 @@ fn addExpr(p: *Parser) Error!Result {
         try rhs.expect(p);
 
         const lhs_ty = lhs.ty;
+        const rhs_ty = rhs.ty;
         if (try lhs.adjustTypes(minus.?, &rhs, p, if (plus != null) .add else .sub)) {
             if (plus != null) {
-                if (try lhs.val.add(lhs.val, rhs.val, lhs.ty, p.comp)) try p.errOverflow(plus.?, lhs);
+                if (try lhs.val.add(lhs.val, rhs.val, lhs.ty, p.comp) and
+                    rhs_ty.signedness(p.comp) != .unsigned and
+                    lhs_ty.signedness(p.comp) != .unsigned) try p.errOverflow(plus.?, lhs);
```
Vexu, line 6451: `lhs.ty` is the result type of the expression after `adjustTypes` so only it needs to be checked.

src/aro/Parser.zig
```diff
@@ -6484,7 +6487,9 @@ fn mulExpr(p: *Parser) Error!Result {
 
         if (try lhs.adjustTypes(percent.?, &rhs, p, if (tag == .mod_expr) .integer else .arithmetic)) {
             if (mul != null) {
-                if (try lhs.val.mul(lhs.val, rhs.val, lhs.ty, p.comp)) try p.errOverflow(mul.?, lhs);
+                if (try lhs.val.mul(lhs.val, rhs.val, lhs.ty, p.comp) and
+                    rhs.ty.signedness(p.comp) != .unsigned and
+                    lhs.ty.signedness(p.comp) != .unsigned) try p.errOverflow(mul.?, lhs);
```
Vexu, line 6492: Same applies here.

Vexu (changes requested):

src/aro/Parser.zig
bingcicle: Ah, thanks, good to know! Fixed

Vexu:
Can you add the check for subtraction and division as well?

PR #599, opened 2023-12-10
Preprocessor whitespace fixes

### LINKED ISSUE #597: Preprocessor: incorrect expansion with whitespace + paste operator
```c
#define CONTINUE(k) DEFER(k##_HOOK)()
#define DEFER(op) op EMPTY
#define EMPTY
#define F_HOOK() F

CONTINUE( F)
CONTINUE(F)
```

gcc and clang both produce:
```
F_HOOK ()
F_HOOK ()
```

aro produces:
```
 F
F_HOOK ()
```

The behavior was the same before the hideset change so I don't think it's related to that.

ehaas:
Treats whitespace-only paste arguments as placemarkers

Replace empty macros with placemarkers (this has the effect of blocking function-like macro expansion in some case, see `empty macro block expansion.c`

Closes #597


src/aro/Preprocessor.zig
```diff
@@ -1462,6 +1466,15 @@ fn handleBuiltinMacro(pp: *Preprocessor, builtin: RawToken.Id, param_toks: []con
     }
 }
 
+fn getPasteArgs(args: []const []const Token, end: u32) []const Token {
+    const placemarker = &[1]Token{tokFromRaw(.{ .id = .placemarker, .source = .generated })};
+    if (args[end].len == 0) return placemarker;
```
ehaas, line 1471: How does this actually work? It looks like it's returning an address of a stack variable - is it because it's a comptime-known constant and it's getting hoisted up to some static-lifetime?

Vexu: `tokFromRaw` is being called at runtime so it is indeed returning an address of a stack variable.

ehaas: I'm surprised it's not crashing then, lol. In any of the modes I tried, at least. I rewrote it so `getPasteArgs` returns an optional instead.

src/aro/Preprocessor.zig
```diff
@@ -1463,6 +1467,17 @@ fn handleBuiltinMacro(pp: *Preprocessor, builtin: RawToken.Id, param_toks: []con
     }
 }
 
+/// Treat whitespace-only paste arguments as empty
+fn getPasteArgs(args: []const Token) []const Token {
+    for (args) |tok| {
+        if (tok.id != .macro_ws) return args;
+    }
+    return &[1]Token{.{
```
ehaas, line 1475: Is this safe to do if there aren't any runtime dependencies (like the old `tokFromRaw` call)?

Vexu: Yes, it's doing pretty much the same as string literals.

Issue #585, opened 2023-11-27
Potential crash with non-integer array size


ehaas:
Strange one found via fuzzing. If the array size specifier is `.invalid`:

```c
[-w-f]" "f [f]" "f[-w-f]
```

```
panic: reached unreachable code
/Users/ehaas/source/arocc/src/aro/Type.zig:2345:21: 0x1080024ce in fromType (arocc)
            else => unreachable,
                    ^
/Users/ehaas/source/arocc/src/aro/Parser.zig:435:30: 0x108000b88 in typeStr (arocc)
    if (Type.Builder.fromType(ty).str(p.comp.langopts)) |str| return str;
                             ^
/Users/ehaas/source/arocc/src/aro/Parser.zig:2984:70: 0x10817ad8b in directDeclarator (arocc)
            try p.errStr(.array_size_non_int, size_tok, try p.typeStr(size.ty));
                                                                     ^
/Users/ehaas/source/arocc/src/aro/Parser.zig:2879:38: 0x1080e7d99 in declarator (arocc)
        d.ty = try p.directDeclarator(d.ty, &d, kind);
                                     ^
/Users/ehaas/source/arocc/src/aro/Parser.zig:1746:31: 0x1080e529c in initDeclarator (arocc)
        .d = (try p.declarator(decl_spec.ty, .normal)) orelse return null,
                              ^
```

Vexu:
Fixed by something like:
```diff
diff --git a/src/aro/Parser.zig b/src/aro/Parser.zig
index ebaa318..3b84321 100644
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -2980,6 +2980,7 @@ fn directDeclarator(p: *Parser, base_type: Type, d: *Declarator, kind: Declarato
         if (max_bits > 61) max_bits = 61;
         const max_bytes = (@as(u64, 1) << @truncate(max_bits)) - 1;
 
+        if (size.ty.is(.invalid)) return Type.invalid;
         if (!size.ty.isInt()) {
             try p.errStr(.array_size_non_int, size_tok, try p.typeStr(size.ty));
             return error.ParsingFailed;
```

ehaas:
Here's an alternate path to trigger it:
```c
struct Foo {
    int a;
} a;

char *f(char *arg) {
    return arg + a?a;
}
```

Vexu:
[...]

PR #601, opened 2023-12-14
Add protection from calling `typeStr` with an invalid type

ehaas:
This checks before calling typeStr with the places that I could generate a crash, and adds a check for non-debug modes so that they'll return an obvious error instead of illegal behavior.


ehaas:
Closes #585

ehaas:
There are no tests because you need lots of errors in the first place to generate these crashes and I don't think the messages themselves would be very useful. I tested manually by making a builtin with a return type of `.invalid` and doing various operations on the returned value.

Vexu:
Sorry to make you check all potential places where this could happen but seeing the change now I agree that it'd be better to just check the type in `typeStr` and return an error.

PR #608, opened 2023-12-29
Fix incorrect line numbers in linemarkers

### LINKED ISSUE #604: Incorrect line directives after `#include` line
Minimal test case, unsure exactly what else this affects (if anything):

`test.rc`:
```
#include "a.h"
foo
```
`a.h` is an empty file

---

```
arocc -E -fuse-line-directives --emulate=msvc test.rc
```
produces
```
#line 1 "test.rc"
#line 1 "<builtin>"
#line 294 "<builtin>"
#line 1 "<command line>"
#line 1 "test.rc"
#line 1 "./a.h"

#line 3 "test.rc"
foo
```

There is no line 3 of `test.rc`, so that last line directive should be for line 2. Note that adding more `#include` lines does not seem to compound the effect (it's still off-by-one).

---

Clang output for comparison (`clang -E -fuse-line-directives -xc test.rc`):

```
#line 1 "test.rc"
#line 1 "<built-in>"
#line 1 "<built-in>"
#line 398 "<built-in>"
#line 1 "<command line>"
#line 1 "<built-in>"
#line 1 "test.rc"
#line 1 "./a.h"
#line 2 "test.rc"
foo
```

squeek502:
Line numbers are 1-indexed, not 0-indexed. Before this commit, include_start/include_resume tokens were being added with a mix of 1-indexed and 0-indexed line numbers. After this commit, all line numbers are 1-indexed.

Closes #604

I would have added test cases, but there's nothing in place for testing linemarkers (if I add a test case, the file paths in the line markers end up as absolute). For example, this test case:

```c
//aro-args -E -fuse-line-directives
this is line 2







this is line 10
```

Expects this output:

```c
#line 2 "/home/ryan/Programming/zig/arocc/test/cases/linemarkers.c"
this is line 2
#line 10 "/home/ryan/Programming/zig/arocc/test/cases/linemarkers.c"
this is line 10
```

Let me know if you have an idea of how to add test cases for this. 

---

Fixes all the reproductions of the problem I was able to find.

---

```c
#include "a.h"


foo
```

Now correctly outputs:
```c
#line 1 "test.c"
#line 1 "<builtin>"
#line 293 "<builtin>"
#line 1 "<command line>"
#line 1 "test.c"
#line 1 "./a.h"

#line 4 "test.c"
foo
```

---

```c
this is line 1








this is line 10
```

Now correctly outputs:

```c
#line 1 "test.c"
#line 1 "<builtin>"
#line 293 "<builtin>"
#line 1 "<command line>"
#line 1 "test.c"
this is line 1
#line 10 "test.c"
this is line 10
```

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -248,7 +248,7 @@ pub fn preprocessSources(pp: *Preprocessor, sources: []const Source) Error!void 
         try pp.addIncludeStart(header);
         _ = try pp.preprocess(header);
     }
-    try pp.addIncludeResume(first.id, 0, 0);
+    try pp.addIncludeResume(first.id, 0, 1);
     const eof = try pp.preprocess(first);
     try pp.tokens.append(pp.comp.gpa, eof);
 }
@@ -290,7 +290,7 @@ pub fn addIncludeStart(pp: *Preprocessor, source: Source) !void {
     try pp.tokens.append(pp.gpa, .{ .id = .include_start, .loc = .{
         .id = source.id,
         .byte_offset = std.math.maxInt(u32),
-        .line = 0,
+        .line = 1,
     } });
 }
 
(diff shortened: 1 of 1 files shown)


Vexu:
Thanks, looks like I confused myself by not being able to decide whether to zero or one index the line numbers when implementing linemarkers.

Issue #611, opened 2024-01-04
Erroneous error with `--comments` option when a comment is after `#endif`


squeek502:
```c
#if 1
#endif // foo
```
or
```c
#if 1
#endif /* foo */
```

(also occurs if the comment is after a `#ifndef`/`#ifdef`/`#else` macro)

Current result:

```
> arocc -E -P --comments comments.c
comments.c:2:8: error: extra tokens at end of macro directive
#endif // foo
       ^
1 error generated.
```

Expected result (this is what `clang` gives):

```
> clang -E -P --comments -xc comments.c

```
(blank/empty output, no errors)

---

Note: If the `--comments` option is not used, there is no error:

```
> arocc -E -P comments.c

```

PR #613, opened 2024-01-06
Preprocessor: fix coments after if preprocessor directives

wrongnull:
Closes https://github.com/Vexu/arocc/issues/611

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -791,8 +791,7 @@ fn expectNl(pp: *Preprocessor, tokenizer: *Tokenizer) Error!void {
     while (true) {
         const tok = tokenizer.next();
         if (tok.id == .nl or tok.id == .eof) return;
-        if (tok.id == .whitespace) continue;
-        if (tok.id == .comment and pp.comp.langopts.preserve_comments) continue;
+        if (tok.id == .whitespace or tok.id == .comment) continue;
         if (!sent_err) {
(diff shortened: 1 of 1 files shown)


wrongnull:
also I'm not sure if a test case is needed here @Vexu

src/aro/Preprocessor.zig
```diff
@@ -792,6 +792,7 @@ fn expectNl(pp: *Preprocessor, tokenizer: *Tokenizer) Error!void {
         const tok = tokenizer.next();
         if (tok.id == .nl or tok.id == .eof) return;
         if (tok.id == .whitespace) continue;
+        if (tok.id == .comment and pp.comp.langopts.preserve_comments) continue;
```
Vexu, line 795: I don't think a test is necessary but `tok.id == .comment` is only possible when `preserve_comments == true` so no need to check again. Similar how it is with `.whitespace`. Only one of these that needs to be checked in the Preprocessor is `preserve_comments_in_macros` but here we are not in a macro.

wrongnull: > I don't think a test is necessary but `tok.id == .comment` is only possible when `preserve_comments == true` so no need to check again. Similar how it is with `.whitespace`. Only one of these that needs to be checked in the Preprocessor is `preserve_comments_in_macros` but here we are not in a macro.

Not shure what you mean but check has been rewritten

PR #620, opened 2024-02-01
Parser: improve return detection for switch statments

### LINKED ISSUE #619: Incorrect warning for non unreachable code after switch with break
The following code:
```c
int test(int a){
	switch(a) {
		default: a=1; break;
	}
	if (a) return 0;
	return 1;
}
```
Generate this warning:
```
test.c:5:2: warning: unreachable code [-Wunreachable-code]
        if (a) return 0;
 ^
1 warning generated.
```
The warning is not generated when the break is removed.

danielsan901998:
resolve #619
The switch statement was not handled correctly by nodeIsNoreturn, since break and continue returned .yes like return_stmt that caused that it was treated like a return from the function instead of just exiting the switch statement.
Also added handling for the default statement for a simple switch+default, the case statement is not handled for now because it requires more complex reasoning when combined with multiple cases or with an default statement inside a compound statement, so it return the default .no value.
And changed the check of compound_stmt to try to find a statement with a return instead of just looking the last one.
Since switch_stmt no longer return .complex i have remove it for now, since if_then_else_stmt will no longer propagate it and no other statement use it.
Because the nodeIsNoreturn function is also used to detect the cases of unreachable code, changing break and continue cause the analysis to no longer detect this type of case:
```c
int test(int a){
	switch(a){
		break;
		a=1;
	}
	return 1;
}
```
This is still not a complete solution since loop statements are not handled correctly, generating a "does not return a value" warning:
```c
int test(int a){
	while(1)
		return 1;
}
```

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -4603,24 +4603,31 @@ fn nodeIsNoreturn(p: *Parser, node: NodeIndex) NoreturnKind {
         },
         .compound_stmt_two => {
             const data = p.nodes.items(.data)[@intFromEnum(node)];
-            if (data.bin.rhs != .none) return p.nodeIsNoreturn(data.bin.rhs);
-            if (data.bin.lhs != .none) return p.nodeIsNoreturn(data.bin.lhs);
+            const lhs_type = if (data.bin.lhs != .none) p.nodeIsNoreturn(data.bin.lhs) else .no;
+            const rhs_type = if (data.bin.rhs != .none) p.nodeIsNoreturn(data.bin.rhs) else .no;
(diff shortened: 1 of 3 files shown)


Vexu:
I'd prefer a solution that makes `nodeIsNoreturn` return false negatives instead of false positives. The loop and switch cases could be changed to return `.complex` since the current code does not cover anything but the simplest cases.

danielsan901998:
Force-pushed and deleted my changes by mistake when trying to pull changes.
So i'm reopening this again with the changes requested.

Vexu:
Thanks!

PR #622, opened 2024-02-07
Parser: allow `~` operator for complex conjugation

ehaas:
Question: should this be its own Tree.Tag or should it reuse `.bit_not_expr`?


Vexu:
Clang also uses `UO_Not` for this and we don't have a separate tag for pointer arithmetic either so I think reusing it is fine.

PR #624, opened 2024-02-10
Parser: implement stdckdint.h

### LINKED ISSUE #578: C23 support: Add `stdckdint.h`
See section 7.20 https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3047.pdf

We can probably just make these macros which expand to the corresponding `__builtin_OP_overflow` builtins e.g. https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html#index-_005f_005fbuiltin_005fadd_005foverflow

ehaas:
Closes #578

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -4794,7 +4794,11 @@ const CallExpr = union(enum) {
                 Builtin.tagFromName("__va_start").?,
                 Builtin.tagFromName("va_start").?,
                 => arg_idx != 1,
-                Builtin.tagFromName("__builtin_complex").? => false,
+                Builtin.tagFromName("__builtin_complex").?,
+                Builtin.tagFromName("__builtin_add_overflow").?,
+                Builtin.tagFromName("__builtin_sub_overflow").?,
+                Builtin.tagFromName("__builtin_mul_overflow").?,
+                => false,
                 else => true,
             },
         };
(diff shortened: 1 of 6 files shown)


ehaas:
Should we move `CallExpr` into its own file (mostly to hold the custom-typechecking code for special builtins)?

src/aro/Parser.zig
```diff
@@ -4794,7 +4794,11 @@ const CallExpr = union(enum) {
                 Builtin.tagFromName("__va_start").?,
                 Builtin.tagFromName("va_start").?,
                 => arg_idx != 1,
-                Builtin.tagFromName("__builtin_complex").? => false,
+                Builtin.tagFromName("__builtin_complex").?,
```
Vexu, line 4797: Are we sure these are memoized properly? Getting a bit worried about something like https://github.com/ziglang/zig/issues/7948 happening here

Vexu (approved):
LGTM!

>Should we move `CallExpr` into its own file (mostly to hold the custom-typechecking code for special builtins)?

It would be nice to keep `Parser.zig` from getting too big if you can extract it cleanly.

src/aro/Parser.zig
ehaas: I'm not sure how to know if that's happening - I had to up the eval branch quote for `checkVarArg` - is that why you ask?

Vexu: > is that why you ask?

No just because there are so many calls now. I checked with `@compileLog` and they are getting memoized properly.

PR #627, opened 2024-02-18
Fix fuzz build

ehaas:
I'm not sure if this is the correct solution; but without this it was generating a static library instead of a .c file.


build.zig
```diff
@@ -13,6 +13,7 @@ fn addFuzzStep(b: *Build, target: std.Build.ResolvedTarget, afl_clang_lto_path:
     const fuzz_step = b.step("fuzz", "Build executable for fuzz testing.");
     var fuzz_target = target;
     fuzz_target.result.ofmt = .c;
+    fuzz_target.query.ofmt = .c;
```
Vexu, line 16: `resolveTargetQuery` does the same thing but if you want to avoid touching internal parts you can use that instead.
```suggestion
    const fuzz_target = blk: {
        var query = target.query;
        query.ofmt = .c;
        break :blk b.resolveTargetQuery(query);
    };
```

Vexu (approved):

PR #635, opened 2024-02-27
Value: support complex values

ehaas:
Still some work to do but this is a first pass at adding complex number support. Only floats currently, still need to handle int <-> complex and complex initializers (2-element init lists for complex produce a complex value)

### DIFF
--- a/src/aro/Builtins/eval.zig
+++ b/src/aro/Builtins/eval.zig
@@ -1,0 +1,41 @@
+const std = @import("std");
+const backend = @import("backend");
+const Interner = backend.Interner;
+const Builtins = @import("../Builtins.zig");
+const Builtin = Builtins.Builtin;
+const Parser = @import("../Parser.zig");
+const Tree = @import("../Tree.zig");
+const NodeIndex = Tree.NodeIndex;
+const Type = @import("../Type.zig");
+const Value = @import("../Value.zig");
+
+pub fn eval(tag: Builtin.Tag, p: *Parser, args: []const NodeIndex) !Value {
+    _ = args;
+
+    const builtin = Builtin.fromTag(tag);
+    if (!builtin.properties.attributes.const_evaluable) return .{};
+
+    switch (tag) {
+        Builtin.tagFromName("__builtin_inff").?,
+        Builtin.tagFromName("__builtin_inf").?,
+        Builtin.tagFromName("__builtin_infl").?,
+        => {
+            const ty: Type = switch (tag) {
+                Builtin.tagFromName("__builtin_inff").? => .{ .specifier = .float },
+                Builtin.tagFromName("__builtin_inf").? => .{ .specifier = .double },
+                Builtin.tagFromName("__builtin_infl").? => .{ .specifier = .long_double },
+                else => unreachable,
+            };
+            const f: Interner.Key.Float = switch (ty.bitSizeof(p.comp).?) {
+                32 => .{ .f32 = std.math.inf(f32) },
+                64 => .{ .f64 = std.math.inf(f64) },
+                80 => .{ .f80 = std.math.inf(f80) },
+                128 => .{ .f128 = std.math.inf(f128) },
+                else => unreachable,
+            };
+            return Value.intern(p.comp, .{ .float = f });
+        },
+        else => {},
+    }
+    return .{};
+}
--- a/src/aro/Builtins/eval.zig
+++ b/src/aro/Builtins/eval.zig
@@ -58,11 +58,7 @@ pub fn eval(tag: Builtin.Tag, p: *Parser, args: []const NodeIndex) !Value {
                 .unknown => {},
                 .finite => return Value.zero,
                 .positive => return Value.one,
-                .negative => {
-                    var negative_one: Value = undefined;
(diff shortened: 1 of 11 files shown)


src/backend/Interner.zig
```diff
@@ -71,6 +73,12 @@ pub const Key = union(enum) {
         f80: f80,
         f128: f128,
     };
+    pub const Complex = union(enum) {
+        cf32: [2]Ref,
+        cf64: [2]Ref,
+        cf80: [2]Ref,
+        cf128: [2]Ref,
```
Vexu, line 80: Since these are complex floats of specific size you should be able to use the specific float types.
```suggestion
        cf32: [2]f32,
        cf64: [2]f64,
        cf80: [2]f80,
        cf128: [2]f128,
```
Integer support can then be added later in a similar manner.

src/aro/Value.zig
```diff
@@ -390,9 +520,65 @@ pub fn sub(res: *Value, lhs: Value, rhs: Value, ty: Type, comp: *Compilation) !b
...
     const bits: usize = @intCast(ty.bitSizeof(comp).?);
-    if (ty.isFloat()) {
+    if (ty.isComplex()) {
+        res.* = switch (bits) {
+            64 => try lhs.complexMul(rhs, f32, comp),
+            128 => try lhs.complexMul(rhs, f64, comp),
+            160 => try lhs.complexMul(rhs, f80, comp),
+            256 => try lhs.complexMul(rhs, f128, comp),
+            else => unreachable,
+        };
```
Vexu, line 579: For future int support it'd be nice if this `isComplex` check was moved to the `isFloat` branch which then called `complexMulFloat`.

Vexu (commented):
I know this is a draft but here are a few ideas that could simplify some parts.

src/backend/Interner.zig
```diff
@@ -154,6 +168,13 @@ pub const Key = union(enum) {
                 128 => return .f128,
                 else => unreachable,
             },
+            .complex_ty => |bits| switch (bits) {
```
ehaas, line 171: I'm not sure how the type interner is used; should these be e.g. 64 bits for `cf32` since the type is 64 bits

src/aro/Builtins/eval.zig
```diff
@@ -0,0 +1,90 @@
...
+            const val = p.value_map.get(args[0]) orelse break :blk;
+            switch (val.isInfSign(p.comp)) {
+                .unknown => {},
+                .finite => return Value.zero,
+                .positive => return Value.one,
+                .negative => {
+                    var negative_one: Value = undefined;
+                    _ = try negative_one.sub(Value.zero, Value.one, .{ .specifier = .int }, p.comp);
+                    return negative_one;
+                },
```
Vexu, line 65:
```suggestion
                .negative => return Value.int(-1, p.comp),
```

src/aro/Parser.zig
```diff
@@ -3248,15 +3325,19 @@ fn initializer(p: *Parser, init_ty: Type) Error!Result {
         return error.ParsingFailed;
     }
 
-    var il: InitList = .{};
-    defer il.deinit(p.gpa);
+    if (init_ty.isComplex()) {
+        return p.complexInitializer(init_ty);
+    } else {
```
Vexu, line 3330: No need for an else branch here.

src/aro/Value.zig
```diff
@@ -326,18 +432,52 @@ pub fn toInt(v: Value, comptime T: type, comp: *const Compilation) ?T {
...
-        res.* = try intern(comp, .{ .float = f });
+        if (ty.isComplex()) {
+            res.* = switch (bits) {
+                64 => try complexAddSub(lhs, rhs, f32, .add, comp),
+                128 => try complexAddSub(lhs, rhs, f64, .add, comp),
+                160 => try complexAddSub(lhs, rhs, f80, .add, comp),
+                256 => try complexAddSub(lhs, rhs, f128, .add, comp),
+                else => unreachable,
+            };
+        } else {
```
Vexu, line 470: In places like this I'd also prefer an extra return in the then branch rather than an else branch.

src/backend/Interner.zig
Vexu: I think the way it's currently written is good since both parts are the same size. It might need to change when complex int support is added but that's a problem for then.

src/aro/Parser.zig
```diff
@@ -3248,15 +3325,19 @@ fn initializer(p: *Parser, init_ty: Type) Error!Result {
         return error.ParsingFailed;
     }
 
-    var il: InitList = .{};
-    defer il.deinit(p.gpa);
+    if (init_ty.isComplex()) {
+        return p.complexInitializer(init_ty);
```
Vexu, line 3329: Shouldn't this be handled somewhere in `initializerItem` instead? How are complex numbers in structs handled?

Vexu (approved):

src/aro/Parser.zig
ehaas: Those still work. I don't love that complex initializers are on a separate path but the problem I ran into is that braced initializers (`Parser.initializer`) do not return a Value in the result, because they use initializerItem / convertInitList. I'm not super familiar with the initializer code but it seemed like it would be tricky to add that. If you think it wouldn't be, I'm happy to take another look at it.

PR #640, opened 2024-03-05
Don't create implicit function definitions during record parsing

### LINKED ISSUE #639: Segfault rendering bad attributes
The following code segfaults when run with `arocc --verbose-ast`

```c
struct S {
  __attribute__((__aligned__(x(long)))) long a;
  __attribute__((packed)) b;
};

void foo(void) {
}
```

```
Segmentation fault at address 0x10
/Users/ehaas/source/arocc/src/aro/Tree.zig:699:10: 0x1021dafbd in dumpFieldAttributes__anon_9646 (arocc)
    for (attributes) |attr| {
         ^
Unwind error at address `arocc:0x1021dafbd` (error.InvalidUnwindInfo), trace may be incomplete
```

ehaas:
This prevents the parser from getting confused about the number of fields if a parser error triggers `parseOrNextDecl` to skip to the next decl.

Fixes #639

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -7680,6 +7680,8 @@ fn primaryExpr(p: *Parser) Error!Result {
                 else
                     try p.errStr(.implicit_func_decl, name_tok, name);
 
+                if (p.record.kind != .invalid) return .{ .ty = Type.invalid };
+
                 const func_ty = try p.arena.create(Type.Func);
                 func_ty.* = .{ .return_type = .{ .specifier = .int }, .params = &.{} };
                 const ty: Type = .{ .specifier = .old_style_func, .data = .{ .func = func_ty } };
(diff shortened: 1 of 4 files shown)


src/aro/Parser.zig
```diff
@@ -7680,6 +7680,8 @@ fn primaryExpr(p: *Parser) Error!Result {
                 else
                     try p.errStr(.implicit_func_decl, name_tok, name);
 
+                if (p.record.kind != .invalid) return .{ .ty = Type.invalid };
```
Vexu, line 7683: This line looks quite strange out of context, could you add a small comment explaining it? Something like `do not create implicit functions when parsing record declarators` if I'm understanding correctly.

Vexu (approved):

ehaas:
Just found another similar crash even with this fix applied, so perhaps a more robust fix is needed:
```c
struct S {
  __attribute__((__aligned__(char:  1))) long a;
  __attribute__((packed)) b;
};

void foo(void) {
}

```

PR #651, opened 2024-03-21
Zig update: use std.posix namespace instead of std.os

ehaas:


src/aro/Driver/Filesystem.zig
```diff
@@ -56,7 +56,7 @@ fn existsFake(entries: []const Filesystem.Entry, path: []const u8) bool {
 }
 
 fn canExecutePosix(path: []const u8) bool {
-    std.os.access(path, std.os.X_OK) catch return false;
+    std.posix.access(path, std.posix.X_OK) catch return false;
```
andrewrk, line 59: [std.Dir.access](https://ziglang.org/documentation/master/std/#std.fs.Dir.access)

ehaas: Ah, thank you! I looked in std.fs because of the doc comment in posix.zig but didn't think to look in Dir.zig

ehaas: I updated the "file exists" access check to use the Dir api but it doesn't look like there is one for checking whether a path is executable? This code is used to search for the linker so I think it will be fairly platform-specific anyway; in that case do you think it's ok to use the std.posix `access`?

andrewrk: Checking whether something is executable is certainly a platform-specific question to be asking. There is not currently a way to ask this question using the cross platform APIs in Zig, although it could be added. For example if the question being asked is, "if the platform supports the concept of an executable bit, return it, otherwise null" then that API would be a cross-platform one. But as of right now, that use case is indeed handled by `std.posix.access`.

src/aro/Driver/Filesystem.zig
```diff
@@ -173,7 +173,7 @@ pub const Filesystem = union(enum) {
     pub fn exists(fs: Filesystem, path: []const u8) bool {
         switch (fs) {
             .real => {
-                std.os.access(path, std.os.F_OK) catch return false;
+                std.fs.cwd().access(path, .{}) catch return false;
```
Vexu, line 176: The `real` field should be changed to a `Dir` instead of `void`.

PR #670, opened 2024-04-06
Parser: boolcast lhs of logical `or` expression even if not evaluating it

### LINKED ISSUE #668: Crash: assign boolean or result to float
```c
void foo(void) {
    float f = 2.0 || 0;
}
```

```
panic: access of union field 'int' while field 'float' is active
/Users/ehaas/source/arocc/src/aro/Value.zig:205:46: 0x103d44402 in intToFloat (arocc)
    return switch (comp.interner.get(v.ref()).int) {
                                             ^
/Users/ehaas/source/arocc/src/aro/Parser.zig:5589:35: 0x103d15b0f in floatCast (arocc)
            try res.val.intToFloat(float_ty, p.comp);
                                  ^
/Users/ehaas/source/arocc/src/aro/Parser.zig:6086:34: 0x103ca033c in coerceExtra (arocc)
                try res.floatCast(p, unqual_ty);
                                 ^
```

ehaas:
Ensure that the value, if present, is zero or one.

Closes #668

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -6415,6 +6415,8 @@ fn lorExpr(p: *Parser) Error!Result {
         if (try lhs.adjustTypes(tok, &rhs, p, .boolean_logic)) {
             const res = lhs.val.toBool(p.comp) or rhs.val.toBool(p.comp);
             lhs.val = Value.fromBool(res);
+        } else {
+            lhs.val.boolCast(p.comp);
         }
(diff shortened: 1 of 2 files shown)


Vexu:
Doesn't this also happen with `landExpr`?

ehaas:
With `landExpr` it short-circuits if the value is zero, but all zeroes have the same representation since they're a special value.

PR #677, opened 2024-04-08
Value: compare for zero against float instead of Value

### LINKED ISSUE #675: Crash: assign imaginary float to unsigned
Only happens with unsigned integer types

```c
unsigned long foo = 2.0i;
```

```
panic: access of union field 'int' while field 'complex' is active
/Users/ehaas/source/arocc/src/aro/Value.zig:352:48: 0x109d5c1ee in toBigInt (arocc)
    return switch (comp.interner.get(val.ref()).int) {
                                               ^
/Users/ehaas/source/arocc/src/aro/Value.zig:876:36: 0x109d8b0ae in compare (arocc)
    const lhs_bigint = lhs.toBigInt(&lhs_bigint_space, comp);
                                   ^
/Users/ehaas/source/arocc/src/aro/Value.zig:146:57: 0x109d8a822 in floatToInt (arocc)
    } else if (dest_ty.isUnsignedInt(comp) and v.compare(.lt, zero, comp)) {
                                                        ^
```

ehaas:
We can't compare complex values; toFloat grabs just the real part. This logic will need to be updated when complex int values are supported.

Fixes #675

### DIFF
--- a/build/GenerateDef.zig
+++ b/build/GenerateDef.zig
@@ -198,7 +198,7 @@ fn generate(self: *GenerateDef, input: []const u8) ![]const u8 {
         if (self.kind == .named) {
             try writer.writeAll("pub const Tag = enum {\n");
             for (values.keys()) |property| {
-                try writer.print("    {s},\n", .{std.zig.fmtId(property)});
+                try writer.print("    {p},\n", .{std.zig.fmtId(property)});
             }
             try writer.writeAll(
                 \\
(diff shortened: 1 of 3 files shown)


ehaas:
Added a fix for a separate crash (comparing two complex values with `!=`) and a zig update for `fmtId`. According to https://github.com/ziglang/zig/pull/18920 `{p}` can be used for the old behavior of `{s}`

PR #694, opened 2024-04-29
Value: use interned values for min and max int for types

ehaas:
This allows _BitInt's to use min/max int


src/aro/Compilation.zig
```diff
@@ -994,13 +994,28 @@ fn generateVaListType(comp: *Compilation) !Type {
 fn generateIntMax(comp: *const Compilation, w: anytype, name: []const u8, ty: Type) !void {
     const bit_count: u8 = @intCast(ty.sizeof(comp).? * 8);
     const unsigned = ty.isUnsignedInt(comp);
-    const max = if (bit_count == 128)
-        @as(u128, if (unsigned) std.math.maxInt(u128) else std.math.maxInt(u128))
-    else
-        ty.maxInt(comp);
+    const max: u128 = switch (bit_count) {
```
ehaas, line 997: This (and `wcharMax`) is to avoid `Compilation` depending on `Value` but can be changed if you want

Vexu: I like it better this way.

Vexu (approved):

Issue #701, opened 2024-05-03
Warn if implicit integer conversion truncates value


ehaas:
```c
unsigned char x = 1000;
```
We currently issue no warning; it should look something like:
```
test.c:1:19: warning: implicit conversion from 'int' to 'unsigned char' changes value from 1000 to 232 [-Wconstant-conversion]
unsigned char x = 1000;
                  ^
```

`Value.intCast` currently returns `void` on success; we can have it return a value that indicates whether truncation happens and then issue a diagnostic if so (similar to how `Value.floatToInt` is handled)

ehaas:
Closed by #717

PR #717, opened 2024-07-03
Parser: Report a Warning if implicit integer conversion truncates value or sign conversion.

Organ1sm:
When I wanted to learn how to write a C23 C language compiler, this project was extremely helpful to me, no less than having a rich big meal right after finishing a workout. I've been going through this project almost commit by commit, although I haven't finished looking through it all yet. This project not only taught me how to write a C compiler for the C23 standard, but also allowed me to learn a new language, Ziglang, which made me feel very happy. So I want to see if I can do something for this project. I looked at the issue section, and because I don't have experience contributing to open-source projects, I can only start with issues labeled as good first issue.

so I attempted to solve this issue.  #701 

Value.intCast` currently  return a value that indicates whether truncation happens and then issue a diagnostic.

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -2623,7 +2623,7 @@ fn enumSpec(p: *Parser) Error!Type {
                 continue;
 
             const symbol = p.syms.getPtr(field.name, .vars);
-            try symbol.val.intCast(dest_ty, p.comp);
+            _ = try symbol.val.intCast(dest_ty, p.comp);
             symbol.ty = dest_ty;
             p.nodes.items(.ty)[@intFromEnum(field_nodes[i])] = dest_ty;
             field.ty = dest_ty;
@@ -5574,7 +5574,11 @@ pub const Result = struct {
                 try res.implicitCast(p, .complex_float_to_complex_int);
             }
         } else if (!res.ty.eql(int_ty, p.comp, true)) {
-            try res.val.intCast(int_ty, p.comp);
+            const old_val = res.val;
+            const value_change_kind = try res.val.intCast(int_ty, p.comp);
+            if (value_change_kind == .value_changed and res.ty.makeIntegerUnsigned().specifier != int_ty.makeIntegerUnsigned().specifier)
+                try p.errStr(.int_value_changed, tok, try p.valueChangedStr(res, old_val, int_ty));
+
             const old_real = res.ty.isReal();
             const new_real = int_ty.isReal();
             if (old_real and new_real) {
@@ -5674,7 +5678,7 @@ pub const Result = struct {
             res.ty = ptr_ty;
             try res.implicitCast(p, .bool_to_pointer);
         } else if (res.ty.isInt()) {
-            try res.val.intCast(ptr_ty, p.comp);
+            _ = try res.val.intCast(ptr_ty, p.comp);
             res.ty = ptr_ty;
             try res.implicitCast(p, .int_to_pointer);
         }
@@ -6007,7 +6011,7 @@ pub const Result = struct {
                     try p.errStr(.cast_to_incomplete_type, l_paren, try p.typeStr(to));
                     return error.ParsingFailed;
                 }
-                try res.val.intCast(to, p.comp);
+                _ = try res.val.intCast(to, p.comp);
             }
         } else if (to.get(.@"union")) |union_ty| {
             if (union_ty.data.record.hasFieldOfType(res.ty, p.comp)) {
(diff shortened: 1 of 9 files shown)


src/aro/Parser.zig
```diff
@@ -5574,7 +5574,11 @@ pub const Result = struct {
                 try res.implicitCast(p, .complex_float_to_complex_int);
             }
         } else if (!res.ty.eql(int_ty, p.comp, true)) {
-            try res.val.intCast(int_ty, p.comp);
+            const old_val = res.val;
+            const value_change_kind = try res.val.intCast(int_ty, p.comp);
+            if (value_change_kind == .value_changed and !(res.ty.makeIntegerUnsigned().specifier == int_ty.makeIntegerUnsigned().specifier))
```
ehaas, line 5579: Is there a specific scenario the `!(res.ty.makeIntegerUnsigned().specifier == int_ty.makeIntegerUnsigned().specifier)` part guards against (probably better expressed using `!=` instead)?

ehaas:
The windows `binary expressions.c` test is failing because `long` is only 4 bytes on windows, so `0xFFFFFFFFFF` gets promoted to `long long`. Changing line 48 to `int baz = 0xFFFFFFFFFFLL + 1u;` and updating the error text to `long long` will make it have the same behavior on all platforms.

ehaas:
For the 2 new warnings in `enum overflow.c` it looks like `C` and `H` really do have type `int` on windows (confirmed on godbolt https://godbolt.org/z/x4PdsKEcs) - so for those two it looks like we should be comparing against `-1` in the windows static asserts

ehaas:
BTW thank you for the kind words and contribution!

Organ1sm:
According to your advice, I revised the code and pushed it. Thank you for your code review. :)

Organ1sm:
Hello, it seems this PR hasn't shown any signs of being merged. I'd like to know if there are any parts of the code that aren't well written so I can continue to improve them.

ehaas:
I had a question about the `res.ty.makeIntegerUnsigned().specifier != int_ty.makeIntegerUnsigned().specifier` condition - it's not clear to me why that's needed. Can you give an example that does the wrong thing if that condition is not included?

Organ1sm:
For example:   ` unsigned int u = -1;`
The left-hand side and right-hand side of this statement only differ in terms of being unsigned and signed. At the semantic level, there is an implicit conversion. So initially, my approach to solving this was not to add this extra condition, and I also wanted to add a warning for this implicit conversion. However, I later found that both [Clang](https://godbolt.org/z/s78rYPEhe) and [GCC](https://godbolt.org/z/9Wvhdvvcf) do not issue warnings for this statement.

ehaas:
I agree that adding a `sign-conversion` diagnostic option kind is the way to go. The bigint `bitCountTwosComp` function may be able to help with figuring out if the value changed due to truncation vs sign change.

src/aro/Value.zig
```diff
@@ -230,23 +230,51 @@ pub fn intToFloat(v: *Value, dest_ty: Type, comp: *Compilation) !void {
...
+
+    const dest_bits: usize = @intCast(dest_ty.bitSizeof(comp).?);
+    const dest_signed = dest_ty.signedness(comp) == .signed;
+
     var space: BigIntSpace = undefined;
     const big = v.toBigInt(&space, comp);
+    const value_bits = big.bitCountTwosComp();
+
+    // if big is negative or zero, then is signed.
+    const src_signed = (!big.positive or big.limbs[0] == 0);
```
ehaas, line 255:
```suggestion
    const src_signed = (!big.positive or big.eqlZero());
```

ehaas:
Looks good to me!

src/aro/Value.zig
```diff
@@ -230,23 +230,51 @@ pub fn intToFloat(v: *Value, dest_ty: Type, comp: *Compilation) !void {
...
+
+    const dest_bits: usize = @intCast(dest_ty.bitSizeof(comp).?);
+    const dest_signed = dest_ty.signedness(comp) == .signed;
+
     var space: BigIntSpace = undefined;
     const big = v.toBigInt(&space, comp);
+    const value_bits = big.bitCountTwosComp();
+
+    // if big is negative or zero, then is signed.
+    const src_signed = (!big.positive or big.eqlZero());
```
Vexu, line 255: Why is zero considered signed? Clang and gcc don't do so.

Vexu (commented):
Thanks for the enhancement and I'm glad the project has been helpful to you!

src/aro/Value.zig
Organ1sm: yes，I tested the behavior of clang and gcc. It was my oversight, so I wrote a new commit to correct this.

PR #719, opened 2024-07-06
New preprocessor, work in progress

ehaas:
This is a rewrite of the Preprocessor, based on the algorithm described in https://www.spinellis.gr/blog/20060626/cpp.algo.pdf

Main inspiration was that the existing one doesn't work with some more esoteric cases, such as https://github.com/Hirrolot/metalang99 and was having trouble figuring out why.

Right now it's hidden behind a flag `-fnew-preprocessor`, and it only preprocesses; it isn't tied into the parser at all. And there is still a lot of work to do - currently I'm not handling expansion locations at all, and there are lots of missing smaller features + probably lots of memory leaks.

One big improvement is using a persistent treap for hideset management, which massively reduces the amount of copying when adding items to a set or performing union/intersection operations.


ehaas:
Gonna close this for now as it's just collecting dust at this point. Will try to get back to it eventually, as there are still some preprocessor edge cases that don't work.

Issue #700, opened 2024-05-03
Print _BitInt size in diagnostics


ehaas:
```c
_BitInt(10) x = 1.2;
```
Currently our warning looks like this:
```
test.c:1:17: warning: implicit conversion from 'double' to '_BitInt' changes value from 1.2 to 1 [-Wfloat-conversion]
_BitInt(10) x = 1.2;
                ^
```
When it should be
```
test.c:1:17: warning: implicit conversion from 'double' to '_BitInt(10)' changes value from 1.2 to 1 [-Wfloat-conversion]
_BitInt(10) x = 1.2;
                ^
```

1. Change the `unsigned_bit_int_too_small`, `signed_bit_int_too_small`, and `bit_int_too_big` messages to not take a string parameter.
2. Change `Builder.Specifier.str` to return `null` for the `_BitInt` specifiers
3. Handle `_BitInt` types in `Type.printPrologue`

PR #720, opened 2024-07-07
print _BitInt size in diagnostics

wrongnull:
Closes https://github.com/Vexu/arocc/issues/700
At first there was redundant _BitInt size check in ```Parser```
Also I didn't delete string parameter from ```unsigned_bit_int_too_small``` because there is a special case like
```c
_Complex unsigned _BitInt(10) x;
```
and I used this parameter for possible ```_Complex``` token
And I don't know so far why I got a segfault when I deleted ```_BitInt``` specifiers from ```Builder.Specifier.str```


Vexu:
> And I don't know so far why I got a segfault when I deleted `_BitInt` specifiers from `Builder.Specifier.str`

Because you didn't update `Type.dump`:
https://github.com/Vexu/arocc/blob/8a885d8f6b65c6ed44d56a9f51e0401d76e6940d/src/aro/Type.zig#L2646-L2650

You can use the `--verbose-ast` flag to test changes to it.

wrongnull:
> > And I don't know so far why I got a segfault when I deleted `_BitInt` specifiers from `Builder.Specifier.str`
> 
> Because you didn't update `Type.dump`:
> 
> https://github.com/Vexu/arocc/blob/8a885d8f6b65c6ed44d56a9f51e0401d76e6940d/src/aro/Type.zig#L2646-L2650
> 
> You can use the `--verbose-ast` flag to test changes to it.

Done. I also decided to always print signedness of _BitInt via @tagName() on flavor of consistency

Issue #718, opened 2024-07-05
Define __STDC_EMBED macros


ehaas:
In `Compilation.generateBuiltinMacros` add

```c
#define __STDC_EMBED_NOT_FOUND__ 0
#define __STDC_EMBED_FOUND__ 1
#define __STDC_EMBED_EMPTY__ 2
```

These are new in C23 but can safely be defined regardless of the standard being used since they are reserved identifiers.

https://en.cppreference.com/w/c/preprocessor/replace

PR #721, opened 2024-07-07
define __STDC macros from C23

wrongnull:
Closes https://github.com/Vexu/arocc/issues/718


ehaas:
Can you also update `test/cases/#embed.c` to use the new constants.

wrongnull:
> Can you also update `test/cases/#embed.c` to use the new constants.

Done

Issue #690, opened 2024-04-22
Integer literal promotion rules should emulate chosen compiler


ehaas:
GCC and clang behave differently with regard to unsuffixed integer literals; see comment here https://github.com/ziglang/zig/issues/19701#issuecomment-2067190162 Note: MSVC follows the clang behavior of promoting to unsigned long long

The current behavior is implemented here: https://github.com/Vexu/arocc/blob/f60a2deae086dc9f60eb0f5ab417eb9a59c6b44c/src/aro/Parser.zig#L8395-L8402

PR #722, opened 2024-07-14
emulate chosen compiler in integer literal promotion

wrongnull:
Closes https://github.com/Vexu/arocc/issues/690


src/aro/Diagnostics/messages.def
```diff
@@ -1294,11 +1294,15 @@ invalid_utf8
...
-implicitly_unsigned_literal
+implicitly_unsigned_literal_clang
     .msg = "integer literal is too large to be represented in a signed integer type, interpreting as unsigned"
     .opt = W("implicitly-unsigned-literal")
     .kind = .warning
 
+implicitly_unsigned_literal_gcc
+    .msg = "integer constant is so large that it is unsigned"
+    .kind = .warning
+
```
Vexu, line 1305: I don't think we need to emulate differently worded diagnostics too.

src/aro/Parser.zig
```diff
@@ -8448,7 +8448,7 @@ fn fixedSizeInt(p: *Parser, base: u8, buf: []const u8, suffix: NumberSuffix, tok
         const max_int = try Value.maxInt(res.ty, p.comp);
         if (interned_val.compare(.lte, max_int, p.comp)) break;
     } else {
-        res.ty = .{ .specifier = .ulong_long };
+        res.ty = .{ .specifier = if (p.comp.langopts.emulate == .gcc) .int128 else .ulong_long };
```
Vexu, line 8451: This needs to check that the target supports `__int128` with `target_utils.hasFloat128`.

rurban: hasFloat128??

Vexu: Sorry, `target_utils.hasInt128`. It looks like there is a helper for `hasFloat128` in `Compilation`, you could add the same for `hasInt128`.

wrongnull: Done

wrongnull:
@Vexu could you review changes now if you have a chance?

Vexu:
Thanks!

PR #725, opened 2024-07-20
refactor: replace std.meta.fields with @typeInfo

### LINKED ISSUE #724: Remove calls to `std.meta.fields`
From the Zig Discord:

```
andrewrk — Today at 6:09 PM
who's calling std.meta.fields anyway, it's a silly function
```

Calls to this function generally can be replaced with `@typeInfo(Foo).Struct.fields` (or `@typeInfo(Foo).Enum.fields` etc. depending on what `Foo` is.)

simonklee:
Updated instances of `std.meta.fields` to use `@typeInfo` for retrieving (struct) fields.

Fix #724

P.S. Haven't really done anything with Zig for a year so. This was just something I did while browsing the source code since I noticed the issue on GH.

### DIFF
--- a/src/aro/Attribute.zig
+++ b/src/aro/Attribute.zig
@@ -67,7 +67,7 @@ pub fn requiredArgCount(attr: Tag) u32 {
         inline else => |tag| {
             comptime var needed = 0;
             comptime {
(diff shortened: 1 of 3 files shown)


ehaas:
Thanks!

PR #726, opened 2024-07-23
Eliminate dead architecture tags

wrongnull:
see https://github.com/ziglang/zig/pull/20708
https://github.com/ziglang/zig/pull/20402


src/aro/target.zig
```diff
@@ -685,9 +641,8 @@ pub fn toLLVMTriple(target: std.Target, buf: []u8) []const u8 {
         .dragonfly => "dragonfly",
         .freebsd => "freebsd",
         .fuchsia => "fuchsia",
-        .kfreebsd => "kfreebsd",
         .linux => "linux",
-        .lv2 => "lv2",
+        .ps3 => "ps3",
```
Vexu, line 645: The LLVM name shouldn't be changed.

wrongnull: Why so? It has been changed here https://github.com/ziglang/zig/pull/20708/files#diff-73b15ac9aed4b9aca85583d8caa883b2a5211a63b411ae6ef088a6d8b92156f6L671

wrongnull: not sure what did you mean

wrongnull: Oh, I'm sorry. Done

PR #735, opened 2024-08-01
Source: remove call to deprecated std.unicode.utf8Decode function

ehaas:


ehaas:
Re-ran CI with latest Zig; green now.

PR #738, opened 2024-08-01
Compilation: store cwd in Compilation

ehaas:
Removes a dependency on global system state. The primary usage would be a library that wants to use aro with an arbitrary working directory without actually having to change the working directory of the process.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -127,22 +127,27 @@ types: struct {
 } = .{},
 string_interner: StrInt = .{},
 interner: Interner = .{},
+/// If this is not null, the directory containing the specified Source will be searched for includes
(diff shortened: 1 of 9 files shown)


src/aro/Compilation.zig
```diff
@@ -127,22 +127,27 @@ types: struct {
 } = .{},
 string_interner: StrInt = .{},
 interner: Interner = .{},
+/// If this is not null, the directory containing the specified Source will be searched for includes
+/// Used by MS extensions which allow searching for includes relative to the directory of the main source file.
 ms_cwd_source_id: ?Source.Id = null,
```
Vexu, line 132: Unrelated to this PR but this field should probably be moved to `Preprocessor`.

Vexu (approved):

PR #740, opened 2024-08-02
Build: Add a build.zig.zon file

ehaas:
Not sure how to handle the `version` field - we also compute it in build.zig

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,0 +1,19 @@
+.{
+    .name = "aro",
+
+    .version = "0.0.0",
+
+    .dependencies = .{},
+
+    .paths = .{
+        "build",
+        "build.zig",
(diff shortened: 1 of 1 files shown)


Vexu (approved):
> Not sure how to handle the version field - we also compute it in build.zig

I guess it should only be updated on tagged releases? Either way we can consider the version reported by the executable more accurate.

ehaas:
This is more of a Zig build system question, but my I was also wondering about the `include` directory - if someone is consuming `aro` as a package, is there any way to programmatically access that directory?

Vexu:
I'm guessing something like `b.dependency("aro").path("include/stddef.h")` should work?

ehaas:
That worked! I'll put together a quick README entry / example for using aro as a dependency.

Issue #349, opened 2022-07-28
Check array bounds for array parameters with `static` keyword 


ehaas:
```c
void foo(int x[static 10]) {

}

void bar(void) {
    int x[5];
    foo(x);
}
```

clang warns:
```
test.c:7:5: warning: array argument is too small; contains 5 elements, callee requires at least 10 [-Warray-bounds]
    foo(x);
    ^   ~
test.c:1:14: note: callee declares array parameter as static here
void foo(int x[static 10]) {
             ^~~~~~~~~~~~
```

PR #742, opened 2024-08-03
Check argument size on static array parameters

wrongnull:
Closes https://github.com/Vexu/arocc/issues/349

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -7685,7 +7685,7 @@ fn callExpr(p: *Parser, lhs: Result) Error!Result {
         }
         const p_ty = params[arg_count].ty;
         if (p_ty.statically_sized_array) {
-            if (arg.ty.data == .array and arg.ty.data.array.*.len != p_ty.data.array.*.len) {
+            if (arg.ty.data == .array and arg.ty.data.array.*.len < p_ty.data.array.*.len) {
                 const extra = Diagnostics.Message.Extra{ .arguments = .{
                     .expected = @intCast(arg.ty.data.array.*.len),
                     .actual = @intCast(p_ty.data.array.*.len),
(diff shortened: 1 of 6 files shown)


src/aro/Parser.zig
```diff
@@ -7683,6 +7684,18 @@ fn callExpr(p: *Parser, lhs: Result) Error!Result {
             continue;
         }
         const p_ty = params[arg_count].ty;
+        if (p_ty.statically_sized_array) {
+            if (arg.ty.data == .array and arg.ty.data.array.*.len < p_ty.data.array.*.len) {
```
Vexu, line 7688: This should use `arrayLen` instead of making the union tagged.

src/aro/Parser.zig
```diff
@@ -7683,6 +7684,18 @@ fn callExpr(p: *Parser, lhs: Result) Error!Result {
...
+                const extra = Diagnostics.Message.Extra{ .arguments = .{
+                    .expected = @intCast(arg.ty.data.array.*.len),
+                    .actual = @intCast(p_ty.data.array.*.len),
+                } };
+                try p.errExtra(.array_argument_too_small, param_tok, extra);
```
Vexu, line 7693: Clang also adds a `callee declares array parameter as static here` note. You should be able to do the same with `params[arg_count].name_tok`.

src/aro/Type.zig
```diff
@@ -425,6 +425,7 @@ data: union {
 specifier: Specifier,
 qual: Qualifiers = .{},
 decayed: bool = false,
+statically_sized_array: bool = false,
```
Vexu, line 428: `static_array` specifier exists but is seemingly unused currently. Could you try using that instead?

(Apparently it was only used for three days between c35d5c22f7adc314202c5dcd220d88d6ff7d068d and 9b4daaa8647d8ea7c24ed6edae66a32433fb4ba6)

Vexu (changes requested):

wrongnull:
@Vexu all the desired changes have been made

Issue #732, opened 2024-07-31
Proposal: add source locations to tree nodes


ehaas:
Currently most tree nodes do not include source location info (only `decl` and `decl_ref` data via `TokenIndex` fields). This is a problem for translate-c when trying to indicate the source location of record/enum decls and static asserts (and possibly other things). My proposal is to add a field `loc` of type `enum(u32) { none = std.math.maxInt(u32), _ }` to `Tree.Node`. The default value would be `.none`. If not `.none`, it would store the `TokenIndex` of the first token of the node. Doing it this way would allow us to add source locations incrementally (there are currently 89 calls to `addNode` in the parser).

PR #745, opened 2024-08-06
Preprocessor: Add debug dump options

ehaas:
Adds support for GCC's -dD, -dM, -dN flags which enable dumping macro names and/or definitions along with or instead of preprocessor output.

Added a `CHECK_PARTIAL_MATCH` option for preprocessor testing, because the output contains things like `__TIME__` which is difficult to fit into our existing preprocessor test framework.

`-dM` is also difficult to test extensively with the existing framework because it dumps the output in arbitrary order (just iterates over `pp.defines`) which seems to match the observed behavior of clang

### DIFF
--- a/src/aro/Driver.zig
+++ b/src/aro/Driver.zig
@@ -47,6 +47,19 @@ color: ?bool = null,
 nobuiltininc: bool = false,
 nostdinc: bool = false,
 nostdlibinc: bool = false,
+debug_dump_letters: packed struct(u3) {
+    d: bool = false,
+    m: bool = false,
+    n: bool = false,
+
+    /// Specifying letters whose behavior conflicts is undefined.
+    pub fn getPreprocessorDumpMode(self: @This()) Preprocessor.DumpMode {
+        if (self.d) return .macros_and_result;
+        if (self.m) return .macros_only;
+        if (self.n) return .macro_names_and_result;
+        return .result_only;
+    }
+} = .{},
 
 /// Full path to the aro executable
 aro_name: []const u8 = "",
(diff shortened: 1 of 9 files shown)


src/aro/Driver.zig
```diff
@@ -47,6 +47,19 @@ color: ?bool = null,
 nobuiltininc: bool = false,
 nostdinc: bool = false,
 nostdlibinc: bool = false,
+debug_dump_letters: packed struct(u3) {
+    d: bool = false,
+    m: bool = false,
+    n: bool = false,
+
+    /// Specifying letters whose behavior conflicts is undefined.
```
Vexu, line 55: This could specify that we match clang's behavior.

Vexu (approved):

src/aro/Driver.zig
ehaas: Does it actually match clang as it is now? I just chose the current order because it was alphabetical.

ehaas: Ok just did a quick check - clang doesn't support `-dN`; and `-dM` wins over `-dD` (it always does `-dM` regardless of order if `-dD` and `-dM` are provided). I'll update it to match that.

ehaas:
@Vexu Once this is merged, would you be able to sync aro to the main zig repo? I'd like to try updating https://github.com/ziglang/zig/blob/master/tools/generate_linux_syscalls.zig to use aro instead of `zig cc` (removing the clang dependency).

Is there a technical reason why Zig can't use aro as a `build.zig.zon` dependency? Doing so would eliminate the pain point of needing to update the Zig files generated from `.def` files.

Vexu:
> Once this is merged, would you be able to sync aro to the main zig repo?

Sure!



> Is there a technical reason why Zig can't use aro as a `build.zig.zon` dependency?

I don't think so as long as bootstrapping doesn't depend on it. I'd ask @andrewrk  before trying to get that to work since he tends to have opinions on stuff like that.

ehaas:
On second thought - let me see if I can make some progress on #732 before the sync.

PR #746, opened 2024-08-07
Fix AFLplusplus fuzz test

Reokodoku:


Reokodoku:
I think the CI failure is unrelated. `zig build test` runs without any error on zig `0.14.0-dev.842+a655c15c4`.

Issue #733, opened 2024-07-31
Proposal: treat typedef types as separate type specifier


ehaas:
Briefly discussed here https://github.com/Vexu/arocc/pull/433#issuecomment-1324825394

This would add a new field `.typedef` to `Type.Specifier` with an associated data `*Typedef`, with `Typedef` defined as:

```zig
const Typedef = struct {
   name: StringId,
   ty: Type,
};
```

This could potentially help solve #437 and #367. Furthermore this will be necessary for translate-c support, since we currently do not maintain type name information. For example, if `uint32_t` immediately resolves to `unsigned int`, we can't translate that as `u32` instead of `c_uint`. The same applies to typedef'd enums and records.

PR #757, opened 2024-08-21
Type: add helpers for managing typeof and attributed types

ehaas:
The idea behind this is to prepare for #733 by making typeof and attributed types slightly less annoying to deal with, by moving the recursive call logic to a single place (2 actually), which should make it easier to add a new `typedef` specifier which will generally just call into its subtype for most of the Type functions.

### DIFF
--- a/src/aro/Attribute.zig
+++ b/src/aro/Attribute.zig
@@ -1014,14 +1014,13 @@ pub fn applyEnumeratorAttributes(p: *Parser, ty: Type, attr_buf_start: usize) !T
 }
 
 fn applyAligned(attr: Attribute, p: *Parser, ty: Type, tag: ?Diagnostics.Tag) !void {
-    const base = ty.canonicalize(.standard);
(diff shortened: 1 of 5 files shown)


Vexu:
Wouldn't something more like `canonicalize` be better? This strategy already requires a duplicate function with a one arg difference.

Maybe a `Type.base()` which would use a while loop instead of recursion. Most places that use `canonicalize` seem like they actually want that kind of functionality.

ehaas:
Is the idea that each function that currently recurses on typeof/attributed types, should instead call canonicalize/base on its argument first? I think that would also work and yeah probably simpler.

ehaas:
This branch has a fairly heavy performance hit (~5% for large files), and even with *just* Type.base replacing `canonicalize` where we only want the specifier, it's still about 1.5% slower. So now I'm kinda leaning toward, we just implement typedef specifiers in the code as-is, and then hopefully clean everything up with Type interning in the future.

PR #767, opened 2024-09-06
Type: add a name field to track typedef names

ehaas:
Just an idea - as a stop-gap before #733 is implemented. This would allow more progress on aro translate-c since one blocker right now is the lack of typedef names. It was about 5 minutes of work so I'm not very strongly advocating for it if you think it's too much of a hack.

We had at least 4 bytes of padding available in `Type` because it doesn't change the size of `Type` in either Safe or Fast modes.

### DIFF
--- a/src/aro/SymbolStack.zig
+++ b/src/aro/SymbolStack.zig
@@ -196,7 +196,12 @@ pub fn defineTypedef(
         .kind = .typedef,
         .name = name,
         .tok = tok,
-        .ty = ty,
+        .ty = .{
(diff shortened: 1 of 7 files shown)


Vexu:
If you want to work on translate-c and this helps with it then I'm OK with it. Let me know if you need help syncing the dependency.

ehaas:
I don't mind trying to put together the dependency sync but is that the kind of commit that would need to come from a core team member?

Vexu:
Good point.

PR #768, opened 2024-09-08
Compilation: Add driver/frontend support for pic/pie options

ehaas:
Some time (real soon now™) I'd like to start writing a simple AST -> x86_64 assembly backend. No IR or optimizations or anything, just something that generates readable/debuggable assembly. 

Is that something that you think should be part of aro? Or a separate project that uses aro as a dependency? If it's part of aro I'd want to put it behind a `-fsimple-asm` or similar type of flag.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -3,7 +3,9 @@ const Allocator = mem.Allocator;
 const assert = std.debug.assert;
 const EpochSeconds = std.time.epoch.EpochSeconds;
 const mem = std.mem;
-const Interner = @import("backend").Interner;
+const backend = @import("backend");
+const Interner = backend.Interner;
+const CodeGenOptions = backend.CodeGenOptions;
 const Builtins = @import("Builtins.zig");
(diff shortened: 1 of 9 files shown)


Vexu (approved):
> Some time (real soon now™) I'd like to start writing a simple AST -> x86_64 assembly backend. No IR or optimizations or anything, just something that generates readable/debuggable assembly.
> 
> Is that something that you think should be part of aro? Or a separate project that uses aro as a dependency? If it's part of aro I'd want to put it behind a -fsimple-asm or similar type of flag.

Absolutely, any kind of functional codegen would be nice. And just `-S` is fine as far as flags go.

ehaas:
`-S` for assembly output makes sense; I was just thinking it would be nice if the toolchain could assemble and link the output; but I didn't want it to interfere with the existing `Ir` codepath. Maybe something like `-O0` for it? So `-O0` without `-S` would generate assembly to a temp file, invoke `as`, and then add the output to the link objects?

Vexu:
Right, didn't think of that, using `-O0` sounds good.

Issue #770, opened 2024-09-09
Implement `target.getPICMode`


ehaas:
Currently `getPICMode` in `src/aro/target.zig` does not consider the actual target. See https://github.com/llvm/llvm-project/blob/f12e10b513686a12f20f0c897dcc9ffc00cbce09/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1804 for reference of what this function should do.

ehaas:
Closed by #771

PR #771, opened 2024-09-17
Implement target.getPICMode

Organ1sm:
I tried to solve #770,  but full implementation of this function is too long.  So for now, i have only implemented these two target-related [default functions](https://github.com/llvm/llvm-project/blob/f12e10b513686a12f20f0c897dcc9ffc00cbce09/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1808).

 I'm unable to verify all the correctness.  Could please review this part first?  Once It's reviewed and passed, I'll submit the remaining code for `target.getPICMode` function.

### DIFF
--- a/src/aro/Driver.zig
+++ b/src/aro/Driver.zig
@@ -63,6 +63,8 @@ apple_kext: bool = false,
 mkernel: bool = false,
 mabicalls: ?bool = null,
 dynamic_nopic: ?bool = null,
+ropi: ?bool = null,
+rwpi: ?bool = null,
 debug_dump_letters: packed struct(u3) {
     d: bool = false,
     m: bool = false,
@@ -158,6 +160,10 @@ pub const usage =
     \\  -fPIC                   Similar to -fpic but avoid any limit on the size of the global offset table
     \\  -fpie                   Similar to -fpic, but the generated position-independent code can only be linked into executables
     \\  -fPIE                   Similar to -fPIC, but the generated position-independent code can only be linked into executables
+    \\  -frwpi                  Generate read-write position independent code (ARM only)  
+    \\  -fno-rwpi               Disable generate read-write position independent code (ARM only). 
+    \\  -fropi                  Generate read-only position independent code (ARM only)  
+    \\  -fno-ropi               Disable generate read-only position independent code (ARM only). 
     \\  -fshort-enums           Use the narrowest possible integer type for enums
     \\  -fno-short-enums        Use "int" as the tag type for enums
     \\  -fsigned-char           "char" is signed
@@ -340,6 +346,14 @@ pub fn parseArgs(
                 d.comp.langopts.allow_half_args_and_returns = true;
             } else if (pic_related_options.has(arg)) {
                 pic_arg = arg;
+            } else if (mem.eql(u8, arg, "-fropi")) {
+                d.ropi = true;
+            } else if (mem.eql(u8, arg, "-fno-ropi")) {
+                d.ropi = false;
+            } else if (mem.eql(u8, arg, "-frwpi")) {
+                d.rwpi = true;
+            } else if (mem.eql(u8, arg, "-fno-rwpi")) {
+                d.rwpi = false;
             } else if (mem.eql(u8, arg, "-fshort-enums")) {
                 d.comp.langopts.short_enums = true;
             } else if (mem.eql(u8, arg, "-fno-short-enums")) {
@@ -1067,25 +1081,15 @@ pub fn getPICMode(d: *Driver, args: []const []const u8, lastpic: []const u8) !st
     }
 
     var ropi = false;
-    const last_ropi_arg = getLastArg(args, StaticStringSet.initComptime(.{ .{"-fropi"}, .{"-fno-ropi"} }));
-    if (last_ropi_arg) |idx| {
-        if (std.mem.eql(u8, args[idx], "-fropi")) {
-            if (!embedded_pi_supported) {
-                try d.unsupportedOptionForTarget(target, args[idx]);
-            }
-            ropi = true;
-        }
+    if (d.ropi) |_| {
+        if (!embedded_pi_supported) try d.unsupportedOptionForTarget(target, "-fropi");
+        ropi = true;
     }
 
     var rwpi = false;
-    const last_rwpi_arg = getLastArg(args, StaticStringSet.initComptime(.{ .{"-frwpi"}, .{"-fno-rwpi"} }));
-    if (last_rwpi_arg) |idx| {
-        if (std.mem.eql(u8, args[idx], "-frwpi")) {
-            if (!embedded_pi_supported) {
-                try d.unsupportedOptionForTarget(target, args[idx]);
-            }
-            rwpi = true;
-        }
+    if (d.rwpi) |_| {
+        if (!embedded_pi_supported) try d.unsupportedOptionForTarget(target, "-frwpi");
+        rwpi = true;
     }
 
     // ROPI and RWPI are not compatible with PIC or PIE.
@@ -1120,12 +1124,4 @@ fn getLastArgValue(args: []const []const u8, option_name: []const u8) ?[]const u
         option_value = option(arg, option_name);
     }
     return option_value;
-}
-
-fn getLastArg(args: []const []const u8, options: StaticStringSet) ?usize {
-    var last_index: ?usize = null;
-    for (args, 0..) |arg, i| {
-        if (options.has(arg)) last_index = i;
-    }
-    return last_index;
 }
--- a/src/aro/Driver.zig
+++ b/src/aro/Driver.zig
@@ -535,7 +535,7 @@ pub fn parseArgs(
         return d.fatal("invalid value '{0s}' in '-fgnuc-version={0s}'", .{gnuc_version});
     }
     d.comp.langopts.gnuc_version = version.toUnsigned();
-    const pic_level, const is_pie = d.getPICMode(args);
+    const pic_level, const is_pie = try d.getPICMode(args);
     d.comp.code_gen_options.pic_level = pic_level;
     d.comp.code_gen_options.is_pie = is_pie;
     return false;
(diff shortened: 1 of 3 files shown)


src/aro/target.zig
```diff
@@ -723,6 +724,168 @@ pub fn toLLVMTriple(target: std.Target, buf: []u8) []const u8 {
     return stream.getWritten();
 }
 
+fn isPIEDefault(d: *const Driver) bool {
```
ehaas, line 727: These functions shouldn't depend on `Driver`; to make them more generalizable I would suggest that we have it return an enum with values `yes`, `no`, or `depends_on_linker`. And then resolve that to a bool in the caller where we know what the linker is.

ehaas:
Other than my one comment I did a quick check of some of the logic and it looks good.

src/aro/target.zig
```diff
@@ -2,6 +2,8 @@ const std = @import("std");
 const LangOpts = @import("LangOpts.zig");
 const Type = @import("Type.zig");
 const TargetSet = @import("Builtins/Properties.zig").TargetSet;
+const Driver = @import("Driver.zig");
```
ehaas, line 5: Looks like this `Driver` import is no longer needed.

src/aro/target.zig
```diff
@@ -723,9 +725,189 @@ pub fn toLLVMTriple(target: std.Target, buf: []u8) []const u8 {
...
 /// This currently just returns the desired settings without considering target defaults / requirements
-pub fn getPICMode(target: std.Target, desired_pic_level: ?backend.CodeGenOptions.PicLevel, wants_pie: ?bool) struct { backend.CodeGenOptions.PicLevel, bool } {
-    _ = target;
+pub fn getPICMode(
+    target: std.Target,
+    use_linker: ?[]const u8,
+    desired_pic_level: ?backend.CodeGenOptions.PicLevel,
+    wants_pie: ?bool,
+) struct { backend.CodeGenOptions.PicLevel, bool } {
+    const linker = use_linker orelse system_defaults.linker;
```
ehaas, line 890: Let's do this `orelse` at the callsite and pass in `linker: []const u8` to this function; then we don't need to import `system_defaults` either.

src/aro/target.zig
```diff
@@ -723,9 +725,189 @@ pub fn toLLVMTriple(target: std.Target, buf: []u8) []const u8 {
...
+    const is_pie = switch (isPIEDefault(target)) {
+        .yes => true,
+        .no => false,
+        .depends_on_linker => if (std.mem.eql(u8, linker, "bfd"))
```
ehaas, line 894: To match the clang code (I'm not sure how much that matters) we should use `std.ascii.startsWithIgnoreCase` here instead of `std.mem.eql`

src/aro/Driver.zig
```diff
@@ -535,7 +535,7 @@ pub fn parseArgs(
     }
     d.comp.langopts.gnuc_version = version.toUnsigned();
     const wants_pie: ?bool = if (d.desired_pie_level) |level| level != .none else null;
-    const pic_level, const is_pie = target_util.getPICMode(d.comp.target, d.desired_pic_level, wants_pie);
+    const pic_level, const is_pie = target_util.getPICMode(d.comp.target, d.use_linker, d.desired_pic_level, wants_pie);
```
Organ1sm, line 538:
```suggestion
   const linker = d.use_linker orelse system_default.linker; 
   const pic_level, const is_pie = target_util.getPICMode(d.comp.target, linker, d.desired_pic_level, wants_pie);
```

src/aro/target.zig
Organ1sm: I just made some comment [here](https://github.com/Vexu/arocc/pull/771/files#r1764488951) , such changes should have achieved the goal, right? But this still introduces system_default into the driver.

src/aro/target.zig
Organ1sm: In LLVM's [implementation](https://github.com/llvm/llvm-project/blob/f12e10b513686a12f20f0c897dcc9ffc00cbce09/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1874), it still relies on parameter options, but it provides a component for parameter processing，aro seem missing. If the getPICMode implementation does not depend on the driver, should the caller wrap the relevant parameter options and pass them through the parameters?

src/aro/target.zig
ehaas: It seems like having `isPICDefault` and `isPIEDefault` live in `target` (as they are now) and moving `getPICMode` to `Driver` is the way to go.

src/aro/Driver.zig
```diff
@@ -892,3 +891,231 @@ fn exitWithCleanup(d: *Driver, code: u8) noreturn {
...
+
+    var pie: bool = is_pie_default;
+    var pic: bool = pie or is_pic_default;
+    // The Darwin/MachO default to use PIC does not apply when using -static.
+    if (target.ofmt == .macho and d.static) {
+        pic, pie = .{ false, false };
+    }
+    var is_piclevel_two = pic;
+
+    const kernel_or_next: bool = d.hasArg(args, .{ "-mkernel", "-fapple_kext" });
```
ehaas, line 928: nitpick: this variable should be named `kernel_or_kext`

ehaas:
@Organ1sm Thanks for tackling this! I think it needs a rebase, and then the last thing to do is to remove `desired_pic_level` and `desired_pie_level` from `Driver` since they're handled in a different way now.

Organ1sm:
@ehaas I have removed the desired_pic/pie_level from the driver and filled in the diagnostic areas needed in the `getPICMode` function. 

Thank you very much for your suggestions during the code review period, which gave me a great experience. I am also very happy to be able to contribute to arocc.

src/aro/Driver.zig
```diff
@@ -892,3 +897,235 @@ fn exitWithCleanup(d: *Driver, code: u8) noreturn {
     }
     std.process.exit(code);
 }
+
+/// This currently just returns the desired settings without considering target defaults / requirements
```
ehaas, line 901: Looks like this comment is no longer accurate :)

src/aro/Driver.zig
```diff
@@ -310,20 +309,15 @@ pub fn parseArgs(
             } else if (mem.eql(u8, arg, "-fnative-half-arguments-and-returns")) {
                 d.comp.langopts.allow_half_args_and_returns = true;
             } else if (mem.eql(u8, arg, "-fno-pic") or mem.eql(u8, arg, "-fno-PIC") or mem.eql(u8, arg, "-fno-pie") or mem.eql(u8, arg, "-fno-PIE")) {
```
ehaas, line 311: We can just remove these `if` statements entirely

Organ1sm: Without these if statements, if there are options related to pic/pie in the arguments, it will directly report an error of type [`.cli_unknown_arg`.](https://github.com/Vexu/arocc/blob/406f323eb97aa2dd2e339ba194e28bfdc3a689cb/src/aro/Driver.zig#L509)

ehaas (approved):

src/aro/Driver.zig
```diff
@@ -892,3 +899,225 @@ fn exitWithCleanup(d: *Driver, code: u8) noreturn {
     }
     std.process.exit(code);
 }
+
+/// Parses the various -fpic/-fPIC/-fpie/-fPIE arguments.
+/// Then, smooshes them together with platform defaults, to decide whether
+/// this compile should be using PIC mode or not.
+/// Returns a tuple of ( backend.CodeGenOptions.PicLevel, IsPIE).
+pub fn getPICMode(d: *Driver, args: []const []const u8) !struct { backend.CodeGenOptions.PicLevel, bool } {
```
Vexu, line 907: Is there a reason why the PIC related args can't be parsed in `parseArgs` before being handed to this function like other args?

Organ1sm: `getPICMode` involves the following parameters：
- `-fPIC/-fpic/-fpie/-fPIE/-fno-pic/-fno-PIC/-fno-pie/-fno-PIE,  -->  confirm which parameter last appeared in the compilation arguments，be parsed in the `parseArgs` function is ok。

(The following parameters arocc currently does not support it, it will be supported in the future, so a complete implementation has been provided based on the implementation in LLVM ). It's mainly these parameters that led me to directly use `args` in `getPICMode`

- `-mkernel`, `-fapple-kext`, `-mno-abicalls`  --> check if certain parameters exist in 'args'.
- `-fropi/-fno-ropi/-frwpi/-fno-rwpi`，`-mdynamic-no-pic`  --> confirm which parameter last appeared in the compilation arguments. 
- `-mcmodel=` --> required obtain the value after the equal.

ehaas: I think what @Vexu is saying (and I agree) is that it's probably better to structure it so that we only have to make one pass over the args array. So this would involve bringing back the if statements and driver fields that I said to take out in my last review (sorry), and add new fields for `kernel`, `apple_kext`, `ropi`, `rwpi`, etc. If you're not entirely sure how to handle it let me know and I'll push a commit to this branch.

src/aro/Driver.zig
```diff
@@ -215,6 +232,7 @@ pub fn parseArgs(
     var comment_arg: []const u8 = "";
     var hosted: ?bool = null;
     var gnuc_version: []const u8 = "4.2.1"; // default value set by clang
+    var lastpic_pos: ?usize = null;
```
Vexu, line 235: Looks like this could be `var pic_arg: []const u8 = ""`.

src/aro/Driver.zig
```diff
@@ -892,3 +916,214 @@ fn exitWithCleanup(d: *Driver, code: u8) noreturn {
...
+        }
+    }
+
+    // The last argument relating to either PIC or PIE wins, and no
+    // other argument is used. If the last argument is any flavor of the
+    // '-fno-...' arguments, both PIC and PIE are disabled. Any PIE
+    // option implicitly enables PIC at the same level.
+    if (target.os.tag == .windows and
+        !target_util.isCygwinMinGW(target) and
+        // d.lastpic_pos != null and
```
Vexu, line 999: Leftover?

src/aro/Driver.zig
```diff
@@ -892,3 +916,214 @@ fn exitWithCleanup(d: *Driver, code: u8) noreturn {
...
+        StaticStringSet.initComptime(.{ .{"-fPIC"}, .{"-fpic"}, .{"-fpie"}, .{"-fPIE"} }).has(lastpic))
```
Vexu, line 1000: Now that `lastpic` arg is checked separately `std.ascii.eqlIgnoreCase` could work nicely for these.

src/aro/Driver.zig
```diff
@@ -892,3 +916,214 @@ fn exitWithCleanup(d: *Driver, code: u8) noreturn {
...
+    };
+    if (!forced) {
+        if (StaticStringSet.initComptime(.{ .{"-fPIC"}, .{"-fpic"}, .{"-fpie"}, .{"-fPIE"} }).has(lastpic)) {
+            pie = StaticStringSet.initComptime(.{ .{"-fpie"}, .{"-fPIE"} }).has(lastpic);
+            pic = pie or StaticStringSet.initComptime(.{ .{"-fpic"}, .{"-fPIC"} }).has(lastpic);
+            is_piclevel_two = StaticStringSet.initComptime(.{ .{"-fPIE"}, .{"-fPIC"} }).has(lastpic);
+        } else {
+            pic, pie = .{ false, false };
+            if (target_util.isPS(target)) {
+                const model_arg = getLastArgValue(args, "-mcmodel=");
```
Vexu, line 1023: Could you move the remaining args as well? The `getLastArg*` would not exist ideally.

Organ1sm: Yeah! I will clean up the rest tomorrow.

Vexu:
Thanks! I fixed some small translation errors for you to avoid another review cycle.

PR #772, opened 2024-09-20
Update the latest zig

Organ1sm:
Context: https://github.com/ziglang/zig/pull/21373, also lead to CI failed.(#771)

zig: rework fuzzing api: https://github.com/ziglang/zig/pull/21370


Vexu:
Thanks!

PR #782, opened 2024-10-05
Start adding frontend support for relocations

ehaas:
Add support for initializing pointers with, and doing static asserts with, offsets from global variables.

I'm very open to bikeshedding on names of things / code organization. I moved `StringInterner` to backend so that `Interner` could hold `StringId` values.

Open areas for enhancement:
   * Should also support static variables and not just globals.
   * Model relocation values with name + bigint instead of name + i64. Right now if the offset overflows we correctly warn but the printed diagnostic is incorrect since the value wraps around.

### DIFF
--- a/src/aro/Value.zig
+++ b/src/aro/Value.zig
@@ -4,6 +4,8 @@ const BigIntConst = std.math.big.int.Const;
 const BigIntMutable = std.math.big.int.Mutable;
 const backend = @import("backend");
 const Interner = backend.Interner;
+const StringInterner = backend.StringInterner;
+const StringId = StringInterner.StringId;
 const BigIntSpace = Interner.Tag.Int.BigIntSpace;
 const Compilation = @import("Compilation.zig");
 const Type = @import("Type.zig");
@@ -30,6 +32,10 @@ pub fn int(i: anytype, comp: *Compilation) !Value {
     } else {
         return intern(comp, .{ .int = .{ .i64 = i } });
     }
+}
+
+pub fn reloc(name: StringId, offset: i64, comp: *Compilation) !Value {
+    return intern(comp, .{ .global_var_offset = .{ .name = name, .offset = offset } });
 }
 
 pub fn ref(v: Value) Interner.Ref {
@@ -407,6 +413,7 @@ pub fn isZero(v: Value, comp: *const Compilation) bool {
             inline else => |data| return data[0] == 0.0 and data[1] == 0.0,
         },
         .bytes => return false,
+        .global_var_offset => return false,
         else => unreachable,
     }
 }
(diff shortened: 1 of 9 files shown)


src/aro/Value.zig
```diff
@@ -927,6 +1003,10 @@ pub fn compare(lhs: Value, op: std.math.CompareOperator, rhs: Value, comp: *cons
     return lhs_bigint.order(rhs_bigint).compare(op);
 }
 
+pub fn compare(lhs: Value, op: std.math.CompareOperator, rhs: Value, comp: *const Compilation) bool {
+    return lhs.compareExtra(op, rhs, comp).?;
```
Vexu, line 1007: This looks dangerous. Is it expected that other places won't reach this or is handling it just out of scope for this PR? Shouldn't be a blocker but a comment reminder wouldn't hurt.

Vexu (commented):
I don't love moving `StringInterner` to the backend since it isn't needed as much there. I'd prefer it to be something more abstract like:
```zig
pub const Ptr = struct {
    decl: DeclHandleTypeOfSomeSort,
    offset: Ref,
};
```

but I don't want to hold you back on my unfinished code maybe something like using `@intFromEnum(decl_name_id)` would work for now?

Looks good otherwise!

src/aro/Value.zig
ehaas: I don't think other paths could reach that, but I changed it to use a separate function for comparing pointers. In doing so I discovered a silly oversight with subtraction where I'd switched the operand order :)

ehaas:
Did MacOS CI always run on arm64 or did it used to be x86_64? The CI failure also happens on master so I'm trying to figure out if it was always lurking in aro or if something in zig changed that is causing it.

ehaas:
This patch fixes it, I think:

```diff
diff --git a/src/aro/Parser.zig b/src/aro/Parser.zig
index 21df662..69540ac 100644
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -5803,13 +5803,10 @@ pub const Result = struct {
                 if (try a.floatConversion(b, a_spec, b_spec, p, float_types[0])) return;
             }
             if (try a.floatConversion(b, a_spec, b_spec, p, float_types[1])) return;
-            if (p.comp.target.cTypeBitSize(.longdouble) == 80) {
+            if (p.comp.target.cTypeBitSize(.longdouble) == 80 or p.comp.target.cTypeBitSize(.longdouble) == 64) {
                 if (try a.floatConversion(b, a_spec, b_spec, p, float_types[0])) return;
             }
             if (try a.floatConversion(b, a_spec, b_spec, p, float_types[2])) return;
-            if (p.comp.target.cTypeBitSize(.longdouble) == 64) {
-                if (try a.floatConversion(b, a_spec, b_spec, p, float_types[0])) return;
-            }
             if (try a.floatConversion(b, a_spec, b_spec, p, float_types[3])) return;
             if (try a.floatConversion(b, a_spec, b_spec, p, float_types[4])) return;
             if (try a.floatConversion(b, a_spec, b_spec, p, float_types[5])) return;
```

Vexu:
> Did MacOS CI always run on arm64 or did it used to be x86_64? The CI failure also happens on master so I'm trying to figure out if it was always lurking in aro or if something in zig changed that is causing it.

The x86-64 MacOS runner has been deprecated and it was probably switched when I updated to actions/checkout v3 in 59e6078df83e45e2898d89689ce8914e81bcb4ab

PR #787, opened 2024-10-12
Preprocessor: remove deprecated PackedIntArray

ehaas:


Vexu (approved):
Looks good, nice to have those plain ints replaced with an enum!

PR #789, opened 2024-10-16
Pointer value improvements

ehaas:
Allow array-to-pointer decay to produce a pointer value
Handle address-of locals as well as globals
Rename "reloc" to pointer for consistency

### DIFF
--- a/test/cases/ast/_Float16.c
+++ b/test/cases/ast/_Float16.c
@@ -30,7 +30,7 @@ fn_def: 'fn (x: int, ...) void'
     builtin_call_expr: 'void'
      name: __builtin_va_start
      args:
-      implicit_cast: (array_to_pointer) 'va_list': '*d[1]struct __va_list_tag' (value: )
+      implicit_cast: (array_to_pointer) 'va_list': '*d[1]struct __va_list_tag' (value: &va)
         decl_ref_expr: 'va_list': '[1]struct __va_list_tag' lvalue
          name: va
       decl_ref_expr: 'int' lvalue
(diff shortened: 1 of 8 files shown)


test/cases/ast/_Float16.c
```diff
@@ -30,7 +30,7 @@ fn_def: 'fn (x: int, ...) void'
     builtin_call_expr: 'void'
      name: __builtin_va_start
      args:
-      implicit_cast: (array_to_pointer) 'va_list': '*d[1]struct __va_list_tag'
+      implicit_cast: (array_to_pointer) 'va_list': '*d[1]struct __va_list_tag' (value: )
```
Vexu, line 33: Would be nice if pointer values printed the offset (if non-zero).

Vexu (approved):

test/cases/ast/_Float16.c
ehaas: I meant to ask about this - should I just print the offset (without the name of the variable)?

Vexu: Printing the `NodeIndex` probably won't be too useful so I guess just the offset is good for now. Maybe add something to distinguish it from just ints.

ehaas: What if we change the return type of `Value.print` from `@TypeOf(w).Error!void` to something like `@TypeOf(w).Error!?NestedPrint` where `NestedPrint` is
```zig
const NestedPrint = union(enum) {
    pointer: struct {
        node: u32,
        offset: Value,
    },
};
```

PR #791, opened 2024-10-18
Compilation: use mtime for timestamp macro

### LINKED ISSUE #711: `__TIMESTAMP__` macro has incorrect value
We render the current time for `__TIMESTAMP__` but it is supposed to be "the date and time of the last modification of the current source file"

https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

Related: #239

danielsan901998:
resolve #711
Added a new argument to generateBuiltinMacros to pass an optional timestamp that is used in the case of not having a valid epoch passed as environment variable, to generate the timestamp from a path  a new function called generateBuiltinMacrosFromPath is created and is used in Driver for each source file instead of having the same builtin.
Changed generateDateAndTime to make the timestamp argument optional, since the stat can fail, and following GCC behavior it will expand timestamp to "??? ??? ?? ??:??:?? ????"

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -546,9 +546,6 @@ fn generateSystemDefines(comp: *Compilation, w: anytype) !void {
 
 /// Generate builtin macros trying to use mtime as timestamp
 pub fn generateBuiltinMacrosFromPath(comp: *Compilation, system_defines_mode: SystemDefinesMode, path: []const u8) !Source {
-    var buf = std.ArrayList(u8).init(comp.gpa);
-    defer buf.deinit();
-
     const stat = std.fs.cwd().statFile(path) catch return try generateBuiltinMacros(comp, system_defines_mode, null);
     const timestamp: i64 = @intCast(@divTrunc(stat.mtime, 1000000000));
     return try generateBuiltinMacros(comp, system_defines_mode, @intCast(std.math.clamp(timestamp, 0, max_timestamp)));
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -546,8 +546,8 @@ fn generateSystemDefines(comp: *Compilation, w: anytype) !void {
 
 /// Generate builtin macros trying to use mtime as timestamp
 pub fn generateBuiltinMacrosFromPath(comp: *Compilation, system_defines_mode: SystemDefinesMode, path: []const u8) !Source {
-    const stat = std.fs.cwd().statFile(path) catch return try generateBuiltinMacros(comp, system_defines_mode, null);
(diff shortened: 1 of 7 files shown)


src/aro/Compilation.zig
```diff
@@ -190,51 +190,45 @@ pub fn getSourceEpoch(self: *const Compilation, max: i64) !?i64 {
...
-        day_name,
-        month_name,
-        month_day.day_index + 1,
-        day_seconds.getHoursIntoDay(),
-        day_seconds.getMinutesIntoHour(),
-        day_seconds.getSecondsIntoMinute(),
-        year_day.year,
-    });
+fn generateDateAndTime(w: anytype, timestamp: ?u47) !void {
+    if (timestamp) |t| {
```
Vexu, line 194: Could you use `orelse` instead so that the entire function doesn't need to be indented?
```suggestion
fn generateDateAndTime(w: anytype, opt_timestamp: ?u47) !void {
    const timestamp = opt_timestamp orelse { ... };
```

src/aro/Compilation.zig
```diff
@@ -550,8 +544,18 @@ fn generateSystemDefines(comp: *Compilation, w: anytype) !void {
     }
 }
 
+/// Generate builtin macros trying to use mtime as timestamp
+pub fn generateBuiltinMacrosFromPath(comp: *Compilation, system_defines_mode: SystemDefinesMode, path: []const u8) !Source {
+    var buf = std.ArrayList(u8).init(comp.gpa);
+    defer buf.deinit();
+
```
Vexu, line 551:
```suggestion
```
These look accidental.

src/aro/Compilation.zig
```diff
@@ -587,9 +591,18 @@ pub fn generateBuiltinMacros(comp: *Compilation, system_defines_mode: SystemDefi
...
-    try generateDateAndTime(buf.writer(), timestamp);
+    const provided: ?i64 = comp.getSourceEpoch(max_timestamp) catch blk: {
+        try comp.addDiagnostic(.{
+            .tag = .invalid_source_epoch,
+            .loc = .{ .id = .unused, .byte_offset = 0, .line = 0 },
+        }, &.{});
+        break :blk null;
+    };
+    if (provided) |epoch| {
+        try generateDateAndTime(buf.writer(), @intCast(std.math.clamp(epoch, 0, max_timestamp)));
```
Vexu, line 602: This clamp could now be moved to `getSourceEpoch` since that's where it's validated.

Vexu (approved):
Looks good, just needs a few small adjustments.

danielsan901998:
Changes applied

Vexu:
Thanks!

PR #799, opened 2024-11-07
backend: basic x86_64 assembly codegen for some global vars

ehaas:
Some basic assembly codegen, tested with:
```c
/* main.c */
int printf(const char *fmt, ...);
extern char foo[];
extern int bar;
extern double baz;

int main(void)
{
    printf("%s\n%d\n%lf\n", foo, bar, baz);
    return 0;
}
```

```c
/* test.c */
char foo[] = "hello";
int bar = 42;
double baz = 3.14;

```

```sh-session
~$ gcc -c main.c
~$ arocc -O0 test.c main.o`
~$ ./a.out
hello
42
3.140000
```

I structured things the way I did so that `aro` (the frontend module) wouldn't depend on this assembly backend (as in, needing to import it so that it can call it from Driver). The way things are set up now, a do-nothing function can be passed as the final argument of `Driver.main`. But I'm not sure if that's the best way to achieve that.

My next goal is to get very basic function generation going and then I can revive the `EXPECTED_OUTPUT` tests.

### DIFF
--- a/src/backend/CodeGenOptions.zig
+++ b/src/backend/CodeGenOptions.zig
@@ -24,7 +24,6 @@ pub const PicLevel = enum(u8) {
 };
 
 pub const OptimizationLevel = enum {
-    unspecified,
     @"0",
     @"1",
     @"2",
@@ -60,6 +59,6 @@ pub const default: @This() = .{
     .data_sections = false,
     .pic_level = .none,
     .is_pie = false,
-    .optimization_level = .unspecified,
+    .optimization_level = .@"0",
     .debug = false,
 };
--- a/src/backend/CodeGenOptions.zig
+++ b/src/backend/CodeGenOptions.zig
@@ -1,3 +1,5 @@
+const std = @import("std");
+
 /// place uninitialized global variables in a common block
 common: bool,
 /// Place each function into its own section in the output file if the target supports arbitrary sections
@@ -7,6 +9,9 @@ data_sections: bool,
 pic_level: PicLevel,
 /// Generate position-independent code that can only be linked into executables
 is_pie: bool,
+optimization_level: OptimizationLevel,
+/// Generate debug information
+debug: bool,
 
 pub const PicLevel = enum(u8) {
     /// Do not generate position-independent code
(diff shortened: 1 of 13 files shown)


src/backend/CodeGenOptions.zig
```diff
@@ -18,10 +23,43 @@ pub const PicLevel = enum(u8) {
     two = 2,
 };
 
+pub const OptimizationLevel = enum {
+    unspecified,
```
Vexu, line 27: Not a fan of this existing, why not assign a default in `Driver`?

Vexu (approved):
> I structured things the way I did so that `aro` (the frontend module) wouldn't depend on this assembly backend (as in, needing to import it so that it can call it from Driver). The way things are set up now, a do-nothing function can be passed as the final argument of `Driver.main`. But I'm not sure if that's the best way to achieve that.

I'd be OK with placing the `genAsm` function in `backend.zig` and the Driver calling it from there unless you want to do it like this.

src/backend/CodeGenOptions.zig
ehaas: That's only there because the default is `-O0`, which is also how we're selecting this assembly backend. I was trying to disambiguate "no optimization selected" vs `-O0` but I could also just set a separate flag if `-O0` is seen.

ehaas:
> I'd be OK with placing the `genAsm` function in `backend.zig` and the Driver calling it from there unless you want to do it like this.

It looks like backend currently does not depend on frontend so I was trying to preserve that property - is it worth keeping it that way?

ehaas:
I changed things around a bit to use an optional function instead of `anytype`, and just have codegen return FatalError for unimplemented things instead of a new kind of error.

Vexu:
> It looks like backend currently does not depend on frontend so I was trying to preserve that property - is it worth keeping it that way?

The backend should be kept separate, I didn't notice this used Compilation for all the type details. I guess this could also be moved to the main module since it's not standalone like the backend module is.

Either way feel free to merge as is.

Issue #788, opened 2024-10-14
Add diagnostic for taking address of packed record member


ehaas:
```c
struct __attribute__((packed)) Foo {
    int x;
};
struct Foo foo;
int *p = &foo.x;
```

```
test.c:5:11: warning: taking address of packed member 'x' of class or structure 'Foo' may result in an unaligned pointer value [-Waddress-of-packed-member]
```

This check can go in the same place as the `addr_of_bitfield` check.

PR #800, opened 2024-11-10
Issue a warning on packed record member access

wrongnull:
Closes https://github.com/Vexu/arocc/issues/788

### DIFF
--- a/src/aro/Diagnostics.zig
+++ b/src/aro/Diagnostics.zig
@@ -21,10 +21,6 @@ pub const Message = struct {
 
     pub const Extra = union {
         str: []const u8,
-        record_member: struct {
-            record: []const u8,
-            member: []const u8,
-        },
         tok_id: struct {
             expected: Tree.Token.Id,
             actual: Tree.Token.Id,
@@ -397,7 +393,6 @@ pub fn renderMessage(comp: *Compilation, m: anytype, msg: Message) void {
     const prop = msg.tag.property();
     switch (prop.extra) {
         .str => printRt(m, prop.msg, .{"{s}"}, .{msg.extra.str}),
-        .record_member => printRt(m, prop.msg, .{ "{s}", "{s}" }, .{ msg.extra.record_member.member, msg.extra.record_member.record }),
         .tok_id => printRt(m, prop.msg, .{ "{s}", "{s}" }, .{
             msg.extra.tok_id.expected.symbol(),
             msg.extra.tok_id.actual.symbol(),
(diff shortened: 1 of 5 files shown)


src/aro/Parser.zig
```diff
@@ -7221,6 +7222,14 @@ fn unExpr(p: *Parser) Error!Result {
                 p.getNode(operand.node, .member_access_ptr_expr)) |member_node|
             {
                 if (tree.isBitfield(member_node)) try p.errTok(.addr_of_bitfield, tok);
+                const data = p.nodes.items(.data)[@intFromEnum(member_node)];
+                const lhs_ty = p.nodes.items(.ty)[@intFromEnum(data.member.lhs)];
+                if (lhs_ty.getAttribute(.@"packed") != null) {
```
ehaas, line 7227:
```suggestion
                if (lhs_ty.hasAttribute(.@"packed")) {
```

src/aro/Diagnostics.zig
```diff
@@ -21,6 +21,10 @@ pub const Message = struct {
 
     pub const Extra = union {
         str: []const u8,
+        record_member: struct {
```
ehaas, line 24: This makes the `Extra` union 32 bytes instead of 16 (on a 64-bit system). We should use something like what is done in `errDeprecated` or `valueChangedStr` - write the string to `p.strings`, dupe it, and then just use the `.str` Extra field

wrongnull: Done

src/aro/Parser.zig
```diff
@@ -7226,8 +7241,7 @@ fn unExpr(p: *Parser) Error!Result {
                 const lhs_ty = p.nodes.items(.ty)[@intFromEnum(data.member.lhs)];
                 if (lhs_ty.hasAttribute(.@"packed")) {
                     const record = lhs_ty.getRecord().?;
-                    const mapper = p.comp.string_interner.getSlowTypeMapper();
-                    const extra = Diagnostics.Message.Extra{ .record_member = .{ .record = mapper.lookup(record.name), .member = mapper.lookup(record.fields[data.member.index].name) } };
+                    const extra = Diagnostics.Message.Extra{ .str = try p.packedMemberAccessStr(record.name, record.fields[data.member.index].name) };
                     try p.errExtra(.packed_member_address, orig_tok_i, extra);
```
ehaas, line 7245: You can use the `errStr` function here with the string

wrongnull: +

ehaas (approved):

Vexu:
Thanks!

PR #807, opened 2024-12-28
Tree: represent nodes as tagged unions with a packed representation

Vexu:
Working with the current AST is annoying and difficult with similar constructs having multiple representations and dealing with multiarray list access and `@intFromEnum` and different data being stored in different data structures.

The goal of this branch is to simplify the usage by having a tagged union containing all the information of that node.


Vexu:
@ehaas do you think the Iterator will be necessary with these changes?

ehaas:
I haven't had nearly as much time to work on things as I'd like, so I haven't even really started using the Tree iterator yet. So if you want to remove it I say go for it and we can reassess later if it's needed.

PR #809, opened 2024-12-29
represent Type as a tagged union

Vexu:
Inspired by the similar changes to the AST (#807) this should finally make Type less painful to use.


Vexu:
Slowly getting there...
```sh
#test-integration:
282 passed; 53 failed.
#test-record:
max mem used = 8.49MiB
8270 passed; 657 failed (518 invalid targets).
```

Looks like there will be a nice bonus 0.4 MiB memory usage drop from https://github.com/Vexu/arocc/pull/817#issuecomment-2568434251

Vexu:
I'll allow some record layout tests to regress to get this merged, those can be fixed later when I recover from this.

This will almost definitely also regress untested code paths considering how long it took to get the tested ones working again, probably a good time to do some fuzz testing with this and #807 touching so many parts of the compiler.

ehaas:
Nice work!

Vexu:
Thanks!

PR #812, opened 2024-12-30
tests: update to new std.Target abis, and remove 32-bit CPU support for Apple platforms

ehaas:
I'll need to do some code archaeology to get the record-test-generator working again, so for now I'm just recording them as invalid tests so that they pass.

### DIFF
--- a/src/aro/target.zig
+++ b/src/aro/target.zig
@@ -958,42 +958,34 @@ pub fn isPICDefaultForced(target: std.Target) DefaultPIStatus {
 }
 
 test "alignment functions - smoke test" {
-    var target: std.Target = undefined;
-    const x86 = std.Target.Cpu.Arch.x86_64;
-    target.abi = .gnu;
-    target.cpu = std.Target.Cpu.baseline(x86, .{ .tag = .linux, .version_range = std.Target.Os.VersionRange.default(x86, .linux, target.abi) });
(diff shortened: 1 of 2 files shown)


src/aro/target.zig
```diff
@@ -960,9 +960,9 @@ pub fn isPICDefaultForced(target: std.Target) DefaultPIStatus {
 test "alignment functions - smoke test" {
     var target: std.Target = undefined;
     const x86 = std.Target.Cpu.Arch.x86_64;
-    target.cpu = std.Target.Cpu.baseline(x86, .{ .tag = .linux, .version_range = std.Target.Os.VersionRange.default(.linux, x86) });
-    target.os = std.Target.Os.Tag.defaultVersionRange(.linux, x86);
-    target.abi = std.Target.Abi.default(x86, target.os);
+    target.abi = .gnu;
+    target.cpu = std.Target.Cpu.baseline(x86, .{ .tag = .linux, .version_range = std.Target.Os.VersionRange.default(x86, .linux, target.abi) });
+    target.os = std.Target.Os.Tag.defaultVersionRange(.linux, x86, target.abi);
```
alexrp, line 965: Just as a general note, you might consider using `Cpu.generic()` instead of `Cpu.baseline()` since you're not really concerned with CPU features here. Similarly, OS version range seems irrelevant to what's being tested.

PR #817, opened 2025-01-02
simplify StringInterner

Vexu:


ehaas:
IIRC this approach uses more memory, but my linux machine is currently in a state of disrepair and I can't benchmark it right now. Even if it does use more memory though I like the simplified API so looks good to me!

ehaas (approved):

Vexu:
I figured something like that might have been the reason it was originally written like this.

Running `zig build test-record` as a benchmark:
```
simplify-string interner:
max mem used = 8.83MiB

master(053ea5b99afb554ad40e0ab53579976389ab65c7):
max mem used = 8.85MiB

before rewrite-ast (!):
max mem used = 11.28MiB
```

I also tried compiling sqlite3 but that crashes currently...

ehaas:
> I also tried compiling sqlite3 but that crashes currently...

There's an issue with `bitcast` in `computeOffsetExtra` as well as something else going on with Value, I can look into those.

PR #827, opened 2025-01-06
Tokenizer: use a def file for keywords

Vexu:
This should have better performance than StaticStringMap but I haven't tested that yet.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -155,6 +155,7 @@ pub fn build(b: *Build) !void {
             GenerateDef.create(b, .{ .name = "Builtins/Builtin.def", .needs_large_dafsa_node = true }),
             GenerateDef.create(b, .{ .name = "Attribute/names.def" }),
             GenerateDef.create(b, .{ .name = "Diagnostics/messages.def", .kind = .named }),
+            GenerateDef.create(b, .{ .name = "Tokenizer/keywords.def" }),
         },
     });
     const assembly_backend = b.addModule("assembly_backend", .{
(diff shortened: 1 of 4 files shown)


Vexu:
> This should have better performance than StaticStringMap but I haven't tested that yet.

```sh-session
$ hyperfine -i -w 1 -r 15 "def/bin/arocc compiler_rt.c -fsyntax-only -I ../zig/lib/" "master/bin/arocc compiler_rt.c -fsyntax-only -I ../zig/lib/"
Benchmark 1: def/bin/arocc compiler_rt.c -fsyntax-only -I ../zig/lib/
  Time (mean ± σ):     230.2 ms ±   3.7 ms    [User: 217.5 ms, System: 11.8 ms]
  Range (min … max):   225.5 ms … 241.3 ms    15 runs
 
  Warning: Ignoring non-zero exit code.
 
Benchmark 2: master/bin/arocc compiler_rt.c -fsyntax-only -I ../zig/lib/
  Time (mean ± σ):     218.4 ms ±   4.1 ms    [User: 204.9 ms, System: 12.5 ms]
  Range (min … max):   212.7 ms … 229.1 ms    15 runs
 
  Warning: Ignoring non-zero exit code.
 
Summary
  master/bin/arocc compiler_rt.c -fsyntax-only -I ../zig/lib/ ran
    1.05 ± 0.03 times faster than def/bin/arocc compiler_rt.c -fsyntax-only -I ../zig/lib/
```

I guess not.

PR #844, opened 2025-02-12
Parser: fix cond parsing after decl in for loop

### LINKED ISSUE #840: Decl in for loop init parsed incorrectly
Found in translate-c test cases:
```c
void foo() {
    for (int i = 2, b = 4; i + 2; i = 2) {
    }
}
```
```
$ clang a.c -fsyntax-only
$ arocc a.c -fsyntax-only
a.c:24:28: error: expected ';', found 'an identifier'
    for (int i = 2, b = 4; i + 2; i = 2) {
                           ^
a.c:24:35: error: use of undeclared identifier 'i'
    for (int i = 2, b = 4; i + 2; i = 2) {
                                  ^
2 errors generated.
```

danielsan901998:
Fix #840
There was a check to skip parsing the cond expression if the init expression was a declaration.
Removed the check and added the failed example of the issue to the test file case.

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -4859,7 +4859,6 @@ fn stmt(p: *Parser) Error!Node.Index {
 
         // for (init; cond
         const cond = cond: {
-            if (got_decl) break :cond null;
             const cond_tok = p.tok_i;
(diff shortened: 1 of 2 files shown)


Vexu:
Thanks!

Issue #848, opened 2025-04-14
Can't include `<stdlib.h>` on GitHub CI machines


Vexu:
`#include <stdlib.h>` works for me locally on `x86_64-linux-gnu` but on github CI machines it fails https://github.com/ziglang/translate-c/actions/runs/14454690218

On `ubuntu-latest`:
```
error: failed to translate /home/runner/.cache/zig/o/49e49dbfe1c33aa152c9de911716c2cf/tmp.c:
/usr/include/wchar.h:27:10: fatal error: 'bits/libc-header-start.h' not found
#include <bits/libc-header-start.h>
         ^
1 error generated.
```

On `macos-latest`:
```
/Users/runner/.cache/zig/o/4ecceef54001d7487d44250ca35fd06c/tmp.c:1:10: fatal error: 'stdlib.h' not found
#include <stdlib.h>
         ^
1 error generated.
```

On `windows-latest`:
```
 C:\Users\runneradmin\AppData\Local\zig\o\fbbcfdb119d3b79dd9890d1ba7d86b0c\tmp.c:2:10: fatal error: 'stdio.h' not found
#include <stdio.h>
         ^
1 error generated.
```

This is preventing enabling 68 `run` test cases in https://github.com/ziglang/translate-c

danielsan901998:
In my local environment(Ubuntu 24.04.2) does not works:
```
~/p/arocc> cat test.c
#include <stdlib.h>
~/p/arocc> ./zig-out/bin/arocc test.c
/usr/include/stdlib.h:26:10: fatal error: 'bits/libc-header-start.h' not found
#include <bits/libc-header-start.h>
         ^
1 error generated.
```
To avoid this error i have to pass -isystem /usr/include/x86_64-linux-gnu.
After checking it i saw that defineSystemIncludes generated usr/include/x86_64-linux-gnu instead of /usr/include/x86_64-linux-gnu.
It looks like it fails to add the triple because sysroot is empty and that cause the program to check  the path as relative to the current dir instead of absolute from /, i have tested this change in src/aro/toolchains/Linux.zig and it fixes it for me:
```diff
     if (getMultiarchTriple(target)) |triple| {
-        const joined = try std.fs.path.join(comp.gpa, &.{ sysroot, "usr", "include", triple });
+        const joined = try std.fs.path.join(comp.gpa, &.{ sysroot, "/usr", "include", triple });

```

Vexu:
[...]

PR #850, opened 2025-04-14
Support including libc on more targets

Vexu:
#848


src/aro/toolchains/Msvc.zig
```diff
@@ -0,0 +1,41 @@
...
+
+
+    if (tc.driver.nostdlibinc) return;
+
+    if (try addSystemIncludeFromEnv(comp, "INCLUDE") or
+        try addSystemIncludeFromEnv(comp, "EXTERNAL_INCLUDE")) return;
+}
+
+fn addSystemIncludeFromEnv(comp: *Compilation, env: []const u8) !bool {
+    const value = std.process.getEnvVarOwned(comp.gpa, env) catch return false;
```
ehaas, line 34: There's an `environment` field in Compilation that I created to hold all the environment variables that might be used (to remove the dependency on the `process` api outside of the top-level code that is initializing aro)

Vexu:
I don't want to work on this so I cherry picked 0b18adb to main and translate-c can use `std.zig.LibCInstallation` for other targets.

Issue #543, opened 2023-11-08
Convert test runners into build steps


Vexu:
For reference look at [`Cases.zig`](https://github.com/perillo/zig/blob/4504e03a18d75d0c9e4695250c807b3b4a953791/test/src/Cases.zig).

bingcicle:
would like to try to tackle this, if this is not being worked on already

Vexu:
Not that I know. I'd recommend starting with the integration tests and converting a few different types of tests first and then making a PR.

bingcicle:
@Vexu sorry, was away for a while and didn't have time to look at this - finally got a good look just now.

Have a quick question to understand how I can approach this: What is the end goal of this change? Is this to be able to eventually move towards being able to classify tests in categories (eg. errors, expected types, etc.)? Right now from my (very amateur) understanding the advantage of the approach in the zig repo is being able to have test manifests and to be able to test according to category (error, compile, translate-c). This contrasts with arocc, which can only do the integration tests wholesale now. I'm imagining the goal here is to move towards a more modular test runner?

Vexu:
No need to apologize, you're volunteering and any help is appreciated!

Being able to more explicitly classify tests is one goal. The tests are already classified but it's based on different flags and macros being set. Other benefits from the change will be the build systems built-in parallelism and caching in some cases.

PR #858, opened 2025-05-18
Preprocessor: make __DATE__, __TIME__, and __TIMESTAMP__ builtin macros

### LINKED ISSUE #852: Make `__TIMESTAMP__` etc. builtin macros like `__FILE__` so that the builtin source can be generated only once

ehaas:
Also makes __DATE__ and __TIME__ behave closer to the way they do in clang/gcc: They represent system time, not the current file's time.

GCC and clang use local time; we currently use UTC time.

Closes #852

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -58,7 +58,7 @@ pub const Environment = struct {
     /// TODO: not implemented yet
     c_include_path: ?[]const u8 = null,
 
-    /// UNIX timestamp to be used instead of the current date and time in the __DATE__ and __TIME__ macros
+    /// UNIX timestamp to be used instead of the current date and time in the __DATE__, __TIME__, and __TIMESTAMP__ macros
     source_date_epoch: ?[]const u8 = null,
 
     /// Load all of the environment variables using the std.process API. Do not use if using Aro as a shared library on Linux without libc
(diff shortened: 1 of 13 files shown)


ehaas:
The `invalid epoch.c` test doesn't quite fit with our test runner - I tried to match the clang behavior of treating it as an immediate fatal error but doing that means there's no way to get the value of the `EXPECTED_ERRORS` macro.

src/aro/Compilation.zig
```diff
@@ -614,15 +604,18 @@ fn generateSystemDefines(comp: *Compilation, w: anytype) !void {
     }
 }
 
-/// Generate builtin macros trying to use mtime as timestamp
-pub fn generateBuiltinMacrosFromPath(comp: *Compilation, system_defines_mode: SystemDefinesMode, path: []const u8) !Source {
-    const stat = comp.cwd.statFile(path) catch return try comp.generateBuiltinMacros(system_defines_mode, null);
-    const timestamp: i64 = @intCast(@divTrunc(stat.mtime, std.time.ns_per_s));
-    return try comp.generateBuiltinMacros(system_defines_mode, @intCast(std.math.clamp(timestamp, 0, max_timestamp)));
+pub fn loadSourceEpochFromEnvironment(comp: *Compilation) !void {
```
Vexu, line 607: Could you move this and the mtimes to the `Preprocessor`? Testing with Clang `SOURCE_DATE_EPOCH=abc gcc a.o` doesn't trigger the error so the behavior would match and it would avoid adding `FatalError` to `Compilation.initDefault()`.

Then again, it doesn't matter too much so use your own judgement.

Vexu (approved):
> The `invalid epoch.c` test doesn't quite fit with our test runner - I tried to match the clang behavior of treating it as an immediate fatal error but doing that means there's no way to get the value of the `EXPECTED_ERRORS` macro.

I guess it'll have to wait for #543.

PR #868, opened 2025-06-19
macOS frameworks

mlugg:
Adds CLI options `-F` and `-iframework` for macOS frameworks. These are similar to include directories, but with a slightly funky search strategy. See commit message for details.

Also adds `-idirafter`, because it's pretty straightforward -- this means Aro now supports all of the "include path" flags supported by the Zig build system.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -128,11 +128,17 @@ code_gen_options: CodeGenOptions = .default,
 environment: Environment = .{},
 sources: std.StringArrayHashMapUnmanaged(Source) = .{},
 /// Allocated into `gpa`, but keys are externally managed.
-include_dirs: std.ArrayListUnmanaged([]const u8) = .{},
+include_dirs: std.ArrayListUnmanaged([]const u8) = .empty,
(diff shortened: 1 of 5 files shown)


src/aro/Compilation.zig
```diff
@@ -1216,103 +1231,176 @@ fn addSourceFromPathExtra(comp: *Compilation, path: []const u8, kind: Source.Kin
...
+            .include_path = include_path,
+            .wait_for = null,
+        };
+
+        if (std.fs.path.isAbsolute(include_path)) {
+            switch (search_strat) {
+                .allow_same_dir, .only_search => {},
+                .only_search_after_dir => return null,
+            }
+            // TODO: classify absolute file as belonging to system includes or not?
```
Vexu, line 1288: Based on line directives and https://github.com/llvm/llvm-project/blob/f4db14229cd975822c41376afda9d56a29f9396c/clang/lib/Lex/HeaderSearch.cpp#L926 they are user includes.

mlugg:
Requested changes applied.

PR #883, opened 2025-08-16
Consume nonnull attribute

ehaas:
This consumes the `nonnull` attribute (but does not validate/apply it). This prevents it from being treated like an unknown attribute. Instead it issues a warning controlled by the "attribute-todo" Option. My thinking on this is that we'd still want to know if an attribute is not fully enabled, but allow disabling that particular warning. Silently dropping `nonnull` does not have any correctness implications, as far as I'm aware - it's only used to issue a warning if `NULL` is used directly as an argument, and perhaps to enable certain optimizations.

### DIFF
--- a/src/aro/Attribute.zig
+++ b/src/aro/Attribute.zig
@@ -345,6 +345,7 @@ fn diagnoseField(
 
 pub fn diagnose(attr: Tag, arguments: *Arguments, arg_idx: u32, res: Parser.Result, arg_start: TokenIndex, node: Tree.Node, p: *Parser) !bool {
     switch (attr) {
+        .nonnull => return false,
         inline else => |tag| {
             const decl = @typeInfo(attributes).@"struct".decls[@intFromEnum(tag)];
             const max_arg_count = comptime maxArgCount(tag);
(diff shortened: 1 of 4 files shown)


andrewrk:
In addition to nonnull, is there any attribute that could be used to help translate-c disambiguate between single-item pointers and many-item pointers?

ehaas:
I'm not aware of any attributes that would let you do that. One heuristic might be to check if a pointer is ever used with pointer arithmetic or an index expression. I think you'd have to propagate that data back up from the leaf nodes because C code might directly dereference a pointer with `*p` and then also pass it into a function which uses it as a multi-pointer, so that parent function would have to treat it as a multipointer and use `p[0]` to dereference.

Alternatively we could define our own attribute in aro like `__attribute__((zig_multiptr))` (a similar for single-pointer):
```c
#include <stddef.h>

#if defined __has_attribute
#  if __has_attribute (zig_multiptr)
#    define ATTR_MULTIPTR __attribute__ ((zig_multiptr))
#  endif
#endif
#if !defined ATTR_MULTIPTR
# define ATTR_MULTIPTR
#endif

int foo(ATTR_MULTIPTR int *p, size_t n);
```

I'm not sure how useful that would be / what the uptake would be for C *code*, but maybe people would do that for headers that export a public API for a C library.

xdBronch:
llvm people seem to be working on something related to the idea of pointer sizes, though im not quite sure i understand what exactly it is, ive only just skimmed it. might be interesting regardless https://clang.llvm.org/docs/BoundsSafety.html

PR #884, opened 2025-08-19
Compilation: do not use sendFile for reading source files

Vexu:


andrewrk:
are you working around a std lib bug?

Vexu:
No, I had to change this part because `sendFileAll` started requiring a non-empty buffer and since I can't benefit from its semantics using `appendRemaining` seemed better.

PR #892, opened 2025-09-07
Preprocessor: use placemarker instead of empty variadic arguments when pasting

### LINKED ISSUE #891: Preprocessor: pasting formed an invalid preprocessing token
Hi,

just tried to use the zig [translate-c](https://github.com/ziglang/translate-c) package on some Linux Kernel headers, which resulted in frequent `pasting formed <TOKEN>, an invalid preprocessing token` errors.

I was able to reproduce it on a specific macro used in the kernel, invoking `arocc` directly:

```c
#define __ASM_SIZE(inst, ...)   __ASM_SEL(inst##l##__VA_ARGS__, \
                                          inst##q##__VA_ARGS__)

__ASM_SIZE(bts)
```

Running only the arocc preprocessor (`arocc -E`):

```
src/kheaders.h:1:43: error: pasting formed 'btsl,', an invalid preprocessing token
#define __ASM_SIZE(inst, ...)   __ASM_SEL(inst##l##__VA_ARGS__, \ 
                                          ^
src/kheaders.h:4:12: note: expanded from here
__ASM_SIZE(bts)
           ^
<scratch space>:1:1: note: expanded from here
btsl
^
src/kheaders.h:2:43: error: pasting formed 'btsq)', an invalid preprocessing token
                                          inst##q##__VA_ARGS__)
                                          ^
src/kheaders.h:4:12: note: expanded from here
__ASM_SIZE(bts)
           ^
<scratch space>:3:1: note: expanded from here
btsq
^
2 errors generated.
```

Output from gcc (`gcc -E`) shows correctly:

```
__ASM_SEL(btsl, btsq)
```

Using latest `arocc` (commit `7c75fed474a9f940c7a292821148b8fd775f56cc`), and zig `0.16.0-dev.65+ca2e17e0a`

Not sure if this is just not (yet) supported or a bug, but just wanted to report it.

ehaas:
Closes #891


src/aro/Preprocessor.zig
```diff
@@ -1703,7 +1703,10 @@ fn expandFuncMacro(
                     else
                         &[1]TokenWithExpansionLocs{tokFromRaw(raw_next)},
                     .macro_param, .macro_param_no_expand => getPasteArgs(args.items[raw_next.end]),
-                    .keyword_va_args => variable_arguments.items,
+                    .keyword_va_args => if (variable_arguments.items.len == 0) blk: {
+                        try pp.err(raw_next, .no_argument_variadic_macro, .{});
+                        break :blk &[1]TokenWithExpansionLocs{.{ .id = .placemarker, .loc = .{ .id = .generated } }};
```
ehaas, line 1708: Is this pointer valid when using it with a `break` statement like this? I'm not totally sure what the lifetime of a stack object like this is if it's inside an if statement.

Vexu: This should be fine because it is a comptime known value but generally I don't think this would be valid.

PR #895, opened 2025-10-05
Use separator-agnostic `#include_next` path comparions on Windows, add test cases

kcbanner:
This resolves https://github.com/ziglang/translate-c/issues/190, see that link for details on the paths that weren't matching in my case. 

- Add test cases for #include_next that test including a file of the same name, and in system include folders
- Fixed test runner using the 0th token for each iteration in the `EXPECTED_ERRORS` logic
- When matching #include_next dirs, use separator-aware comparisons on windows](https://github.com/Vexu/arocc/pull/895/commits/141aa4f0784ab17d9800ec709c2d824bbe26182d)

Something I noticed with the include_next test cases: I expected to see an error like this:

`"same_name_system.h:1:2: warning: #include_next is a language extension [-Wgnu-include-next]" \`

One is present for the non-system include. I'm unsure if this is a bug or just an intentional difference in how system includes are handled.


kcbanner:
Hm, although this solution doesn't handle this case (from https://github.com/ziglang/translate-c/issues/190#issuecomment-3368612481):

`warning: vadefs.h -> include_dir (skip): C:\cygwin64\home\kcbanner\kit\zig\build-stage3-release\lib\zig\compiler/aro/include`

This mixed-separator path is provided by the TranslateC build step logic I believe. Perhaps all include dirs should be resolved?

kcbanner:
> In other parts dealing with include paths backslashes are replaced with regular slashes when ms extensions are enabled which should work here as well.

Ah thanks, I see that now.

However, in the case of the target being `*-windows-gnu`, I think the emulated compiler would still be set to `.gcc` (via `target_util.systemCompiler`), so I'm not sure if we could rely on `ms_extensions` for doing path conversions in that case. 

In this specific case, the `\compiler/aro/include` mixup is a bug in the zig compiler itself, in how it generates that path - I'll fix that in separate PR there.

kcbanner:
I reworked the checks to do a separator-agnostic comparison on Windows. This approach doesn't require allocating, and will handle mixed-separator paths.

kcbanner:
> Yes but the wait_for path is coming from a Source whose path would have been normalized in check.

Ah, I missed this before - the `ms_extensions` flag causing the path to change is exactly the problem.

I just pushed a commit with a test case illustrating the problem. 

> Could this be fixed by always converting backslashes when the host OS is Windows?

To be clear, do you mean always converting *to* backslashes? This doesn't work as it sets `wait_for` to `E:\dev\arocc\test\cases\include\\mixed` but then it's trying to match `E:\dev\arocc\test\cases\include//mixed`. In general we can't normalize to one kind of separator as the path in `comp.include_dirs` can be mixed, due to user input - any normalization will break plain string compares. 

So the options I see:
- Remove the `ms_extensions` behaviour of normalizing paths - I'm probably missing context on why that happens in the first place. This would leave the path with the original mixed separators so the `std.mem.eql` comparison matches.
- Use a separator-agnostic compare (ie. the 2nd commit here)

kcbanner:
Ah, that makes sense!

I removed my earlier changes and pushed a commit that skips the normalization if the compiler is running on Windows.

PR #898, opened 2025-10-08
Preprocessor: Refactor handling of builtin functions

ehaas:
Previously, each preprocessor builtin would generate its own unique token which would then get expanded appropriately. Refactor the code so that instead, there are only two special macro tokens: one for object macros and one for macro functions. Those get expanded by looking at the builtin that is being expanded (instead of encoding that info into the token itself). By doing things this way, we can add new builtins without adding more token types.

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -64,7 +64,7 @@ const IfContext = struct {
 };
 
 pub const Macro = struct {
-    const BuiltinKind = union(enum) {
+    const Builtin = union(enum) {
         const Object = enum {
             file,
             line,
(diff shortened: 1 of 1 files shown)


ehaas:
My plan is to use this to implement `__is_target_arch`, `__is_target_vendor`, `__is_target_os`,  `__is_target_environment`, `__building_module`, etc.

Also: even before this change I noticed that the unit and integration tests seem significantly slower; is it possible this is a Zig regression? At first I thought it might be related to the new aarch64 backend, but the record tests remain fast. It seems like only a single thread is being used for the unit tests (this is on an M4 Max mac):

```
ehaas@Mac arocc % zig version
0.16.0-dev.624+606c7bcc8
ehaas@Mac arocc % time zig build test-unit
zig build test-unit  7.65s user 0.09s system 99% cpu 7.753 total
ehaas@Mac arocc % time zig build test-record
max mem used = 10.02MiB
7967 passed; 1042 skipped (518 invalid targets).
zig build test-record  63.40s user 0.35s system 1403% cpu 4.540 total
```

ehaas:
Looks like it's not a threading issue; maybe just the backend? I'm not sure when the switch happened

```
ehaas@Mac arocc % ~/source/zig/build/stage3/bin/zig version
0.16.0-dev.195+ac42eaaad
ehaas@Mac arocc % time ~/source/zig/build/stage3/bin/zig build test-unit
~/source/zig/build/stage3/bin/zig build test-unit  0.11s user 0.06s system 93% cpu 0.184 total
ehaas@Mac arocc % time ~/source/zig/build/stage3/bin/zig build test-record
max mem used = 9.39MiB
7967 passed; 1042 skipped (518 invalid targets).
~/source/zig/build/stage3/bin/zig build test-record  52.02s user 0.39s system 1424% cpu 3.680 total
```

PR #910, opened 2025-10-22
Driver: parse target with aliases, vendor and subarch

Vexu:


src/aro/target.zig
```diff
@@ -425,40 +425,102 @@ fn toLower(src: []const u8, dest: []u8) ?[]const u8 {
...
+            .{ "win32", .windows },
+            .{ "xros", .visionos },
+        }).get(lower) orelse return null;
+}
+
+pub fn isOs(target: *const std.Target, query: []const u8) bool {
+    const parsed = parseOs(query) orelse return false;
 
-    if (query_os.isDarwin()) {
+    if (target.os.tag.isDarwin()) {
```
ehaas: should be `parsed.isDarwin()`, right?

PR #924, opened 2025-11-13
new test runner

### LINKED ISSUE #543: Convert test runners into build steps

Vexu:
~~#543  for integration tests. Converting the record runner to call Aro as a child process should be easy as well.~~

Closes #543

### DIFF
--- a/test/runner.zig
+++ b/test/runner.zig
@@ -137,6 +137,18 @@ fn runCaseExtra(
         child.env_map = &env_map;
     }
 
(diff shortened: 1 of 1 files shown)


Vexu:
@ehaas any thoughts on this before I convert all the old cases?

PR #934, opened 2025-11-24
Attribute: handle additional ways calling convention can be applied

### LINKED ISSUE #879: some attributes dont work on `typedef`s
```
a.c:1:24: warning: attribute 'sysv_abi' ignored on types [-Wignored-attributes]
typedef __attribute__((sysv_abi)) void (*foo)(void);
                       ^
```

ehaas:
Closes #879


Vexu (approved):
> Looking at this more...do we need to move the attribute into the function type instead of the pointer?

Yes but that might be easier to do when rewriting the attribute application system.

PR #936, opened 2025-11-25
LangOpts: do not emulate by default

Vexu:


steeve:
We need for translate-c/aro to process the files exactly as clang did when the libs were compiled. We go so far as sandboxing glibc and linux headers (see https://github.com/cerisier/toolchains_llvm_bootstrapped)

So in order to do that that we save the system defines of the clang invocation and include them in the translate-c invocation file and use `-undef -nostdlibinc -nobuiltininc`. This works well.

ehaas (approved):

PR #937, opened 2025-11-27
Attribute: start adding bounds safety attributes

ehaas:
This starts adding clang-style bounds safety attributes (hidden behind the `-fexperimental-bounds-safety` flag). Right now it's just single pointers, which are the easiest (parser disallowed arithmetic and indexing unless the index is statically known to be zero).

If we go down the route of adding these, I think a big question is how to model pointers - do we add additional pointer types, or do we just use attributes? Some of the pointer types (bidi_indexable and indexable) have different ABI than the others, and there are various rules about which pointer types are allowed to implicitly convert to others.

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -3558,7 +3558,6 @@ fn declarator(
 
         const pointer_qt = try p.comp.type_store.put(p.comp.gpa, .{ .pointer = .{
             .child = d.qt,
-            .decayed = null,
         } });
         d.qt = try builder.finishQuals(pointer_qt);
     }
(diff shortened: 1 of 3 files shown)


src/aro/TypeStore.zig
```diff
@@ -1642,6 +1646,7 @@ pub const Type = union(enum) {
     pub const Pointer = struct {
         child: QualType,
         decayed: ?QualType,
+        bounds: Attribute.PointerBounds,
```
Vexu: Could you make this default to `.c`? Probably could be applied to `decayed` as well since these are both rarely set to anything else.

src/aro/Parser.zig
```diff
@@ -8679,6 +8687,9 @@ fn unExpr(p: *Parser) Error!?Result {
                 try p.err(tok, .not_assignable, .{});
                 return error.ParsingFailed;
             }
+            if (operand.qt.get(p.comp, .pointer)) |pointer| {
+                try p.checkPtrArithmeticAllowed(pointer, p.tok_i, operand.node);
```
Vexu: `get` doesn't handle `.invalid`, instead move this to `checkPtrArithmeticAllowed` behind a `.isInvalid()` check.

Issue #945, opened 2025-12-02
Figure out why some integration tests block on Windows


Vexu:
Three test cases are skipped on Windows because they cause the test runner to get stuck.

7f9e777cc5351c40a90187aa0ac212e54fed666c

PR #952, opened 2025-12-15
Preprocessor: track hideset for generated tokens

### LINKED ISSUE #950: Infinite loop in preprocessor
I have preserved the helpful comment from which this reduced case was generated: https://github.com/xiph/opus-tools/blob/master/src/speex_resampler.h

```c
/********* WARNING: MENTAL SANITY ENDS HERE *************/

#define CAT_PREFIX2(a,b) a ## b
#define CAT_PREFIX(a,b) CAT_PREFIX2(a, b)

#define speex_resampler_init CAT_PREFIX(speex,_resampler_init)
#define speex_resampler_init_frac CAT_PREFIX(speex,_resampler_init_frac)

void speex_resampler_init(void);
void speex_resampler_init_frac(void);
```

`arocc -E test.h` will get stuck in an infinite loop during macro expansion

ehaas:
Fixes #950

### DIFF
--- a/test/runner.zig
+++ b/test/runner.zig
@@ -141,7 +141,8 @@ fn runCaseExtra(
         // I have no idea why these block on windows
         if (mem.eql(u8, case.name, "assignment") or
             mem.eql(u8, case.name, "attribute errors") or
-            mem.eql(u8, case.name, "initializers"))
+            mem.eql(u8, case.name, "initializers") or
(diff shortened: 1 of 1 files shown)


Vexu (approved):
Looks good, I can check which tests are getting stuck on Windows a bit later (#945).

PR #958, opened 2026-01-05
update to Zig 0.16.0-dev.1912+0cbaaa5eb

Vexu:


Vexu:
It is in comparison to master branch so Zig `0.16.0-dev.1634+b27bdd5af`, I suspect it to be caused by errors on my part.



> A few months ago I started seeing a dramatic slowdown in the unit tests due to debug allocator changes on MacOS.

Didn't that get fixed by not capturing stack traces by default?

ehaas:
Hm, disregard my comment. I thought I had experienced it when working on #957 but I'm not actually seeing any problems now. Zig version `0.16.0-dev.1634+b27bdd5af`

Issue #941, opened 2025-11-30
Rename `TypeStore` to `TypeMap`


Vexu:
I think store is too vague and that map better describes what the type actually does.

PR #967, opened 2026-02-22
Rename 'TypeStore' to 'TypeMap'

wrongnull:
Closes https://github.com/Vexu/arocc/issues/941


Vexu:
Variables and function arguments should also be changed (`ts` -> `tm`).

Issue #825, opened 2025-01-06
Ensure Clang C extension compatibility


Vexu:
https://clang.llvm.org/docs/LanguageExtensions.html

* [ ] `__BASE_FILE__`
* [ ] `__FILE_NAME__`
* [ ] `__INCLUDE_LEVEL__`
* [ ] Matrix types
* [ ] `__bf16`
* [ ] blocks?
* [ ] Verify that all currently supported Clang extensions have warnings

PR #969, opened 2026-02-26
feat: Block type support

dotcarmen:
work towards closing #825 

This is step 1 towards supporting Blocks in arocc: parsing block types. A 2nd PR will be opened soon to support parsing block literals and finally, a 3rd PR will be opened to support lowering block literals and invocations.

Note that some limited support for the Blocks extension in GCC [is underway but undocumented](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78352), so I chose to only support blocks when clang is emulated or without emulation and either:
- target macOS >= 10.6 (as noted [in the Clang implementation specification](https://clang.llvm.org/docs/Block-ABI-Apple.html#id1)) or other darwin targets
  - i'm not entirely sure how to handle when `target.os.isAtLeast(.macos, ...)` returns `null`, so support is disabled in this case
  - the specification doesn't say when other Darwin platforms got Blocks support, but i suspect they all have always supported Blocks. the only one I feel less secure about is iOS, which might have gone a release or 2 without support, but I'm fuzzy on timing
- `-fblocks` is an argument

This implementation treats blocks as a pointer, just like function pointers.

The type _should_ only successfully parse when the pointee is a function type, I did my best to follow the logic in `Parser.zig` but I'd appreciate any holes in the logic being pointed out

Note that the `__BLOCKS__` preprocessor symbol doesn't currently get defined in this PR, since the implementation is still incomplete


src/aro/Parser.zig
```diff
@@ -2022,6 +2022,9 @@ fn initDeclarator(p: *Parser, decl_spec: *DeclSpec, attr_buf_top: usize, decl_no
     try p.attributeSpecifierExtra(init_d.d.name);
 
     switch (init_d.d.declarator_type) {
+        .block => {
+            // TODO: Blocks validation?
```
dotcarmen: any idea on what/how i should validate here? not immediately clear to me...

ehaas:
> CI is failing on ubuntu and windows due to AST output differences on these platforms. I'm not sure how to fix this without modifying the test suite to only partially compare the output using `std.mem.indexOf`?

Since blocks are only supported on MacOS you can put `--target=aarch64-macos` or similar in the args for the block test; without that it will use the host as the target.

src/aro/Parser/Diagnostic.zig
```diff
@@ -2540,3 +2540,24 @@ pub const multiple_bounds_annotations: Diagnostic = .{
...
+    .fmt = "Blocks are a Clang extension",
+    .opt = .@"blocks-extension",
+    .kind = .off,
+    .extension = true,
+};
+
+pub const blocks_not_enabled: Diagnostic = .{
+    .fmt = "Blocks are not enabled",
+    .opt = .@"blocks-extension",
+    .kind = .warning,
```
Vexu: Only the extension diagnostic should have `opt` and `extension` set and this should be an error.

Vexu (changes requested):

src/aro/Parser/Diagnostic.zig
Vexu: I meant that only `blocks_are_clang_extension` should have them set.

dotcarmen:
... oh no, apparently a new master _just_ dropped `dev.2905+5d71e3051` (it's still not even on `ziglang.org` lol), and it updates the formatter. gonna open a separate PR that fixes formatting first, then i'll rebase ~+reformat~ (only `Compilation.zig` is affected) this branch

test/cases/block types.c
```diff
@@ -0,0 +1,53 @@
...
+
+/** manifest:
+syntax
+args = -fblocks -Wpedantic --target=aarch64-macos
+
+block types.c:4:26: error: block pointer to non-function type is invalid
+block types.c:10:9: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
+block types.c:10:15: error: block pointer to non-function type is invalid
+block types.c:12:14: error: expected identifier or '('
+block types.c:23:11: error: initializing 'int' from incompatible type 'bar' (aka 'int (^)(int)')
```
dotcarmen: clang has this output:
```
test/cases/block types.c:23:7: error: initializing 'int' with an expression of incompatible type 'bar' (aka 'aFunc ^')
```

note that the position is different, and the displayed type of `bar` in the clang output isn't desugared. i get matching type desugaring behavior with:

<details>
<summary>patch: don't desugar when printing block.func</summary>

```diff
diff --git a/src/aro/TypeStore.zig b/src/aro/TypeStore.zig
index 572d8fa96c..f663182189 100644
--- a/src/aro/TypeStore.zig
+++ b/src/aro/TypeStore.zig
@@ -1285,7 +1285,7 @@
             .block => |block| {
                 if (!block.func.is(comp, .func)) unreachable;
 
-                const simple = try block.func.printPrologue(comp, desugar, w);
+                const simple = try block.func.printPrologue(comp, false, w);
                 try w.writeAll(if (simple) " ^" else "(^");
                 if (qt.@"const") try w.writeAll("const");
                 if (qt.@"volatile") {
@@ -1416,7 +1416,7 @@
                 }
                 continue :loop pointer.child.type(comp);
             },
-            .block => |block| continue :loop block.func.type(comp),
+            .block => |block| try block.func.printEpilogue(comp, false, w),
             .func => |func| {
                 try w.writeByte('(');
                 for (func.params, 0..) |param, i| {
```
</details>

test/cases/block types.c
```diff
@@ -0,0 +1,53 @@
...
+block types.c:24:10: error: block pointer to non-function type is invalid
+block types.c:27:23: error: initializing 'MyInt (^)(MyInt)' from incompatible type 'MyInt' (aka 'int')
```
dotcarmen: clang has this output:
```
test/cases/block types.c:27:11: error: invalid block pointer conversion initializing 'MyInt (^)(MyInt)' with an expression of type 'MyInt' (aka 'int')
```

clang seems to have block-specific alternatives to `incompatible_{assign,init}` diagnostics? that seems rather unnecessary, but i can add them if desired

src/aro/TypeStore.zig
```diff
@@ -1279,6 +1282,23 @@ pub const QualType = packed struct(u32) {
                 if (simple) try w.writeByte(' ');
                 return false;
             },
+            .block => |block| {
+                if (!block.func.is(comp, .func)) unreachable;
+
+                const simple = try block.func.printPrologue(comp, desugar, w);
+                try w.writeAll(if (simple) " ^" else "(^");
```
dotcarmen, line 1292: i think i misunderstood `simple` before, i believe this is more correct though i don't know how to test the simple path...

test/cases/block types.c
Vexu: I think there are other places where we desugar more than Clang does, I think this is good as is.

test/cases/block types.c
Vexu: I'll trust your judgement on this.

src/aro/TypeStore.zig
Vexu: Works fine:
```c
typedef int foo(int);
foo ^b = 1; // error: initializing 'foo ^' from incompatible type 'int'
```

src/aro/Parser.zig
```diff
@@ -7389,6 +7420,10 @@ pub const Result = struct {
             if (dest_unqual.eql(res.qt, p.comp)) {
                 return; // ok
             }
+        } else if (dest_unqual.is(p.comp, .block)) {
```
Vexu: You also need to handle blocks in `Result.castType`:
```c
typedef int foo(int);
typedef int bar(float);
void qux(void) {
    foo ^a;
    bar ^b;
    a = (foo ^)b;
}
```

test/cases/block types.c
```diff
@@ -27,12 +27,28 @@ void my_func_accepting_block(double (*afunc)(int), double (^ablock)(int)) {
...
+  d = e;
 
   afunc = ablock;
   ablock = afunc;
 
+  void *anyptr;
+
   int (^f)(int) = a;
   a = f;
+  f = anyptr;
```
dotcarmen, line 44: clang doesn't emit a diagnostic on this - i personally feel like this should be a warning, but maybe it shouldn't be included in any 'group' options? (ie `-Wpedantic` even though it _would_ be a pedantic diagnostic)

same with `f = 0xdeadbeef`, though `anyptr = f` makes sense as an implicit coercion

dotcarmen: should've justified why it should be a diagnostic

[the spec](https://clang.llvm.org/docs/BlockLanguageSpec.html#block-variable-declarations) says block pointers can be _casted_, but doesn't say anything about coercion/implicit casts:

```
A Block reference may be cast to a pointer of arbitrary type and vice versa.
```

block pointers are already treated differently from other pointer types when coercing, and IMO the rest of the spec indicates enough special behavior that it doesn't make sense to at least warn about implicit casting to a pointer type or vice-versa

src/aro/Parser.zig
```diff
@@ -3535,14 +3540,22 @@ const Declarator = struct {
...
                 return .normal;
             },
+            .block => |block_ty| {
+                const func_qt = block_ty.func;
+                if (func_qt._index == .declarator_combine) return .normal;
+                if (!func_qt.is(p.comp, .func)) {
+                    try p.err(source_tok, .block_to_non_function, .{});
+                    return .nested_invalid;
+                }
+                return validateExtra(p, func_qt, source_tok);
```
Vexu:
```suggestion
                const child_res = try validateExtra(p, func_qt, source_tok);
                if (child_res != .normal) return child_res;

                if (!func_qt.is(p.comp, .func)) {
                    try p.err(source_tok, .block_to_non_function, .{});
                    return .nested_invalid;
                }
                return .normal;
```

This removes a redundant error and matches the other branches.
```
__auto_type ^a;
```

It's not the same error as with Clang but I think it is good enough.

Vexu (approved):
Sorry for taking so long with reviewing this.

dotcarmen:
thanks @Vexu :) no worries about the delay - i appreciate all the reviews you put in, especially all the failing/unexpected tests you gave. i'll be more rigorous in my testing in the future to limit review fatigue

i'll rebase #971 onto latest main now

PR #971, opened 2026-02-26
Parser: block literals and __block variables

dotcarmen:
~~blocked on/~~ part 2 of #969, wip towards closing #825

This PR implements parsing of block literals according to the [language extension specification by Clang](https://clang.llvm.org/docs/BlockLanguageSpec.html). Caveat: while the spec document [specifies](https://clang.llvm.org/docs/BlockLanguageSpec.html#block-literal-expressions) that the parameter list may only be omitted when the return type is omitted, Clang actually [compiles programs that omit the return type but not the parameter list](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:1,endLineNumber:10,positionColumn:1,positionLineNumber:10,selectionStartColumn:1,selectionStartLineNumber:10,startColumn:1,startLineNumber:10),source:'typedef+int+(%5EMyBlock)(int)%3B%0A%0Aint+main(void)+%7B%0A++++__block+int+count+%3D+0%3B%0A++++int+(%5Eblock)(int)+%3D+%5E+(int+arg)+%7B%0A++++++++count+%2B%3D+1+%2B+arg%3B%0A++++++++return+count%3B%0A++++%7D%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:cclang2110,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-fblocks',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+21.1.0+(Editor+%231)',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'zig+0.14.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+21.1.0+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) (excuse the linker error). The behavior here matches that behavior

Additionally, this change makes no attempt to track `__block` variable references within block bodies, since that's not necessary for the parse to complete, and that extra information would not be used anywhere yet. in the future, when implementing lowering support for blocks, this information tracking can be added


dotcarmen:
actually i'm gonna convert this to a draft, i realized i forgot to add the `__block` storage class - i'll get to that soon

dotcarmen:
i just realized aro doesn't support lowering to `aarch64-macos` - since macos is currently the only platform that works for testing blocks, i think this is enough to mark Blocks support as "completed" with the caveat that blocks don't get lowered, and an additional ticket should be opened for lowering literals and calls.

test/cases/block literals.c
```diff
@@ -0,0 +1,53 @@
...
+/** manifest:
+syntax
+args = -fblocks -Wpedantic --target=aarch64-macos
+
+block literals.c:9:0: error: variable is not assignable (missing __block type specifier)
+block literals.c:19:3: error: use of undeclared identifier 'res'
+block literals.c:20:3: warning: expression result unused [-Wunused-value]
+block literals.c:21:3: error: use of undeclared identifier 'arg'
+block literals.c:27:3: error: use of undeclared identifier 'anotherArg'
+block literals.c:37:13: error: __block attribute not allowed, only allowed on local variables
```
dotcarmen: currently stuck on figuring out how to resolve this error... i need to somehow determine whether the lval of an `assignExpr` has the `__block` storage spec.

dotcarmen (commented):
btw i'm tracking all changes for this PR only in the last commit until #969 lands :)

dotcarmen:
CI is blocked on #988, `zig build test` passes locally

src/aro/Preprocessor.zig
```diff
@@ -384,6 +386,10 @@ pub fn addBuiltinMacros(pp: *Preprocessor) !void {
         try pp.addBuiltinMacro("__FILE_NAME__", .{ .obj = .file_basename });
         try pp.addBuiltinMacro("__INCLUDE_LEVEL__", .{ .obj = .include_level });
     }
+
+    if (pp.comp.langopts.blocks) {
+        try pp.addBuiltinMacro("__BLOCKS__", .defined);
```
Vexu: This should be defined in `Compilation.generateSystemDefines`.

src/aro/Parser.zig
```diff
@@ -5571,37 +5585,62 @@ fn compoundStmt(p: *Parser, is_fn_body: bool, stmt_expr_state: ?*StmtExprState)
...
-            if (last_noreturn == .no) switch (ret_qt.base(p.comp).type) {
-                .void => {},
-                .func, .array => {}, // Invalid, error reported elsewhere
-                else => {
-                    const func_name = p.tokSlice(p.func.name);
-                    const interned_name = try p.comp.internString(func_name);
+        if (p.block) |*block| {
+            if (block.return_type) |ret_qt| {
+                if (last_noreturn != .yes) {
+                    if (last_noreturn == .no) switch (ret_qt.type(p.comp)) {
```
Vexu: This should use `ret_qt.base(p.comp).type`.

Vexu (commented):
Looks pretty good.

dotcarmen:
sorry for the delay, latest commit addresses the review comments :)

Vexu:
I just noticed that `__block` isn't a keyword in Clang but rather a builtin macro that expands to an attribute definition:

```c
#define __block __attribute__((__blocks__(byref)))
```

I'm not sure how much of a difference that makes in practice but we should probably match it.

dotcarmen:
oh! good to know - thank you for pointing that out. that'll take a bit more work to rework in, i'll do that soon

also worth noting, i haven't completed my own re-review since the last PR landed, so I might have another push before i do that

src/aro/Attribute/names.def
```diff
@@ -501,3 +501,9 @@ neon_vector_type
 neon_polyvector_type
     .tag = .neon_polyvector_type
     .gnu = true
+
+blocks
+    .tag = .blocks
+    .gnu = true
+    .c23 = true
+    .clang = true
```
dotcarmen: added clang to Attribute to fully support the namespaced attribute https://clang.llvm.org/docs/AttributeReference.html#blocks
however, it seems a bit non-ideal that "namespaces" aren't separate from "styles" - right now i don't think `[[gnu::blocks]]` should be explicitly supported, though `__attribute__((blocks))` should be supported since that's what clang uses.

i'm working on an issue for this i'll link to soon

src/aro/Parser/Diagnostic.zig
```diff
@@ -2591,3 +2591,59 @@ pub const c23_attribute: Diagnostic = .{
     .suppress_version = .c23,
     .extension = true,
 };
+
+pub const invalid_block_noreturn: Diagnostic = .{
+    .fmt = "block declared 'noreturn' should not return",
+    .kind = .@"error",
+    .opt = .@"invalid-noreturn",
+};
```
dotcarmen, line 2378: still needs a test

src/aro/Parser/Diagnostic.zig
```diff
@@ -2591,3 +2591,59 @@ pub const c23_attribute: Diagnostic = .{
...
+
+pub const invalid_block_noreturn_block_defined_here: Diagnostic = .{
+    .fmt = "block defined here",
+    .kind = .note,
+    .opt = .@"invalid-noreturn",
+};
```
dotcarmen, line 2384: still needs a test

src/aro/Parser/Diagnostic.zig
```diff
@@ -2591,3 +2591,59 @@ pub const c23_attribute: Diagnostic = .{
...
+
+pub const non_void_block_does_not_return: Diagnostic = .{
+    .fmt = "non-void block does not return a value",
+    .kind = .warning,
+    .opt = .@"return-type",
+};
```
dotcarmen, line 2390: still needs a test

src/aro/Parser/Diagnostic.zig
```diff
@@ -2591,3 +2591,59 @@ pub const c23_attribute: Diagnostic = .{
...
+
+pub const void_block_returns_value: Diagnostic = .{
+    .fmt = "void block should not return a value",
+    .kind = .warning,
+    .opt = .@"return-type",
+};
```
dotcarmen, line 2396: still needs a test

src/aro/Parser/Diagnostic.zig
```diff
@@ -2591,3 +2591,59 @@ pub const c23_attribute: Diagnostic = .{
...
+
+pub const block_should_return: Diagnostic = .{
+    .fmt = "non-void block should return a value",
+    .kind = .warning,
+    .opt = .@"return-type",
+};
```
dotcarmen, line 2402: still needs a test

src/aro/Parser/Diagnostic.zig
```diff
@@ -2591,3 +2591,59 @@ pub const c23_attribute: Diagnostic = .{
...
+pub const block_return_block_defined_here: Diagnostic = .{
+    .fmt = "block defined here",
+    .kind = .note,
+    .opt = .@"return-type",
+};
+
+pub const missing_block_literal_body: Diagnostic = .{
+    .fmt = "missing body for block literal",
+    .kind = .@"fatal error",
+};
```
dotcarmen, line 2413: still needs a test

dotcarmen (commented):
still working on review but here are some TODOs

src/aro/Attribute/names.def
dotcarmen: ... i'm just gonna open a PR (too many words, better discussed next to code). my idea is to redefine the above as:
```zig
blocks
    .tag = .blocks
    .gnu = true
    .c23 = false
    .namespaces = .{ .clang }
```

it can be implemented in a way to avoid a breaking change, though imo we should eventually make a breaking change for the other attrs

src/aro/Attribute.zig
```diff
@@ -23,18 +23,21 @@ pub const Syntax = enum {
...
     declspec,
     gnu,
+    clang,
 
     pub fn toSyntax(kind: Kind) Syntax {
         return switch (kind) {
             .c23 => .c23,
             .declspec => .declspec,
             .gnu => .gnu,
+            .clang => .clang,
```
dotcarmen: see https://github.com/Vexu/arocc/pull/971/changes#r3373876086

i can rebase this PR onto the new one if desired

src/aro/Attribute.zig
```diff
@@ -795,6 +808,8 @@ pub fn fromString(kind: Kind, namespace: ?[]const u8, name: []const u8) ?Tag {
         const normalized_ns = normalize(ns);
         if (mem.eql(u8, normalized_ns, "gnu")) {
             break :blk .gnu;
+        } else if (mem.eql(u8, normalized_ns, "clang")) {
+            break :blk .clang;
```
dotcarmen: see https://github.com/Vexu/arocc/pull/971/changes#r3373876086

dotcarmen:
realized that i hadn't implemented support for attributes on block literals.

note to self: c23 attributes aren't specified for block literals, not supported by clang

src/aro/Attribute/names.def
Vexu: I've been working on some improvements to attributes and can add proper support for Clang attributes.

dotcarmen: do you mind pushing that work? I can rebase onto your branch, otherwise i was planning on working on my design today

no worries if you think this work can be done in parallel and we can fix whichever PR lands 2nd :)

Vexu: I haven't finished yet, I meant that you shouldn't put too much effort into adding proper support for the `clang::` namespace since I'm about to change how that works.

src/aro/Parser/Diagnostic.zig
dotcarmen: doesn't need a test (it's a fatal error and breaks the ast output checking)

src/aro/Attribute.zig
```diff
@@ -1164,11 +1176,92 @@ pub fn applyFunctionAttributes(p: *Parser, qt: QualType, attr_buf_start: usize)
         .weakref,
         .zero_call_used_regs,
         => |t| try p.err(tok, .attribute_todo, .{ @tagName(t), "functions" }),
+        // TODO: clang doesn't emit a dx here, but behavior for `__block <function>` is unspecified
```
dotcarmen: i wonder if this might apply `__block` to the variable when initialized with this

eg for `auto` or `id` it could be eg `id b = ^__block {};` instantiates `b` as a `__block` variable

src/aro/Attribute.zig
```diff
@@ -1164,11 +1176,92 @@ pub fn applyFunctionAttributes(p: *Parser, qt: QualType, attr_buf_start: usize)
...
+        => try applyGnuAttrCallingConvention(attr, p, tok, qt),
+
+        .nonnull => {
+            const block_ty = qt.get(p.comp, .block).?;
+            const func_ty = block_ty.func.get(p.comp, .func).?;
+            try applyNonnull(p, attr, tok, func_ty);
+        },
+
+        .sentinel,
+        .blocks, // TODO: clang doesn't emit any diagnostic, but it's not clear what this does
```
dotcarmen: same as https://github.com/Vexu/arocc/pull/971/changes#r3408893811

src/aro/Parser.zig
```diff
@@ -174,6 +174,14 @@ func: struct {
     pretty_ident: ?Result = null,
 } = .{},
 
+/// null if not in Block literal
+block: ?struct {
+    caret: TokenIndex,
+    return_type: ?QualType = null,
```
dotcarmen: i feel like this should be `TypeStore.Builder` for the full block type, especially now that the attributes might be queried.

this is difficult because the return type of the Block is the only part that may be unknown until the body is finished parsing. `TypeStore.Builder` doesn't have a utility for that, and such a utility would really only be used for Blocks (though this might be used for eg c++ `auto` return type)

src/aro/Attribute.zig
dotcarmen: looked into this by checking `-dump-ast` (no useful info besides "my hypothesis above is incorrect") and by glancing at the LLVM discourse a bit, couldn't find anything.

this is probably only a detail when lowering, the question is - should i remove this TODO? if clang truly doesn't do anything here, then i think an attribute ignored warning would be nice when not `--emulate=clang`. but if it does somehow affect compilation, then such a warning would be frivolous

src/aro/Parser.zig
```diff
@@ -9971,6 +10051,92 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+    }
+
+    const attrs = p.attr_buf.items(.attr)[attr_buf_top..];
+    for (attrs) |attr| switch (attr.tag) {
+        .noreturn => p.block.?.noreturn = true,
+        else => {},
+    };
+
+    var maybe_ret_builder: TypeStore.Builder = .{ .parser = p };
+    if (try p.typeSpec(&maybe_ret_builder)) {
```
Vexu: `typeSpec` isn't enough here:
```c
__auto_type a = ^int (*(void))(int a) {
    return 0;
};
```

src/aro/Parser.zig
```diff
@@ -9971,6 +10051,92 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+    const node = try p.addNode(.{ .block_literal = .{
+        .caret = caret,
+        .body = body,
+        .qt = block_type,
+    } });
+
+    return .{
+        .node = node,
+        .qt = block_type,
+        .val = try .block(@intFromEnum(node), p.comp),
```
Vexu: Does a `block` value need to exist? The only place where you handle it you explicitly ignore it.

src/aro/Parser.zig
Vexu: I'm not sure I understand your concern, where and what attributes might be getting queried?

dotcarmen: Right now only `noreturn` is queried, but it's possible more might be queried in the future

Vexu: I don't think the return type can have any attributes that would affect the parsing of the block body.

src/aro/Parser.zig
```diff
@@ -5879,27 +5912,54 @@ fn returnStmt(p: *Parser) Error!?Node.Index {
...
+                    try some.coerce(p, ret_qt, e_tok, .ret);
+                    try some.saveValue(p);
+                }
+            } else if (!ret_qt.is(p.comp, .void)) {
+                try p.err(ret_tok, .block_should_return, .{});
+                try p.err(block.caret, .block_return_block_defined_here, .{});
             }
         } else {
-            try some.coerce(p, ret_qt, e_tok, .ret);
+            block.return_type = if (ret_expr) |*some| some.qt else .void;
```
Vexu: Missing lvalue conversion:
```c
__auto_type a = ^(int a) {
    int arr[2];
    return arr;
};
```

```suggestion
        } else if (ret_expr) |*some| {
            try some.lvalConversion(p, e_tok);
            block.return_type = some.qt;
        } else {
            block.return_type = .void
```

src/aro/Attribute/Wip.zig
```diff
@@ -155,7 +155,7 @@ fn arg(wip: *Wip, comptime WantedBase: type) !?WantedBase {
         if (Wanted == Tree.Node.DeclRef and node == .decl_ref_expr) {
             return node.decl_ref_expr;
         }
-        if (@typeInfo(@TypeOf(Wanted)) == .@"enum" and Wanted.opts.enum_kind == .identifier and node == .identifier_arg) {
+        if (@typeInfo(Wanted) == .@"enum" and Wanted.opts.enum_kind == .identifier and node == .identifier_arg) {
```
dotcarmen, line 158: bug fix for this previously-unused code path

dotcarmen (commented):
still have to finish handling the attributes that i handled before the attributes rework, and finish addressing previous review comments. attributes rework was nice - only had to fix 1 bug in an unused codepath. 

the biggest change from the merge: i updated `p.block` to use a Node with a QualType, both of which get updated as more block literal information is parsed. this allows `p.tree.attr_map.hasAttribute` to work as expected.

test/cases/block literals.c
```diff
@@ -85,18 +94,28 @@ void noreturn_attribute() {
...
+  b = ^__attribute((warning)) {};
 }
 
 /** manifest:
 syntax
-args = -fblocks -Wpedantic --target=aarch64-macos -std=c23
+args = -fblocks -Wpedantic --target=aarch64-macos -std=c23 -Wno-gnu-auto-type
 
-block literals.c:23:17: error: __block attribute not allowed, only allowed on local variables
+block literals.c:23:17: warning: 'blocks' attribute only applies to local variables
```
dotcarmen: some of the attribute diagnostics emitted by clang are errors, not warnings - but currently only warnings are emitted. i'm assuming this is fine?

src/aro/Parser.zig
dotcarmen: technically it doesn't need to exist. i just created it
1. to satisfy the return type of `fn primaryExpr`
2. because it'd be needed when lowering

i can remove it and leave the value as `undefined`?

src/aro/Parser.zig
dotcarmen: clang warns about returning a pointer to a value on the stack. i couldn't find a similar diagnostic, so i'm assuming such detection is unimplemented.

src/aro/Parser.zig
```diff
@@ -10207,6 +10294,100 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+
+    const node = try p.addNode(.{ .empty_decl = .{ .semicolon = caret } });
+    p.block = .{ .node = .pack(node), .scope_depth = p.syms.active_len };
+
+    try p.syms.pushScope(p);
+    defer p.syms.popScope();
+
+    const attr_state = p.wip_attrs.state(true);
+    defer p.wip_attrs.restore(attr_state);
+    try p.attributeSpecifier();
```
dotcarmen: should keyword attributes be handled as part of `attributeSpecifier`?

dotcarmen:
btw i'm working on splitting out my merge commit to better show the updated blocks-related handling. i'll try to push it tonight

src/aro/Attribute/Map.zig
```diff
@@ -159,6 +160,12 @@ pub fn put(map: *Map, gpa: mem.Allocator, attribute: Attribute) !Ref {
             try map.extra.append(gpa, attribute.tok);
             try map.putString(gpa, name);
         },
+        .blocks => |byref| {
+            repr.tag = .blocks;
+            repr.data = @intCast(map.extra.items.len);
+            try map.extra.append(gpa, attribute.tok);
+            try map.extra.append(gpa, @intFromEnum(byref));
```
Vexu: You are using 32 bits to store 0 bits of information, use a comptime assert instead.

src/aro/Attribute/Wip.zig
```diff
@@ -357,11 +368,17 @@ pub fn applyDeclAttrsExtra(
 
     if (prev_decl.unpack()) |prev| try wip.inherit(p, prev);
 
+    const is_block_literal = wip.blockLiteralTarget();
     for (wip.attrs.items[wip.top..]) |*attr| {
         if (attr.used_as_type_attr) continue;
         wip.current.attr = attr;
         wip.current.arg_i = 0;
 
+        if (is_block_literal and attr.syntax != .gnu) {
```
Vexu: I preferred when you added this error in the parser.

src/aro/Attribute/Wip.zig
```diff
@@ -390,7 +407,7 @@ pub fn applyDeclAttrsExtra(
                     try wip.add(.hot);
                 },
                 .@"const", .__const => {
-                    if (try wip.argCount(0)) continue;
+                    if (try wip.argCount(0)) return;
```
Vexu: Why did you change these?

test/cases/block literals.c
Vexu: You don't have to use `checkTarget` if you want to add a more specific diagnostic.

src/aro/Parser.zig
Vexu: I don't see why you would need it when lowering since the same information is available from the node. You can just leave the value empty for non compile time known values.

src/aro/Parser.zig
Vexu: As in `__cdecl`? Those are available via `msTypeAttribute`.

src/aro/Attribute/Wip.zig
```diff
@@ -1323,3 +1333,18 @@ fn applyFalltrhough(wip: *Wip) !void {
...
     try wip.add(.fallthrough);
 }
+
+fn applyBlocks(wip: *Wip) !void {
+    if (try wip.argCount(1)) return;
+    // clang doesn't emit a dx, but it's not clear what this does on block literals
+    if (try wip.checkTarget(&.{ .local_variable, .block_literal })) return;
+
+    const attr = wip.current.attr;
+    switch (attr.syntax) {
```
Vexu: Unnecessary.

src/aro/Parser.zig
```diff
@@ -5716,37 +5724,64 @@ fn compoundStmt(p: *Parser, is_fn_body: bool, stmt_expr_state: ?*StmtExprState)
...
-                    const func_name = p.tokSlice(p.func.name);
-                    const interned_name = try p.comp.internString(func_name);
+        if (p.block.node.unpack()) |block_node| {
+            const block = block_node.get(&p.tree).block_literal;
+            var func_ty = block.qt.get(p.comp, .func).?;
+            const ret_qt = func_ty.return_type;
+            if (ret_qt.isBlockLiteralAutoReturn()) {
+                func_ty.return_type = .void;
+                try block.qt.set(p.comp, .{ .func = func_ty });
+            } else if (last_noreturn != .yes) {
```
Vexu: Missing `isInvalid()` check.

src/aro/Parser.zig
```diff
@@ -10207,6 +10294,100 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+
+    const attr_state = p.wip_attrs.state(true);
+    defer p.wip_attrs.restore(attr_state);
+    try p.attributeSpecifier();
+
+    var qt: QualType = undefined;
+    if (try p.typeName()) |typename| {
+        if (typename.is(p.comp, .func)) {
+            qt = typename;
+        } else {
```
Vexu: Needs a check for array types:
```c
^int [2] {};
```

src/aro/Parser.zig
```diff
@@ -10207,6 +10294,100 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+    try p.tree.setNode(.{ .block_literal = .{
+        .caret = caret,
+        .qt = qt,
+        .body = undefined,
+    } }, @intFromEnum(node));
+    try p.wip_attrs.applyDeclAttrsExtra(p, node, qt, .null);
+
+    var compound_stmt_state: StmtExprState = .{};
+    const body = try p.compoundStmt(true, &compound_stmt_state) orelse {
+        try p.err(p.tok_i, .missing_block_literal_body, .{});
```
Vexu, line 10475: This should not be fatal, the note is never added because of it.

src/aro/Parser.zig
```diff
@@ -10207,6 +10294,100 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+        try p.err(caret, .block_defined_here, .{});
+        unreachable;
+    };
+
+    // could have been updated by returnStmt
```
Vexu: Looks like `compoundStmt` now always updates the return type so this could be an assertion.

src/aro/Parser.zig
```diff
@@ -5716,37 +5724,64 @@ fn compoundStmt(p: *Parser, is_fn_body: bool, stmt_expr_state: ?*StmtExprState)
...
+                    else => {
+                        try p.err(r_brace, .non_void_block_does_not_return, .{});
+                        try p.err(block.caret, .block_return_block_defined_here, .{});
+                    },
+                };
+
+                const implicit_ret = try p.addNode(.{ .return_stmt = .{
+                    .return_tok = r_brace,
+                    .return_qt = ret_qt,
+                    .operand = .{ .implicit = false },
```
Vexu, line 6038:
```suggestion
                    .operand = .{ .implicit = true },
```

Vexu (commented):
Here's a diff to prevent a crash when using `-fblocks` without `-fsyntax-only`:
```diff
diff --git a/src/aro/CodeGen.zig b/src/aro/CodeGen.zig
index 082f41cf..4b322540 100644
--- a/src/aro/CodeGen.zig
+++ b/src/aro/CodeGen.zig
@@ -876,6 +876,7 @@ fn genExpr(c: *CodeGen, node_index: Node.Index) Error!Ir.Ref {
         .sizeof_expr,
         .builtin_va_arg_pack,
         .builtin_va_arg_pack_len,
+        .block_literal,
         => return c.fail("TODO CodeGen.genExpr {t}\n", .{node}),
         .codegen_diagnostic => |diagnostic| {
             try c.comp.diagnostics.addWithLocation(c.comp, .{
```

src/aro/Attribute/Wip.zig
dotcarmen: oops, i originally was adding attributes to blocks via a `applyBlockLiteralAttributes` fn but realized i could make it work here, and i guess i forgot to fully revert the changes...

src/aro/Parser.zig
dotcarmen: as in bare `noinline`, `_Noreturn`, etc - the keywords handled directly in `fn declSpec`

Vexu: Those shouldn't be allowed here.

dotcarmen:
^ still working on addressing review comments

src/aro/Parser.zig
dotcarmen: clang's error messages state eg `Type name does not allow function specifier to be specified` (using keyword `inline`) and `Type name does not allow storage class to be specified` (using keyword `auto`). right now, using these keywords results in a panic.

i've changed these messages to `block literal does not allow {function specifier,storage class} to be specified` since the errors don't otherwise exist. these messages are better anyways...

src/aro/Attribute/Map.zig
```diff
@@ -172,6 +173,12 @@ pub fn put(map: *Map, gpa: mem.Allocator, attribute: Attribute) !Ref {
             try map.extra.append(gpa, attribute.tok);
             try map.putString(gpa, name);
         },
+        .blocks => |byref| {
+            repr.tag = .blocks;
+            repr.data = @intCast(map.extra.items.len);
+            try map.extra.append(gpa, attribute.tok);
```
Vexu: You can leave the token in `repr.data`.
```suggestion
```

src/aro/features.zig
```diff
@@ -87,6 +87,7 @@ pub fn hasExtension(comp: *Compilation, ext_raw: []const u8) bool {
         .gnu_asm_goto_with_outputs = true,
         .matrix_types = false, // TODO
         .matrix_types_scalar_division = false, // TODO
+        .blocks = comp.langopts.blocks,
```
Vexu:
```suggestion
```
#1075

test/cases/block literals.c
```diff
@@ -0,0 +1,191 @@
...
+};
+
+void blocks_attribute() {
+  __attribute__((blocks(byref))) int attr_cap;
+  __attribute__((blocks(byvalue))) int yet_another_attr_cap;
+  __attribute__((blocks)) int so_many_attr_caps;
+  [[blocks(byref)]] int c23_attr_cap; // TODO: this should emit an unknown attribute warning
+  [[clang::blocks(byref)]] int more_attr_caps;
+  [[clang::blocks(byval)]] int this_is_probably_enough_attr_capping;
+  [[gnu::blocks(byref)]] int gnu_blocks_not_supported_tho; // TODO: this should emit an unknown attribute warning
```
Vexu: Outdated comments.
```suggestion
  [[blocks(byref)]] int c23_attr_cap;
  [[clang::blocks(byref)]] int more_attr_caps;
  [[clang::blocks(byval)]] int this_is_probably_enough_attr_capping;
  [[gnu::blocks(byref)]] int gnu_blocks_not_supported_tho;
```

src/aro/Parser.zig
```diff
@@ -5908,37 +5916,64 @@ fn compoundStmt(p: *Parser, is_fn_body: bool, stmt_expr_state: ?*StmtExprState)
...
+                    },
+                };
 
-                    if (interned_name == p.string_ids.main_id) {
-                        if (ret_qt.get(p.comp, .int)) |int_ty| {
-                            if (int_ty == .int) return_zero = true;
+                const implicit_ret = try p.addNode(.{ .return_stmt = .{
+                    .return_tok = r_brace,
+                    .return_qt = ret_qt,
+                    .operand = .{ .implicit = true },
```
Vexu: Oh sorry you had this right originally, I thought it was a flag instead of being an union member.
```suggestion
                    .operand = .{ .implicit = false },
```

src/aro/Parser.zig
```diff
@@ -10394,6 +10481,141 @@ fn checkArrayBounds(p: *Parser, index: Result, array: Result, tok: TokenIndex) !
...
+        switch (p.tok_ids[p.tok_i]) {
+            .keyword_inline,
+            .keyword_inline1,
+            .keyword_inline2,
+            .keyword_noreturn,
+            .keyword_auto_type,
+            .keyword_forceinline,
+            .keyword_forceinline2,
+            => {
+                try p.err(bad_attr_tok_i, .block_does_not_allow_specifier, .{"function specifier"});
```
Vexu: I added this warning in 275027941f17760e6516f5a93fe51780c2b38ec3. You'll probably want to inline `typeNameExtra` here to handle decl attributes in the type properly:

```c
int (^a)(int) = ^int __attribute__((const)) (int a){
    return a;
};
```

dotcarmen:
uh, wat. different results for block literals depending on os?

test/cases/block literals.c
```diff
@@ -0,0 +1,235 @@
...
+void other_attributes() {
+  void *b = ^__attribute((ohno)) {};
+  b = ^__attribute((deprecated)) {};
+  b = ^__attribute((unavailable)) {};
+  b = ^__attribute((pure)) {};
+  b = ^__attribute((const("var"))) {}; // comment out to show more clang errors below (aro is better)
+  b = ^__attribute((visibility("bad"))) {}; // comment out to show more clang errors below (aro is better)
+  b = ^__attribute((aligned(3))) {}; // comment out to show more clang errors below (aro is better)
+  b = ^__attribute((format)) {}; // comment out to show more clang errors below (aro is better)
+  b = ^__attribute((nonnull(0))) (void *a) {}; // comment out to show more clang errors below (aro is better)
```
dotcarmen, line 127: > uh, wat. different results for block literals depending on os?

on aarch64-linux it's apparently due to this line (i tested in a vm on my macbook, the diagnostics pass with this line commented out and the diagnostic removed from the expected output)

i'm still not sure why this behavior is host-platform-specific...

i'm really busy over the next couple of weeks, but when i have time i'll debug this more

Issue #970, opened 2026-02-26
objective-c support


dotcarmen:
Unlike C++, Objective-C is a strict superset of C that is supported by GCC and Clang. Since these features are purely additive, I believe it'd be beneficial to support Objective-C in Aro - though if it's determined out of scope, I'd start and maintain a fork of Aro that supports Objective-C. My primary motivation for this feature is to use Aro to generate bindings to Objective-C frameworks using my [objc bindings library](https://codeberg.org/dotcarmen/zig-objc).

I propose the following plan for Objective-C support in arocc. To match Clang behavior, Objective-C support will automatically be enabled on Darwin targets, and will be supported on other platforms with the `-ObjC` flag.

Note that lots of Objective-C usage relies on the Blocks C extension, which I've already started working on. I believe it makes sense to automatically toggle Blocks support when Objective-C support is enabled, but I'm not yet sure if that happens on non-Darwin platforms.

Initially, Objective-C support will be focused on Darwin targets by Clang emulation. Once this list of features is complete, support for GCC emulation (which uses GNU Step and may have some variation in compiler output) and potentially non-emulated output can be tackled in parallel with bug squashing and other missing features.

This list is mostly "vibes-based" based on my research into the Objective-C runtime for my [bindings library](https://codeberg.org/dotcarmen/zig-objc). The ordering is a guideline, not a rule, for implementation, and some features may be un/intentionally excluded. However, this set of features _should_ cover the majority of Objective-C programs.

- [ ] `__OBJC__`, `OBJC_TYPES_DEFINED`, `__OBJC_BOOL_IS_BOOL`, `OBJC_API_VERSION`, `OBJC_NO_GC`
    - `OBJC_TYPES_DEFINED` requires builtin definitions for `Class`, `Method`, `SEL`, `Object`, and more.
    - `__OBJC_BOOL_IS_BOOL` indicates using the builtin `bool` type for the objective-c `BOOL` type.
        - `__has_feature(objc_bool)` isn't required, since the fallback is fine for arocc usage.
    - `OBJC_API_VERSION` initially will always be set to `2`
        - Support for the legacy api version 0 can be added in the future, but since it's long obsolete, I don't see much value in supporting it immediately.
    - since objc ARC has been the standard for a long time, I also don't see value in immediately supporting objc garbage collection
- [ ] `#import` ~~and `@import`~~
    - `#import` is the same as `#include` but eliminates the need for include guards
[...]

PR #974, opened 2026-03-05
Toolchain: Darwin toolchain for SDK paths

dotcarmen:
as i'm working on objective-c support (per #970), i've realized that support is somewhat blocked on fetching the SDK path for the target platform - specifically, for objc headers, which i believe are implicitly included in Objective-C files.

this PR adds a new `toolchains/Darwin.zig` file for Darwin-specific target behavior. Right now, this is only used to:
1. add the SDK library path (`$SDKROOT/usr/lib`)
2. add the SDK include path (`$SDKROOT/usr/include`)

In a future PR, this file will be expanded to add the SDK framework path (`$SDKROOT/System/Library/Frameworks`)

### DIFF
--- a/include/stdint.h
+++ b/include/stdint.h
@@ -7,6 +7,29 @@
 
 # include_next <stdint.h>
 
+# if defined(__APPLE__) && __ARO_EMULATE__ == __ARO_EMULATE_NO__
+
+#  if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
+#   define INTMAX_WIDTH   __INTMAX_WIDTH__
+#   define UINTMAX_WIDTH  __UINTMAX_WIDTH__
+#   define INTPTR_WIDTH   __INTPTR_WIDTH__
+#   define UINTPTR_WIDTH  __UINTPTR_WIDTH__
+#   define PTRDIFF_WIDTH  __PTRDIFF_WIDTH__
+#   define SIZE_WIDTH     __SIZE_WIDTH__
+#   define WCHAR_WIDTH    __WCHAR_WIDTH__
+
+#   define UINT8_WIDTH    8
+#   define UINT16_WIDTH   16
+#   define UINT32_WIDTH   32
+#   define UINT64_WIDTH   64
+#   define INT8_WIDTH     UINT8_WIDTH
+#   define INT16_WIDTH    UINT16_WIDTH
+#   define INT32_WIDTH    UINT32_WIDTH
+#   define INT64_WIDTH    UINT64_WIDTH
+#  endif
+
+# endif
+
 #else
 
 #define __stdint_int_c_cat(X, Y) X ## Y
--- a/include/stdint.h
+++ b/include/stdint.h
@@ -7,7 +7,7 @@
 
 # include_next <stdint.h>
 
-# if defined(__APPLE__) && __ARO_EMULATE__ == __ARO_EMULATE_NO__
+# if defined(__APPLE__)
 
 #  if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
 #   define INTMAX_WIDTH   __INTMAX_WIDTH__
--- a/src/aro/Toolchain.zig
+++ b/src/aro/Toolchain.zig
@@ -529,6 +529,11 @@ pub fn defineSystemIncludes(tc: *Toolchain) !void {
     };
 }
 
+pub fn addSystemIncludeDirJoined(tc: *const Toolchain, components: []const []const u8) !void {
+    const path = try std.fs.path.join(tc.driver.comp.arena, components);
+    try tc.addSystemIncludeDir(path);
+}
+
 pub fn addSystemIncludeDir(tc: *const Toolchain, path: []const u8) !void {
     const d = tc.driver;
     _ = try d.includes.append(d.comp.gpa, .{ .kind = .system, .path = try d.comp.arena.dupe(u8, path) });
--- a/src/aro/Toolchain.zig
+++ b/src/aro/Toolchain.zig
@@ -8,6 +8,7 @@ const Driver = @import("Driver.zig");
 const Multilib = @import("Driver/Multilib.zig");
 const Target = @import("Target.zig");
 const Linux = @import("toolchains/Linux.zig");
+const Darwin = @import("toolchains/Darwin.zig");
 
 pub const PathList = std.ArrayList([]const u8);
 
@@ -36,13 +37,14 @@ pub const UnwindLibKind = enum {
 
 const Inner = union(enum) {
     uninitialized,
+    darwin: Darwin,
     linux: Linux,
     unknown: void,
 
     fn deinit(self: *Inner, allocator: mem.Allocator) void {
         switch (self.*) {
             .linux => |*linux| linux.deinit(allocator),
-            .uninitialized, .unknown => {},
+            .darwin, .uninitialized, .unknown => {},
         }
     }
 };
@@ -72,7 +74,7 @@ fn getDefaultLinker(tc: *const Toolchain) []const u8 {
     return switch (tc.inner) {
         .uninitialized => unreachable,
         .linux => |linux| linux.getDefaultLinker(tc.getTarget()),
-        .unknown => "ld",
+        .darwin, .unknown => "ld",
     };
 }
 
@@ -93,11 +95,15 @@ pub fn discover(tc: *Toolchain) !void {
             .{ .unknown = {} } // TODO
         else
             .{ .linux = .{} },
-        else => .{ .unknown = {} }, // TODO
+        else => if (target.os.tag.isDarwin())
+            .{ .darwin = .{ .sdk = .fromTarget(tc.getTarget().*) } }
+        else
+            .{ .unknown = {} }, // TODO
     };
     return switch (tc.inner) {
         .uninitialized => unreachable,
         .linux => |*linux| linux.discover(tc),
+        .darwin => |*darwin| try darwin.discover(tc),
         .unknown => {},
     };
 }
@@ -357,7 +363,7 @@ pub fn buildLinkerArgs(tc: *Toolchain, argv: *std.ArrayList([]const u8)) !void {
     return switch (tc.inner) {
         .uninitialized => unreachable,
         .linux => |*linux| linux.buildLinkerArgs(tc, argv),
-        .unknown => @panic("This toolchain does not support linking yet"),
+        .darwin, .unknown => @panic("This toolchain does not support linking yet"),
     };
 }
 
@@ -509,6 +515,7 @@ pub fn defineSystemIncludes(tc: *Toolchain) !void {
     return switch (tc.inner) {
         .uninitialized => unreachable,
         .linux => |*linux| linux.defineSystemIncludes(tc),
+        .darwin => |*darwin| darwin.defineSystemIncludes(tc),
         .unknown => {
             if (tc.driver.nostdinc) return;
 
--- a/src/aro/Toolchain.zig
+++ b/src/aro/Toolchain.zig
@@ -103,8 +103,7 @@ pub fn discover(tc: *Toolchain) !void {
     return switch (tc.inner) {
         .uninitialized => unreachable,
         .linux => |*linux| linux.discover(tc),
-        .darwin => |*darwin| try darwin.discover(tc),
-        .unknown => {},
+        .darwin, .unknown => {},
     };
 }
 
--- a/src/aro/toolchains/Darwin.zig
+++ b/src/aro/toolchains/Darwin.zig
@@ -1,0 +1,150 @@
+const Target = @import("../Target.zig");
+const Toolchain = @import("../Toolchain.zig");
+
+const std = @import("std");
+const builtin = @import("builtin");
+
+const Darwin = @This();
+
+sdk: Sdk,
+sdk_path: ?[]const u8 = null,
+
+pub fn discover(darwin: *Darwin, tc: *Toolchain) !void {
+    if (tc.driver.sysroot) |sysroot| {
+        darwin.sdk_path = sysroot;
+    } else if (builtin.target.os.tag.isDarwin()) {
+        try tc.driver.diagnostics.add(.{ .kind = .note, .text = "--sysroot may be required when cross-compiling to Darwin targets", .location = null });
+    } else {
+        darwin.sdk_path = try darwin.sdk.xcrun(tc, "--show-sdk-path");
+    }
+}
+
+// clang args (via --verbose):
+// -resource-dir <toolchain-path>/usr/lib/clang/21
+// -isysroot <sdk-path>
+// -I/usr/local/include
+// -internal-isystem <sdk-path>/usr/local/include
+// -internal-isystem <toolchain-path>/usr/lib/clang/21/include
+// -internal-externc-isystem <sdk-path>/usr/include
+// -internal-externc-isystem <toolchain-path>/usr/include
+// -internal-iframework <sdk-path>/System/Library/Frameworks
+// -internal-iframework <sdk-path>/System/Library/SubFrameworks
+// -internal-iframework <sdk-path>/Library/Frameworks
+//
+// clang messages (via -E):
+// ignoring nonexistent directory "/usr/local/include"
+// ignoring nonexistent directory "<sdk-path>/usr/local/include"
+// ignoring nonexistent directory "<sdk-path>/Library/Frameworks"
+// #include "..." search starts here:
+// #include <...> search starts here:
+//  <toolchain-path>/usr/lib/clang/21/include
+//  <sdk-path>/usr/include
+//  <toolchain-path>/usr/include
+//  <sdk-path>/System/Library/Frameworks (framework directory)
+//  <sdk-path>/System/Library/SubFrameworks (framework directory)
+pub fn defineSystemIncludes(self: *Darwin, tc: *Toolchain) !void {
+    if (tc.driver.nostdinc) return;
+
+    var effective = self.*;
+    if (tc.driver.nostdlibinc) {
+        effective = .{ .sdk = self.sdk };
+    } else {
+        // -I/usr/local/include
+        try tc.addPathIfExists(&.{ "usr", "local", "include" }, .file);
+    }
+
+    // -internal-isystem <sdk-path>/usr/local/include
+    if (effective.sdk_path) |sdk| try tc.addPathIfExists(&.{ sdk, "usr", "local", "include" }, .file);
+    // -internal-isystem <toolchain-path>/usr/lib/clang/21/include but aro
+    if (!tc.driver.nobuiltininc) try tc.addBuiltinIncludeDir();
+
+    // -internal-externc-isystem <sdk-path>/usr/include
+    if (effective.sdk_path) |sdk| try tc.addPathIfExists(&.{ sdk, "usr", "include" }, .file);
+    // <toolchain-path>/usr/include is unnecessary - it's FlexLexer.h, libcodedirectory.h, and swift bridging
+}
+
+const Sdk = union(enum) {
+    driverkit,
+    iphoneos,
+    iphonesimulator,
+    macosx: packed struct { catalyst: bool },
+    appletvos,
+    appletvsimulator,
+    xros,
+    xrsimulator,
+    watchos,
+    watchsimulator,
+
+    pub fn fromTarget(target: Target) Sdk {
+        return switch (target.os.tag) {
+            .driverkit => .driverkit,
+            .ios => if (target.abi == .simulator) .iphonesimulator else .iphoneos,
+            .maccatalyst => .{ .macosx = .{ .catalyst = true } },
+            .macos => .{ .macosx = .{ .catalyst = false } },
+            .tvos => if (target.abi == .simulator) .appletvsimulator else .appletvos,
+            .watchos => if (target.abi == .simulator) .watchsimulator else .watchos,
+            else => {
+                std.debug.assert(!target.os.tag.isDarwin());
+                unreachable;
+            },
+        };
+    }
+
+    fn xcrun(sdk: Sdk, tc: *Toolchain, arg: []const u8) !?[]const u8 {
+        const arena = tc.driver.comp.arena;
+        const gpa = tc.driver.comp.gpa;
+
+        var message_writer = std.Io.Writer.Allocating.init(gpa);
+        defer message_writer.deinit();
+        const msg: *std.Io.Writer = &message_writer.writer;
+
+        msg.print("`xcrun --sdk {t} {s}` ", .{ sdk, arg }) catch return error.OutOfMemory;
+        const msg_reset = msg.end;
+
+        const result = std.process.run(tc.driver.comp.gpa, tc.driver.comp.io, .{
+            .argv = &.{ "xcrun", "--sdk", @tagName(sdk), arg },
+        }) catch |err| {
+            msg.print("failed to run: {t}", .{err}) catch return error.OutOfMemory;
+            return null;
+        };
+        defer gpa.free(result.stderr);
+        defer gpa.free(result.stdout);
+
+        if (result.term != .exited) {
+            msg.print("stopped unexpectedly: {}\n{s}", .{ result.term, result.stderr }) catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .@"error", .text = try arena.dupe(u8, msg.buffered()), .location = null });
+            return null;
+        } else if (result.term.exited != 0) {
+            msg.print("exited with exit code {d}\n{s}", .{ result.term.exited, result.stderr }) catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .@"error", .text = try arena.dupe(u8, msg.buffered()), .location = null });
+            return null;
+        } else if (result.stderr.len > 0) {
+            msg.print("stderr:\n{s}", .{result.stderr}) catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .warning, .text = try arena.dupe(u8, msg.buffered()), .location = null });
+            msg.end = msg_reset;
+        }
+
+        const nl = std.mem.indexOfScalar(u8, result.stdout, '\n');
+        if (nl == null or nl.? == 0) {
+            msg.writeAll("had no output") catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .@"error", .text = "had no output", .location = null });
+            return null;
+        }
+
+        if (result.stdout[nl.? + 1 ..].len > 0) {
+            msg.print("had unexpected output:\n{s}", .{result.stdout}) catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .@"error", .text = try arena.dupe(u8, msg.buffered()), .location = null });
+            return null;
+        }
+
+        const path = result.stdout[0..nl.?];
+
+        if (!tc.exists(path)) {
+            msg.print("returned non-existent path '{s}'", .{path}) catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .warning, .text = try arena.dupe(u8, msg.buffered()), .location = null });
+            msg.end = msg_reset;
+        }
+
+        return try arena.dupe(u8, path);
+    }
+};
--- a/src/aro/toolchains/Darwin.zig
+++ b/src/aro/toolchains/Darwin.zig
@@ -9,58 +9,25 @@ const Darwin = @This();
 sdk: Sdk,
 sdk_path: ?[]const u8 = null,
 
-pub fn discover(darwin: *Darwin, tc: *Toolchain) !void {
-    if (tc.driver.sysroot) |sysroot| {
-        darwin.sdk_path = sysroot;
-    } else if (builtin.target.os.tag.isDarwin()) {
-        try tc.driver.diagnostics.add(.{ .kind = .note, .text = "--sysroot may be required when cross-compiling to Darwin targets", .location = null });
-    } else {
-        darwin.sdk_path = try darwin.sdk.xcrun(tc, "--show-sdk-path");
-    }
-}
-
-// clang args (via --verbose):
-// -resource-dir <toolchain-path>/usr/lib/clang/21
-// -isysroot <sdk-path>
-// -I/usr/local/include
-// -internal-isystem <sdk-path>/usr/local/include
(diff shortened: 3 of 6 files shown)


dotcarmen:
hmmm, things broke... (besides maccatalyst, which i'm still figuring out)

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,108 @@
+const Target = @import("../Target.zig");
+const Toolchain = @import("../Toolchain.zig");
+
+const std = @import("std");
+
+const Sdk = union(enum) {
+    driverkit,
+    iphoneos,
+    iphonesimulator,
+    macosx: packed struct { catalyst: bool },
```
dotcarmen: i'm really unsure what the specific behavior for maccatalyst is... and unfortunately i can't download the ios toolchain until tomorrow. 

Bazel [uses ios minimum version](https://github.com/bazelbuild/bazel/blob/5bde1f6a9ef2917df561967287bbbb91204784e3/src/main/starlark/builtins_bzl/common/xcode/providers.bzl#L83-L89) for compatibility checks, but [uses the macosx SDK](https://github.com/bazelbuild/bazel/blob/5bde1f6a9ef2917df561967287bbbb91204784e3/src/main/starlark/builtins_bzl/common/xcode/providers.bzl#L116-L121) to build.

this week, i'll download the ios toolchain and make a fake maccatalyst app to see what the clang arguments are.

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,136 @@
...
+const Toolchain = @import("../Toolchain.zig");
+
+const std = @import("std");
+const builtin = @import("builtin");
+
+const Sdk = union(enum) {
+    driverkit,
+    iphoneos,
+    iphonesimulator,
+    macosx: packed struct { catalyst: bool },
```
dotcarmen, line 11: currently without internet at home, and i can't download the ios toolchain. [Bazel says the macosx sdk is used to build, but the ios sdk is used for version compatibility checks](https://github.com/bazelbuild/bazel/blob/5bde1f6a9ef2917df561967287bbbb91204784e3/src/main/starlark/builtins_bzl/common/xcode/providers.bzl#L116-L121)

once i'm able to download the ios sdk, i'll check the clang commands for a fake maccatalyst application

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,136 @@
...
+            },
+        };
+    }
+};
+
+const Darwin = @This();
+
+sdk: Sdk,
+sdk_path: ?[]const u8,
+// toolchain_path: ?[]const u8,
```
dotcarmen, line 39: tests pass without handling the toolchain path. however, i think the 'correct' behavior is to append toolchain paths _before_ aro paths?

if i _do_ handle this, then i need to do similar handling for `tc.driver.resource_dir` that i do fo r`tc.driver.sysroot` below

src/aro/toolchains/Darwin.zig
dotcarmen: ok i did this, but still can't figure it out :( i saw some invocations of a wrapper tool that seems to be compiling things for ios, but that can't be right?? i'm gonna try an strace next week

src/aro/toolchains/Darwin.zig
dotcarmen: nah. aro is the toolchain. we shouldn't rely on a clang toolchain

brodo:
You may want to allow uses to specify the system framework and include paths independently of `--sysroot`. We did need that for [dvui](https://github.com/david-vanderson/dvui/blob/c3921e845f520c0e6dccaf40db687dbe31613e7e/build.zig#L215-L216). I can't really remember why, though. Also, maybe [this](https://codeberg.org/brodo/xcode-frameworks) is helpful.

dotcarmen:
> You may want to allow uses to specify the system framework and include paths independently of `--sysroot`.

i'm not going to do that in this PR - that could probably done similarly to eg clang's `-internal-isystem`, which i think bypasses the automatic path-adding done here. but overall i think it's usually unnecessary. (more on this below)

framework paths will come via objc pr(s)

> We did need that for [dvui](https://github.com/david-vanderson/dvui/blob/c3921e845f520c0e6dccaf40db687dbe31613e7e/build.zig#L215-L216). I can't really remember why, though.

looks like it's for sdl3. [this is the commit](https://github.com/david-vanderson/SDL/tree/5bbcf665494fe14acadf50601edf6d5f74ebf96a) referenced in dvui's `build.zig.zon` (that commit doesn't appear to be in a branch btw). the reason why it's needed seems to be [here](https://github.com/david-vanderson/SDL/blob/5bbcf665494fe14acadf50601edf6d5f74ebf96a/build.zig#L98-L102). that use case is already covered by `-isystem` with aro :) no need to add anything new for it.

i will note as an aside that darwin system clang is actually a wrapper for clang that seems to run `xcrun` to find the toolchain and sdk for the sysroot. run any clang command with `--verbose` and something like this shows up (formatted by me):

<details>
<summary>clang output</summary>

```shell
"/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" \
  -cc1 \
  -triple arm64-apple-macosx26.0.0 \
  -Wundef-prefix=TARGET_OS_ \
  -Wdeprecated-objc-isa-usage \
  -Werror=deprecated-objc-isa-usage \
  -Werror=implicit-function-declaration \
  -emit-obj \
  -dumpdir a- \
  -disable-free \
  -clear-ast-before-backend \
  -disable-llvm-verifier \
  -discard-value-names \
  -main-file-name - \
  -mrelocation-model pic \
  -pic-level 2 \
  -mframe-pointer=non-leaf \
  -fno-strict-return \
  -ffp-contract=on \
  -fno-rounding-math \
  -funwind-tables=1 \
  -fobjc-msgsend-selector-stubs \
  -fno-sized-deallocation \
  -target-sdk-version=26.5 \
  -fvisibility-inlines-hidden-static-local-var \
  -fno-modulemap-allow-subdirectory-search \
  -fdefine-target-os-macros \
  -fno-assume-unique-vtables \
  -fno-modulemap-allow-subdirectory-search \
  -enable-tlsdesc \
  -target-cpu apple-m1
  -target-feature +v8.5a \
  -target-feature +aes \
  -target-feature +altnzcv \
  -target-feature +ccdp \
  -target-feature +ccpp \
  -target-feature +complxnum \
  -target-feature +crc \
  -target-feature +dotprod \
  -target-feature +flagm \
  -target-feature +fp-armv8 \
  -target-feature +fp16fml \
  -target-feature +fptoint \
  -target-feature +fullfp16 \
  -target-feature +jsconv \
  -target-feature +lse \
  -target-feature +neon \
  -target-feature +pauth \
  -target-feature +perfmon \
  -target-feature +predres \
  -target-feature +ras \
  -target-feature +rcpc \
  -target-feature +rdm \
  -target-feature +sb \
  -target-feature +sha2 \
  -target-feature +sha3 \
  -target-feature +specrestrict \
  -target-feature +ssbs \
  -target-abi darwinpcs \
  -debugger-tuning=lldb \
  -fdebug-compilation-dir=/Users/carmen/github.com/Vexu/arocc \
  -target-linker-version 1267 \
  -v \
  -fcoverage-compilation-dir=/Users/carmen/github.com/Vexu/arocc \
  -resource-dir /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/21 \
  -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk \
  -I/usr/local/include \
  -internal-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include \
  -internal-isystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/21/include \
  -internal-externc-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include \
  -internal-externc-isystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include \
  -internal-iframework /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks \
  -internal-iframework /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/SubFrameworks \
  -internal-iframework /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/Library/Frameworks \
  -Wno-elaborated-enum-base \
  -ferror-limit 19 \
  -fmessage-length=101 \
  -stack-protector 1 \
  -fstack-check \
  -mdarwin-stkchk-strong-link \
  -fblocks \
  -fencode-extended-block-signature \
  -fregister-global-dtors-with-atexit \
  -fgnuc-version=4.2.1 \
  -fskip-odr-check-in-gmf \
  -fmax-type-align=16 \
  -fcommon \
  -fcolor-diagnostics \
  -clang-vendor-feature=+disableNonDependentMemberExprInCurrentInstantiation \
  -fno-odr-hash-protocols \
  -clang-vendor-feature=+enableAggressiveVLAFolding \
  -clang-vendor-feature=+revert09abecef7bbf \
  -clang-vendor-feature=+thisNoAlignAttr \
  -clang-vendor-feature=+thisNoNullAttr \
  -clang-vendor-feature=+disableAtImportPrivateFrameworkInImplementationError \
  -Wno-error=allocator-wrappers \
  -fdwarf2-cfi-asm \
  -o /var/folders/5h/0zrj5b294l96v9hpw7gyk_qw0000gn/T/--d78b35.o \
  -x c \
  -
```

</details>

those `-internal-*` flags look like they might be used to prevent adding the relevant paths based on `sysroot`. i also won't be implementing that (not particularly helpful, i believe), but it's something to consider

> Also, maybe [this](https://codeberg.org/brodo/xcode-frameworks) is helpful.

that would be helpful for cross-platform testing objc work :) it would probably only be a test dependency though - i imagine if users want to cross-compile macos applications, they would probably want to control the SDK versions

src/aro/toolchains/Darwin.zig
dotcarmen: btw turns out strace is really hard on macos. i'm gonna follow bazel's lead here

dotcarmen:
> that could probably done similarly to eg clang's `-internal-isystem`, which i think bypasses the automatic path-adding done here.

i forgot about `-nostdinc` :) so yeah, `-nostdinc -isystem ...` would be the way to go if you really need to bypass the logic

src/aro/toolchains/Darwin.zig
dotcarmen, line 1: i had another commit that used `xcrun xcodebuild -sdk -version`, but that looked a lot different and i think this is good enough, at least for now

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,150 @@
...
+// ignoring nonexistent directory "/usr/local/include"
+// ignoring nonexistent directory "<sdk-path>/usr/local/include"
+// ignoring nonexistent directory "<sdk-path>/Library/Frameworks"
+// #include "..." search starts here:
+// #include <...> search starts here:
+//  <toolchain-path>/usr/lib/clang/21/include
+//  <sdk-path>/usr/include
+//  <toolchain-path>/usr/include
+//  <sdk-path>/System/Library/Frameworks (framework directory)
+//  <sdk-path>/System/Library/SubFrameworks (framework directory)
```
dotcarmen: comments here and below might be helpful in the future, but lmk if i should remove

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,150 @@
...
+const Darwin = @This();
+
+sdk: Sdk,
+sdk_path: ?[]const u8 = null,
+
+pub fn discover(darwin: *Darwin, tc: *Toolchain) !void {
+    if (tc.driver.sysroot) |sysroot| {
+        darwin.sdk_path = sysroot;
+    } else if (builtin.target.os.tag.isDarwin()) {
+        try tc.driver.diagnostics.add(.{ .kind = .note, .text = "--sysroot may be required when cross-compiling to Darwin targets", .location = null });
```
dotcarmen: apparently this can be a fatal error with `-Weverything -Werror -Wfatal`, which is why `Driver.main` now has to handle `error.FatalError`. this can be printed with `std.debug.print` instead, i can do that if desired

Vexu: None of those should affect a note, the warnings in  `xrun()` would be affected but that could be moved to `defineSystemIncludes`?

src/aro/toolchains/Darwin.zig
Vexu: Please do.

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,150 @@
...
+
+        var message_writer = std.Io.Writer.Allocating.init(gpa);
+        defer message_writer.deinit();
+        const msg: *std.Io.Writer = &message_writer.writer;
+
+        msg.print("`xcrun --sdk {t} {s}` ", .{ sdk, arg }) catch return error.OutOfMemory;
+        const msg_reset = msg.end;
+
+        const result = std.process.run(tc.driver.comp.gpa, tc.driver.comp.io, .{
+            .argv = &.{ "xcrun", "--sdk", @tagName(sdk), arg },
```
Vexu, line 71: Is `catalyst` not passed to `xrun`?

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,150 @@
...
+        }) catch |err| {
+            msg.print("failed to run: {t}", .{err}) catch return error.OutOfMemory;
+            return null;
+        };
+        defer gpa.free(result.stderr);
+        defer gpa.free(result.stdout);
+
+        if (result.term != .exited) {
+            msg.print("stopped unexpectedly: {}\n{s}", .{ result.term, result.stderr }) catch return error.OutOfMemory;
+            try tc.driver.diagnostics.add(.{ .kind = .@"error", .text = try arena.dupe(u8, msg.buffered()), .location = null });
```
Vexu: These dupes are unnecessary, also maybe consider moving some of this verbosity into a helper function.

src/aro/toolchains/Darwin.zig
Vexu: If this note is about cross compilation shouldn't it also check that the target is not native.

src/aro/toolchains/Darwin.zig
dotcarmen: no - see [bazel's notes](https://github.com/bazelbuild/bazel/blob/5bde1f6a9ef2917df561967287bbbb91204784e3/src/main/starlark/builtins_bzl/common/xcode/providers.bzl#L116-L121)

short answer: there is no catalyst sdk. catalyst instead seems to compile using macos libraries, but with ios version checks... which i don't believe are currently handled (i'll be adding `@available` as part of objc work)

src/aro/toolchains/Darwin.zig
dotcarmen: the darwin toolchain is only selected for darwin targets, so i'll just assert that here

Vexu: I meant cross compilation as in from macos to ios (or to a different architecture), or are you supposed to get this error when compiling natively on darwin?

ehaas:
I tried testing this out with
```c
#include <stdio.h>
```
(uname is `25.5.0 Darwin Kernel Version 25.5.0: Tue Jun  9 22:28:34 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T6041 arm64`)

`arocc test.c` gave me : `note: --sysroot may be required when cross-compiling to Darwin targets`
Then I tried: `arocc test.c -isysroot "$(xcrun --sdk macosx --show-sdk-path)"` and got:
```
test.c:7:10: fatal error: 'stdio.h' not found
#include <stdio.h>
         ^
```

My sdk path is `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk`

Am I missing any steps / is there some other way I can test it?

dotcarmen:
ah, i had the logic backwards, and forgot to test with a local file (just relied on tests not breaking). sorry yall.

the latest commit fixes boolean logic and also the logic for adding the include paths. i've tested locally and including `stdio.h` now works :)

src/aro/toolchains/Darwin.zig
```diff
@@ -69,17 +70,23 @@ const Sdk = union(enum) {
...
         };
         defer gpa.free(result.stderr);
         defer gpa.free(result.stdout);
 
         if (result.term != .exited) {
-            try tc.driver.err("{s} stopped unexpectedly: {}\n{s}", .{ pretty_cmd, result.term, result.stderr });
+            var aw = std.Io.Writer.Allocating.init(gpa);
+            defer aw.deinit();
+            aw.writer.print("{any}", .{result.term}) catch return error.OutOfMemory;
+            try tc.driver.err("{s} stopped unexpectedly: {}\n{s}", .{ pretty_cmd, aw.writer.buffered(), result.stderr });
```
dotcarmen, line 83: considered adding support for `format` method to `Diagnostics.formatArgs`, lmk if that's preferred

src/aro/Toolchain.zig
```diff
@@ -529,6 +529,11 @@ pub fn defineSystemIncludes(tc: *Toolchain) !void {
     };
 }
 
+pub fn addSystemIncludeDirJoined(tc: *const Toolchain, components: []const []const u8) !void {
```
dotcarmen, line 532: considered changing the 2nd param to `addSystemIncludeDir`, lmk if that's preferred

dotcarmen:
ci is failing because macos `stdint.h` doesn't have some integer types, ie
```
INTPTR_WIDTH
UINTPTR_WIDTH
INTMAX_WIDTH
UINTMAX_WIDTH
PTRDIFF_WIDTH
SIZE_WIDTH
WCHAR_WIDTH

INT64_WIDTH
UINT64_WIDTH
INT32_WIDTH
UINT32_WIDTH
INT16_WIDTH
UINT16_WIDTH
INT8_WIDTH
UINT8_WIDTH
```

what should i do?
- add a builtin "compatibility" header (to make sure users' code doesn't break)
  - but should this only affect `--emulate=none`?
- check for `__x_WIDTH__` for the first half of symbols, and define the 2nd half of symbols? (the 1st half are simply renamed with surrounding `__`, i can't find corresponding symbols for the 2nd half)
- redefine the first half of symbols, define the 2nd half in `test_helpers.h`?

i have commits ready for the other 2, but i think the 1st is the best solution so i'll be working on that

dotcarmen:
> i have commits ready for the other 2, but i think the 1st is the best solution so i'll be working on that

realizing this would imply aro attempts to ensure c23 compatibility on darwin, even though darwin doesn't have full c23 compatibility ootb. lmk if that's not preferred

src/aro/toolchains/Darwin.zig
dotcarmen: fixed in [fef45bb](https://github.com/Vexu/arocc/pull/974/commits/fef45bb1cb1df3dc65ffc9141c0638f1a1c4514d)

include/stdint.h
```diff
@@ -7,6 +7,29 @@
 
 # include_next <stdint.h>
 
+# if defined(__APPLE__) && __ARO_EMULATE__ == __ARO_EMULATE_NO__
```
Vexu: I don't think this check is correct since C23 should work with non-apple clang. Redefining them also doesn't matter so no reason not to.

src/aro/toolchains/Darwin.zig
```diff
@@ -0,0 +1,114 @@
...
+    if (tc.driver.nostdinc) return;
+
+    if (!tc.driver.nostdlibinc) {
+        if (tc.driver.sysroot) |sysroot| {
+            self.sdk_path = sysroot;
+        } else if (builtin.target.os.tag.isDarwin()) {
+            self.sdk_path = try self.sdk.xcrun(tc, "--show-sdk-path");
+        } else {
+            std.debug.assert(tc.getTarget().os.tag.isDarwin());
+            try tc.driver.diagnostics.add(.{ .kind = .note, .text = "--sysroot may be required when cross-compiling to Darwin targets", .location = null });
```
Vexu, line 22: Do you not need to set sysroot when cross-compiling from one darwin OS to another?

Vexu (approved):
I don't have an Apple device to test this on so I'll trust you to have done so.

include/stdint.h
dotcarmen: wouldn't redefining them result in a redefined macro diagnostic? i can instead use ifdefs?

i'll at least remove the emulate check

Vexu: > wouldn't redefining them result in a redefined macro diagnostic?

Not in a system header.

src/aro/toolchains/Darwin.zig
dotcarmen: so, for clang, it seems you do (relevant: `clang: warning: using sysroot for 'MacOSX' but targeting 'iPhone' [-Wincompatible-sysroot]`):

```console
$ clang -target arm64-apple-ios -fsyntax-only --verbose foo.c
Apple clang version 21.0.0 (clang-2100.1.1.101)
Target: arm64-apple-ios
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
clang: warning: using sysroot for 'MacOSX' but targeting 'iPhone' [-Wincompatible-sysroot]
 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" -cc1 -triple arm64-apple-ios7.0.0 -Wundef-prefix=TARGET_OS_ -Wdeprecated-objc-isa-usage -Werror=deprecated-objc-isa-usage -Werror=implicit-function-declaration -fsyntax-only -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name foo.c -mrelocation-model pic -pic-level 2 -mframe-pointer=non-leaf -fno-strict-return -ffp-contract=on -fno-rounding-math -funwind-tables=1 -faligned-alloc-unavailable -fobjc-msgsend-selector-stubs -fno-sized-deallocation -target-sdk-version=26.5 -fvisibility-inlines-hidden-static-local-var -fno-modulemap-allow-subdirectory-search -fdefine-target-os-macros -fno-assume-unique-vtables -fno-modulemap-allow-subdirectory-search -enable-tlsdesc -target-cpu apple-a7 -target-feature +v8a -target-feature +aes -target-feature +fp-armv8 -target-feature +neon -target-feature +perfmon -target-feature +sha2 -target-abi darwinpcs -debugger-tuning=lldb -fdebug-compilation-dir=/Users/carmen/github.com/Vexu/arocc -target-linker-version 1267 -v -fcoverage-compilation-dir=/Users/carmen/github.com/Vexu/arocc -resource-dir /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/21 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -I/usr/local/include -internal-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include -internal-isystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/21/include -internal-externc-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include -internal-externc-isystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include -internal-iframework /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks -internal-iframework /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/SubFrameworks -internal-iframework /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/Library/Frameworks -Wno-elaborated-enum-base -ferror-limit 19 -stack-protector 1 -mdarwin-stkchk-strong-link -fblocks -fencode-extended-block-signature -fregister-global-dtors-with-atexit -fgnuc-version=4.2.1 -fskip-odr-check-in-gmf -fmax-type-align=16 -fcommon -clang-vendor-feature=+disableNonDependentMemberExprInCurrentInstantiation -fno-odr-hash-protocols -clang-vendor-feature=+enableAggressiveVLAFolding -clang-vendor-feature=+revert09abecef7bbf -clang-vendor-feature=+thisNoAlignAttr -clang-vendor-feature=+thisNoNullAttr -clang-vendor-feature=+disableAtImportPrivateFrameworkInImplementationError -Wno-error=allocator-wrappers -fdwarf2-cfi-asm -x c foo.c
clang -cc1 version 21.0.0 (clang-2100.1.1.101) default target arm64-apple-darwin25.5.0
ignoring nonexistent directory "/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/Library/Frameworks"
#include "..." search starts here:
#include <...> search starts here:
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/21/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/SubFrameworks (framework directory)
End of search list.

```

however, i think this is rather annoying, and we're invoking `xcrun` anyways... why not just call `xcrun` with the proper sdk we need?

if you think we should match clang's behavior, or we only shouldn't when `--emulate=none`, i can update to match

dotcarmen:
> I don't have an Apple device to test this on so I'll trust you to have done so.

should i add a platform-based test suite? this seems like something worth having tests for to detect breakages in the future (in case apple does something shocking and weird, like supporting c23)

include/stdint.h
dotcarmen: ok, should i keep the apple check? in case these defines would undefine the symbols?

src/aro/toolchains/Darwin.zig
Vexu: I guess clang doesn't use xcrun then? I agree that your version does seem better. (Is xcrun always available?)

Vexu:
> > I don't have an Apple device to test this on so I'll trust you to have done so.
> 
> should i add a platform-based test suite? this seems like something worth having tests for to detect breakages in the future (in case apple does something shocking and weird, like supporting c23)

What would those tests be like?

src/aro/toolchains/Darwin.zig
dotcarmen: > I guess clang doesn't use xcrun then?

i guess not? it might use it to get the macosx sdk path, but it's just as likely that it hardcodes the macosx sdk path or copies some of the logic from `xcrun`

> Is xcrun always available?

honestly, i'm not sure. i don't have a machine immediately accessible that i can wipe, either. however, its full path is `/usr/bin/xcrun`, so i think it's safe to assume it is? I believe if it was installed as part of the command line tools or xcode, it would be in one of their prefixes

it does appear `libxcrun.dylib` is distributed with both xcode and via CommandLineTools: 

```console
$ fd xcrun /Applications/Xcode.app/Contents/ /Library/Developer/CommandLineTools
/Library/Developer/CommandLineTools/usr/share/man/man1/xcrun.1
/Library/Developer/CommandLineTools/usr/lib/libxcrun.dylib
/Applications/Xcode.app/Contents/Developer/usr/share/man/man1/xcrun.1
/Applications/Xcode.app/Contents/Developer/usr/lib/libxcrun.dylib
```

i *believe* invoking `xcrun` without having installed CommandLineTools or Xcode (thus `libxcrun.dylib` isn't available) results in a prompt to the user to invoke the command to accept EULA themselves. that or it results in the EULA prompt. i'm not sure clang works at all without these.

i wonder if `xcrun` has a certain exit code when `libxcrun.dylib` isn't available? if so, then the `xcrun` logic can throw a special error to signal that, which can be caught to result in an empty sysroot (the current behavior of aro)

i can certainly test all of this, but it's gonna be some time before i can. i'm moving soon and the next few weeks are quite busy for me

dotcarmen:
> What would those tests be like?

i guess it can be an extension of the current integration tests. basically recognize eg `.macos.c` (so the extension before `.c`) which would only run when `@import("builtin").target.os.tag.isDarwin()`. then could add tests for targeting eg ios, this might need additional setup in CI.

then in macos-only CI we can try including macos-only headers, like `<macos sdk path>/usr/include/xcselect.h`

now that i think about it, it might be good to have a flag like maybe `--no-native-toolchain` which forces toolchain to be `.unknown` for linux and macos. then we can compare baseline vs native integration

include/stdint.h
Vexu: I guess that's fine.

src/aro/toolchains/Darwin.zig
Vexu: No need, I was just curious. I'm fine with using it as long as there's no reason not to.

Vexu:
It'd probably be good to have some toolchain tests in the future but it's not a requirement for this PR.

dotcarmen:
~~hm, hello world is failing. one sec~~

nevermind, wrong branch

dotcarmen:
^ is a good point actually. i believe the only reason hello world doesn't work in the integration test suite is because it wouldn't pass windows. i'm gonna add it anyways and use ifdef to prevent the test from running on windows. lmk if you object

dotcarmen:
actually, since `hello world.c` is already a test, i'm gonna change it to use `stdio.h` on linux and darwin

PR #976, opened 2026-03-12
Normalize include paths before source-cache lookup

0pendev:
### Summary

- Fix duplicate source loading when the same header is reached via different relative `../` paths (e.g. `common/shared.h` vs `handler/../common/shared.h`).
- Normalize include paths with `std.fs.path.resolvePosix` before keying into the source cache, so that logically identical files map to a single cache entry.
- Store the normalized path in the cache so all subsequent lookups (regardless of the original include spelling) hit the same entry.

### Problem

The source cache in `Compilation.addSourceFromPathExtra` was keyed on the **raw include-path string** as provided by the preprocessor. When the same physical header file was included from different directories, each directory produced a different relative path containing `../` segments that were never collapsed.

For example, given this layout:

```
project/
├── common/
│   └── shared.h          (#pragma once, defines shared_t)
├── handler/
│   └── handler_types.h   (#include "../common/shared.h")
└── main.c                (#include "common/shared.h"
                            #include "handler/handler_types.h")
```

When the preprocessor processes these includes, the cache sees two different keys for the same file:

| Include site | Path passed to source cache |
|---|---|
| `main.c` | `common/shared.h` |
| `handler/handler_types.h` | `handler/../common/shared.h` |

Because these two strings differ, the cache treated them as **two distinct sources**, even though they refer to the same file on disk. This caused:

1. **`#pragma once` failures** — the guard is per-source-ID, so the second load didn't recognize the file as already included, leading to duplicate type definitions or, in pathological `./`-loop cases, infinite self-include recursion.
2. **Redundant preprocessing** — the same header was lexed and preprocessed multiple times, wasting work proportional to the number of distinct path spellings.

### Reproducer

Create the following file tree:

**`common/shared.h`**
```c
#pragma once

typedef struct {
    int id;
    int value;
} shared_t;
```

**`handler/handler_types.h`**
```c
#pragma once
#include "../common/shared.h"

typedef struct {
    shared_t state;
    int flags;
} handler_info_t;
```

**`main.c`**
```c
#include "common/shared.h"
#include "handler/handler_types.h"

shared_t global_state;
handler_info_t handler_info;
```

Without the fix, compiling `main.c` with arocc produces a duplicate definition error for `shared_t` because `#pragma once` fails to deduplicate the two source-cache entries for `common/shared.h`.

### Fix

In `src/aro/Compilation.zig`, inside `addSourceFromPathExtra`:

1. After the initial raw-path cache lookup, **normalize** the path using `std.fs.path.resolvePosix` to collapse `/./"` and `X/../` segments.
2. If the normalized path differs from the raw path, perform a **second cache lookup** with the normalized form.
3. When inserting a new source into the cache, use the **normalized path** as the key so future lookups (from any spelling) converge on the same entry.

```1669:1693:src/aro/Compilation.zig
fn addSourceFromPathExtra(comp: *Compilation, path: []const u8, kind: Source.Kind) !Source {
    if (comp.sources.get(path)) |some| return some;

    // Normalize the path by collapsing "/./" and "X/../" segments so that
    // different relative paths to the same file resolve to one cache entry.
    // This prevents #pragma once failures and redundant preprocessing.
    const normalized = try std.fs.path.resolvePosix(comp.gpa, &.{path});
    defer comp.gpa.free(normalized);
    if (!mem.eql(u8, normalized, path)) {
        if (comp.sources.get(normalized)) |some| return some;
    }

    if (mem.indexOfScalar(u8, path, 0) != null) {
        return error.FileNotFound;
    }

    const file = try comp.cwd.openFile(comp.io, path, .{});
    defer file.close(comp.io);
    return comp.addSourceFromFile(file, normalized, kind);
}
```


Vexu (changes requested):

Vexu (approved):
Could you also make `addSourceFromOwnedBuffer` take a `[]u8` for path so that it doesn't need to be duped twice? And a test would be nice.

0pendev:
> Could you also make `addSourceFromOwnedBuffer` take a `[]u8` for path so that it doesn't need to be duped twice? And a test would be nice.

How do want me to add the test case ? Is it okay to add a subdirectory for the includes needed for the test?

src/aro/Compilation.zig
```diff
@@ -1914,7 +1926,13 @@ const FindInclude = struct {
     }
     fn checkIncludeDir(find: *FindInclude, include_dir: []const u8, kind: Source.Kind) Allocator.Error!?Result {
         if (find.wait_for) |wait_for| {
-            if (std.mem.eql(u8, include_dir, wait_for)) find.wait_for = null;
+            if (std.mem.eql(u8, include_dir, wait_for)) {
+                find.wait_for = null;
+            } else {
+                const resolved = try std.fs.path.resolve(find.comp.gpa, &.{include_dir});
+                defer find.comp.gpa.free(resolved);
```
Vexu: Resolving the path here instead of `run` will lead to it being resolved multiple times redundantly.

Vexu:
Looking at `FindInclude` more I think replacing the `allocPrint` in `check` with a resolve would make it unnecessary to resolve the path in `addSourceFromPathExtra` avoiding having to modify `addSourceFromOwnedBuffer`.

src/aro/Compilation.zig
```diff
@@ -1900,17 +1893,19 @@ const FindInclude = struct {
             },
         }
         for (comp.search_path.items) |include| {
+            const resolved_path = try std.fs.path.resolve(find.comp.gpa, &.{include.path});
```
Vexu, line 1896: This looks redundant.

0pendev: From the tests I did The resolve is needed for #include_next support.

0pendev:
@Vexu Anything missing here I could add ? :pray:

bandithedoge:
Since Zig now uses Aro for C translation, this can block projects from updating ([example workaround](https://codeberg.org/bandithedoge/zigplug/src/commit/fad7f3ae14efc4477ccf466f37f96313f2a474da/src/clap/build.zig#L21-L29)), so it would be great if we could get this merged during the 0.16.1 release cycle.

Vexu (approved):
Sorry for taking so long with this. I updated this to some Zig changes and adjusted the include_next check a bit.

0pendev:
> Sorry for taking so long with this. I updated this to some Zig changes and adjusted the include_next check a bit.

Thanks :pray:

PR #985, opened 2026-03-15
Parser: support MSVC-style fixed-size integer suffixes

### LINKED ISSUE #956: Missing support for MSVC integer suffixes like `i64`, `ui64`
```c
#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>

bool f(void) {
    return (size_t)0 < SIZE_MAX;
}
```
```
zig translate-c foo.h -lc -target x86_64-windows-msvc
```
```
error: translation failure
foo.h:6:24: error: invalid suffix 'ui64' on integer constant
    return (size_t)0 < SIZE_MAX;
                       ^
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.44.35207\include\stdint.h:108:26: note: expanded from here
        #define SIZE_MAX 0xffffffffffffffffui64
```

`i64` and `ui64` (case-insensitive) are officially documented MSVC integer suffix extensions.
https://learn.microsoft.com/en-us/cpp/c-language/c-integer-constants?view=msvc-170

Apparently `i8`, `ui8`, `i16`, `ui16`, `i32` and `ui32` are also supported, but undocumented, and are used by MSVC headers for e.g. `#define UINT8_MAX 0xffui8`.
https://stackoverflow.com/questions/33659846/microsoft-integer-literal-extensions-where-documented

ehaas:
Closes #956


src/aro/Parser.zig
```diff
@@ -10607,7 +10650,8 @@ fn getExponent(p: *Parser, buf: []const u8, prefix: NumberPrefix, tok_i: TokenIn
 /// to parse numbers in pragma handlers.
 pub fn parseNumberToken(p: *Parser, tok_i: TokenIndex) !Result {
     const buf = p.tokSlice(tok_i);
-    const prefix = NumberPrefix.fromString(buf);
+    const is_msvc = p.comp.langopts.emulate == .msvc;
```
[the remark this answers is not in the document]
ehaas: Upon looking into this more, it looks like clang (I tested x86_64) with ms-extensions treats `__int64` as `long long`, but an `i64` literal as `int64_t` (long). On x86-64 MSVC `__int64` and `int64_t` are both `long long` so it doesn't matter. So basically it looks like those literals will need to use the platform-specific int64_t type. `i8` suffix on the other hand uses `char` (same as `__int8`), whereas `int8_t` is `signed char`. I'll have to dig into this more.

kafji:
Hello. I want to compile my Zig program for MSVC but blocked by this.

Is there anything I can help? What's holding this PR from being merged?


I write the following assertions to check their types and are true for:
- MSVC x64
- MSVC x84
- Clang -target x86_64-pc-windows-msvc
- Clang -target i386-pc-windows-msvc

```c
#include <Windows.h>
#include <type_traits>

#define I8 0I8
#define UI8 0UI8
#define I16 0I16
#define UI16 0UI16
#define I32 0I32
#define UI32 0UI32
#define I64 0I64
#define UI64 0UI64

#define ASSERT_SIZE(v, s) static_assert(sizeof(decltype(v)) == s)

#define ASSERT_SIGNEDNESS(v, s)                                                \
  static_assert(std::is_signed_v<decltype(v)> == s)

#define ASSERT_TYPE(v, t)                                                      \
  static_assert(sizeof(decltype(v)) == sizeof(t) &&                            \
                std::is_signed_v<decltype(v)> == std::is_signed_v<t>)

#define ASSERT_EQUAL(t, u)                                                     \
  static_assert(sizeof(t) == sizeof(u) &&                                      \
                std::is_signed_v<t> == std::is_signed_v<u>)

ASSERT_SIZE(I8, 1);
ASSERT_SIZE(UI8, 1);
ASSERT_SIZE(I16, 2);
ASSERT_SIZE(UI16, 2);
ASSERT_SIZE(I32, 4);
ASSERT_SIZE(UI32, 4);
ASSERT_SIZE(I64, 8);
ASSERT_SIZE(UI64, 8);

ASSERT_SIGNEDNESS(I8, true);
ASSERT_SIGNEDNESS(UI8, false);
ASSERT_SIGNEDNESS(I16, true);
ASSERT_SIGNEDNESS(UI16, false);
ASSERT_SIGNEDNESS(I32, true);
ASSERT_SIGNEDNESS(UI32, false);
ASSERT_SIGNEDNESS(I64, true);
ASSERT_SIGNEDNESS(UI64, false);

ASSERT_TYPE(I8, char);
ASSERT_TYPE(UI8, unsigned char);
ASSERT_TYPE(I16, short);
ASSERT_TYPE(UI16, unsigned short);
ASSERT_TYPE(I32, int);
ASSERT_TYPE(UI32, unsigned int);
ASSERT_TYPE(I64, long long);
ASSERT_TYPE(UI64, unsigned long long);

ASSERT_TYPE(UI64, UINT64);
ASSERT_EQUAL(UINT64, unsigned __int64);
ASSERT_EQUAL(__int64, long long);
ASSERT_EQUAL(__int64, int64_t);
```


---

@ehaas 

> it looks like clang (I tested x86_64) with ms-extensions treats __int64 as long long, but an i64 literal as int64_t (long)

`int64_t` is `long long` in MSVC, regardless of which compiler.

```c
// stdint.h
typedef long long          int64_t;
```

ehaas:
This is the test I did: https://godbolt.org/z/9zhszxr5h (the issue was with clang, not msvc)

I don't think the remaining work is too extensive, I just unfortunately haven't had time to take it over the finish line. I think the biggest help would be an exhaustive catalog of how clang and msvc treat these suffixes on a variety of platforms, for test cases.  The slightly tricky bit is that if it's a platform-specific type (e.g. something defined in `stdint.h`) we can't just directly map it to a C integer type, we have to do a sort of runtime lookup - the parser sees "integer with `i64` suffix" but in clang mode that's not enough information to know what type it is. It's "whatever the current target defines int64_t" to be, which could be `long` or `long long`.

kafji:
Ah, didn't realize that you can `-fms-extensions` while targeting non-Windows.

Here's the result of probing `clang -fms-extensions -target $target` for common Clang targets:

| Target \ Suffix | `I8` | `UI8` | `I16` | `UI16` | `I32` | `UI32` | `I64` | `UI64` |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `arm64-apple-darwin` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long` | `unsigned long` |
| `arm64-linux-gnu` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long` | `unsigned long` |
| `arm64-windows-gnu` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `arm64-windows-msvc` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `i386-apple-darwin` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `i386-linux-gnu` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `i386-windows-gnu` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `i386-windows-msvc` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `x86_64-apple-darwin` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long` | `unsigned long` |
| `x86_64-linux-gnu` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long` | `unsigned long` |
| `x86_64-windows-gnu` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |
| `x86_64-windows-msvc` | `char` | `unsigned char` | `short` | `unsigned short` | `int` | `unsigned int` | `long long` | `unsigned long long` |

What target triples does Aro support?

These are {aarch64, x86, x86-64}-ANY_OS-ANY_ABI from `zig targets`:

https://gist.github.com/kafji/9e4ff2f165a9953c4feba3d7b6aea1cd

I can extend this to other arches as needed.

> The slightly tricky bit is that if it's a platform-specific type (e.g. something defined in stdint.h) we can't just directly map it to a C integer type, we have to do a sort of runtime lookup - the parser sees "integer with i64 suffix" but in clang that's not enough information to know what type it is. It's "whatever the current target defines int64_t" to be, which could be long or long long.

I think I'm missing something here. Why is the size of these suffixes depend on types in `stdint.h`?

src/aro/Parser.zig
```diff
@@ -2194,6 +2194,29 @@ fn initDeclarator(p: *Parser, decl_spec: *DeclSpec, attr_buf_top: usize, decl_no
...
+        else => unreachable,
+    };
+}
+
+fn getMSVCIntSuffixType(p: *Parser, bits: enum { @"8", @"16", @"32", @"64" }, signedness: std.builtin.Signedness) QualType {
+    return switch (bits) {
+        .@"8" => if (signedness == .signed) .char else .uchar,
+        .@"16" => if (signedness == .signed) .short else .ushort,
+        .@"32" => if (signedness == .signed) .int else .uint,
+        .@"64" => if (p.comp.langopts.emulate != .msvc and
```
ehaas: This feels wrong. I initially had this:
```zig
        .@"64" => if (p.comp.langopts.emulate != .msvc and signedness == .signed) p.comp.target.int64Type() else if (p.comp.langopts.emulate != .msvc and signedness == .unsigned) p.comp.target.int64Type().makeIntUnsigned(p.comp) catch unreachable else if (signedness == .signed) .long_long else .ulong_long,
```
just to try to get something working but it was failing for `x86_64-apple-darwin` and `arm64-apple-darwin.c`

ehaas:
@kafji I added all your tests and pushed an update where they all pass, but I'm still not sure about the 64-bit path. I'm not sure if our int64 type is wrong or there is some other rule being followed that I'm just missing. I find it very hard to read C++ code but I was trying to go through how clang handles it and I *think* they just find the smallest 64-bit type, rather than use the platform's int64_t. So we should probably do that instead of looking at pointer widths. I can test out that approach tomorrow but I think we're close to having a solution here.

ehaas:
Blocked on https://codeberg.org/ziglang/zig/issues/35442

src/aro/LangOpts.zig
```diff
@@ -190,3 +190,11 @@ pub fn setFpEvalMethod(self: *LangOpts, fp_eval_method: FPEvalMethod) void {
 pub fn setCharSignedness(self: *LangOpts, signedness: std.builtin.Signedness) void {
     self.char_signedness_override = signedness;
 }
+
+pub fn allowFixedSizedIntSuffixes(self: *const LangOpts) bool {
+    return switch (self.emulate) {
+        .msvc => true,
+        .clang => self.ms_extensions,
+        .no, .gcc => false,
```
Vexu:
```suggestion
        .clang, .no => self.ms_extensions,
        .gcc => false,
```
Why not?

ehaas:
@Vexu I put in a workaround for the linked issue but should we just remove that target from the tests? It hasn't been supported by Apple since 2019 as far as I can tell.

src/aro/Parser.zig
```diff
@@ -2194,6 +2194,25 @@ fn initDeclarator(p: *Parser, decl_spec: *DeclSpec, attr_buf_top: usize, decl_no
...
+        else => unreachable,
+    };
+}
+
+fn getMSVCIntSuffixType(p: *Parser, bits: enum { @"8", @"16", @"32", @"64" }, signedness: std.builtin.Signedness) QualType {
+    return switch (bits) {
+        .@"8" => if (signedness == .signed) .char else .uchar,
+        .@"16" => p.comp.intLeastN(16, signedness),
+        .@"32" => p.comp.intLeastN(32, signedness),
+        .@"64" => if (p.comp.langopts.emulate != .msvc) p.comp.smallestNBitIntTargetIndependent(64, signedness) else if (signedness == .signed) .long_long else .ulong_long,
```
Vexu, line 2212: Does this emulate check matter on any target? `long long` is always the smallest 64 bit int type on Windows.

Vexu (approved):
> Vexu I put in a workaround for the linked issue but should we just remove that target from the tests? It hasn't been supported by Apple since 2019 as far as I can tell.

Sure, I don't think this feature needs this many tests any way.

src/aro/Parser.zig
ehaas: without that check it would fail when emulating msvc on non-windows platforms

ehaas:
Removed a bunch of tests + the `long` size workaround for `x86-macos`.

kafji:
Thank you!

PR #992, opened 2026-04-12
fix Apple compiler detection regression

brodo:
This PR sets TARGET_CPU_ and TARGET_OS_ when targeting Apple OSes.

`TargetConditionals.h` has been updated recently and which leads to arocc not being able to compile Libraries that use it. As an example, see the `zig16-dev` branch of [dvui](https://github.com/david-vanderson/dvui/tree/zig16-dev).

In `TargetConditionals.h`, the suggested fix is described on line 385:

```
The short term work around is to set the TARGET_CPU_ and TARGET_OS_
on the command line to the compiler (e.g. -DTARGET_CPU_MIPS=1 -DTARGET_OS_UNIX=1)
```

This PR does just that.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -376,7 +376,19 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
         => {
             try define(w, "__APPLE__");
             try w.writeAll("#define __APPLE_CC__ 6000\n");
+            try define(w, "TARGET_OS_MAC");
 
+            const target_str = switch (target.os.tag) {
+                .maccatalyst => "TARGET_OS_MACCATALYST",
+                .macos => "TARGET_OS_OSX",
+                .tvos => "TARGET_OS_TV",
+                .ios => "TARGET_OS_IOS",
+                .driverkit => "TARGET_OS_DRIVERKIT",
+                .visionos => "TARGET_OS_VISION",
+                .watchos => "TARGET_OS_WATCH",
+                else => unreachable,
+            };
+            try define(w, target_str);
             const version = target.os.version_range.semver.min;
             var version_buf: [8]u8 = undefined;
             const version_str = if (target.os.tag == .macos and version.order(.{ .major = 10, .minor = 10, .patch = 0 }) == .lt)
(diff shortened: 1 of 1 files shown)


Vexu:
Can you point to me to the version of `TargetConditionals.h` that is causing these issues? The one included with IOS 26 SDK seems like it should still work as is.

To properly match the Clang extension all the `TARGET_OS_` macros should be defined for all targets and it should be possible to enable them even on non Apple OSes with `-fdefine-target-os-macros`.

And I don't think we need to define any `TARGET_CPU_` macros since Clang doesn't and we emulate Clang on MacOS.

brodo:
> Can you point to me to the version of TargetConditionals.h that is causing these issues? The one included with IOS 26 SDK seems like it should still work as is.

On macOS 16.24.1 with the newest Xcode it is at `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/TargetConditionals.h`

You can also view it [here](https://codeberg.org/brodo/xcode-frameworks/src/branch/main/include/TargetConditionals.h).

> To properly match the Clang extension all the TARGET_OS_ macros should be defined for all targets and it should be possible to enable them even on non Apple OSes with -fdefine-target-os-macros.

I can implement this tomorrow.

> And I don't think we need to define any TARGET_CPU_ macros since Clang doesn't and we emulate Clang on MacOS.

Clang implements two special functions:` __is_target_arch` and `__is_target_os`, so it does not need the defines. Implementing these functions in Aro would also be an option.

ehaas:
> Clang implements two special functions: __is_target_arch and __is_target_os, so it does not need the defines. Implementing these functions in Aro would also be an option.

Aro has those functions; they aren't heavily tested or used though so it's possible they don't work exactly how clang's do.

brodo:
Ok, so IMHO the best solution would be to fix any differences between Aro's and clang's version. My plan is to first port Clangs tests for __is_target_arch and __is_target_os to Aro. Here is a list:

- https://github.com/llvm/llvm-project/blob/main/clang/test/Preprocessor/is_target.c
- https://github.com/llvm/llvm-project/blob/main/clang/test/Preprocessor/is_target_os_darwin.c
- https://github.com/llvm/llvm-project/blob/main/clang/test/Preprocessor/is_target_unknown.c

brodo:
I'm closing this because I've 90% of a PR done which makes the clang tests run under Aro. New PR comes tomorrow!

PR #993, opened 2026-04-13
Fix regression in checking if qt is a pointer

### LINKED ISSUE #989: Regression: `nullability specifier cannot be applied to non-pointer type`
I [opened an issue](https://codeberg.org/ziglang/translate-c/issues/315) for `translate-c` because I noticed an error mentioned in the title when `cImport` failed, but I think `arocc` is the bottom of the stack?

In short, this is valid C:
```c
void f(int [_Nullable 2]);
```

I _think_ the regression was introduced in [`79d66eee`](https://github.com/Vexu/arocc/commit/79d66eeec6e926cd2406b995a1a863e3e9388e8f), arrays aren't considered a pointer for this specific case, but they should be.

The same check _is_ being made for `restrict` just [~30LOC below](https://github.com/Vexu/arocc/blob/5f5a050569a95ecc40a426f0c3666ae7ef987ede/src/aro/TypeStore.zig#L2873-L2878):
```zig
switch (qt.base(b.parser.comp).type) {
    .array, .pointer => result_qt.restrict = true,
    else => {
        try b.parser.err(restrict_tok, .restrict_non_pointer, .{qt});
    },
}
```

This fixes the error, but I am not nearly qualified to judge if it even makes sense:
```diff
diff --git a/src/aro/TypeStore.zig b/src/aro/TypeStore.zig
index 70b48ba..b0b8b39 100644
--- a/src/aro/TypeStore.zig
+++ b/src/aro/TypeStore.zig
@@ -2841,7 +2841,10 @@ pub const Builder = struct {
         }
 
         // We can't use `qt.isPointer()` because `qt` might contain a `.declarator_combine`.
-        const is_pointer = qt.isAutoType() or qt.isC23Auto() or qt.base(b.parser.comp).type == .pointer;
+        const is_pointer = qt.isAutoType() or qt.isC23Auto() or switch (qt.base(b.parser.comp).type) {
+            .array, .pointer => true,
+            else => false,
+        };
 
         if (b.unaligned != null and !is_pointer) {
             result_qt = (try b.parser.comp.type_store.put(gpa, .{ .attributed = .{
```

If it's reasonable, I can open up a PR.
I'm not sure how to add this to the test suite, though. 
I tested this by adding the function signature to the `test/cases/nullability.c` integration test, but it shouldn't result in an error so I'm not sure if it's the right approach. 

--- 
*Disclaimer: I used LLMs to find the root cause, but the issue was stumbled upon by importing a real header, namely from android's NDK.*

mark2185:
Nullability annotations on array parameters now work again.

Please point me to where I should add tests for this.

Closes #989.

### DIFF
--- a/src/aro/TypeStore.zig
+++ b/src/aro/TypeStore.zig
@@ -2841,7 +2841,10 @@ pub const Builder = struct {
         }
 
         // We can't use `qt.isPointer()` because `qt` might contain a `.declarator_combine`.
-        const is_pointer = qt.isAutoType() or qt.isC23Auto() or qt.base(b.parser.comp).type == .pointer;
+        const is_pointer = qt.isAutoType() or qt.isC23Auto() or switch (qt.base(b.parser.comp).type) {
+            .array, .pointer => true,
+            else => false,
+        };
 
         if (b.unaligned != null and !is_pointer) {
             result_qt = (try b.parser.comp.type_store.put(gpa, .{ .attributed = .{
--- a/test/cases/ast/msvc attribute keywords.c
+++ b/test/cases/ast/msvc attribute keywords.c
@@ -27,7 +27,7 @@ function: 'attributed(kr (...) *int)'
  attr: calling_convention cc: stdcall
  name: bar
 
-function: 'fn (decayed *[]attributed(int), decayed *attributed([]int)) int'
+function: 'fn (decayed *[]attributed(int), decayed *[]int) int'
  name: baz
 
 function: 'fn (fn_ptr: *fn () void) void'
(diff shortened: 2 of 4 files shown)


Vexu (approved):
> I tested this by adding the function signature to the `test/cases/nullability.c` integration test, but it shouldn't result in an error so I'm not sure if it's the right approach.

It is the right approach, those syntax tests are meant test both correct and incorrect syntax.

mark2185:
I've added the tests and ran `zig build test` with `zig` `0.16.0`, passes without any failures.

PR #995, opened 2026-04-17
Parser: make condExpr work properly in const contexts

ehaas:
If the condition and the selected branch of a condExpr are both statically known, make the condExpr have that statically known value. Also fixes parsing of binary conditional expressions (previous issues were masked by casting the result to void in `binary expressions.c` 

I believe this should fix https://codeberg.org/ziglang/translate-c/issues/322


MasonRemaley:
Confirmed on my end!

Vexu (approved):

PR #999, opened 2026-04-26
Parser: fix pointer index post increment

archaistvolts:
closes [330](https://codeberg.org/ziglang/translate-c/issues/330) by applying suggested patch from Vexu

removes unExpr/suffixExpr casts

### DIFF
--- a/test/cases/pointer index post increment.c
+++ b/test/cases/pointer index post increment.c
@@ -1,0 +1,9 @@
+void foo(void) {
+    unsigned short *ptr;
+    ptr[0]++;
+}
+
+/** manifest:
+syntax
+args = --target=x86_64-linux
+*/
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -8732,7 +8732,6 @@ fn unExpr(p: *Parser) Error!?Result {
             if (operand.qt.get(p.comp, .pointer)) |pointer| {
                 try p.checkPtrArithmeticAllowed(pointer, p.tok_i, operand.node);
             }
-            try operand.usualUnaryConversion(p, tok);
 
             if (operand.val.is(.int, p.comp) or operand.val.is(.int, p.comp)) {
                 if (try operand.val.add(operand.val, .one, operand.qt, p.comp))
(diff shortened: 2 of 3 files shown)


archaistvolts:
I'm not sure about the test case.  Please advise if any more changes are needed.  I ran test-integration and it shows a new case, 406 passed.  But I'm not sure how to verify this is working beyond that.

Before this patch an assertion on src/aro/CodeGen.zig:880 was failing.  With this patch it fails differently:
```console
$ cat /tmp/tmp.c
void foo(void) {
    unsigned short values1[10];
    values1[0]++;
    unsigned short *values;
    values[0]++;
}
$ zig-out/bin/arocc /tmp/tmp.c
fatal error: TODO CodeGen.genLval array_access_expr
```

test/cases/pointer index post increment.c
```diff
@@ -0,0 +1,8 @@
+void foo(void) {
+    unsigned short *ptr;
+    ptr[0]++;
```
Vexu, line 3: You could detect your change by adding the result of `--verbose-ast` on this test case to `test/cases/ast`. Note that you'll need to specify a `-target`.

Vexu (approved):

test/cases/pointer index post increment.c
archaistvolts: Ive added to the commit the result of
```console
zig build run -- test/cases/pointer\ index\ post\ increment.c --verbose-ast --target=native-native-musl > test/cases/ast/pointer\ index\ post\ increment.c
```

EDIT: updated command

archaistvolts: Result is different and now +1 ubuntu passing CI jobs :sweat_smile:  So I must need a different target?

Vexu: You need to add the target to the test manifest `args = -target x86_64-linux`.

archaistvolts: Ah, thanks.  I've added the target args to the new test/cases file.  All CI seem to be passing now. :slightly_smiling_face:

archaistvolts:
Thanks for the guidance @Vexu :heart:

PR #1000, opened 2026-05-02
init support for including nested darwin frameworks

kristoff-it:
This PR is for progressing towards solving https://codeberg.org/ziglang/translate-c/issues/343

The problem shown in the issue linked above is that when including a header file from CoreServices, the resolution chain at some point hits `#include <AE/AE.h>`, which is a framework nested inside of CoreServices.

Ignore below, see latest comments for the new approach.
> My initial understanding is that there are two approaches to solving this:
> 
> 1. Require users to provide `-framework CoreServices` when invoking arocc and append to the list of import search path also "CoreServices" (after resolving it based off one of the initial paths).
> 2. Detect when an include statement is found in a header file that belongs to a given framework, and extend the normal search process by also searching within potentially nested frameworks.
> 
> In this first commit I've attempted the first approach just to see if I could make some progress (also was the easier thing to implement) but I believe there are significant semantic differences between the two approaches and I have no idea if this is the right path or not.
> 
> In any case, this PR, when tested against the repro referenced in the issue linked above results in succesfully resolving AE.h to then fail when hitting an apple block definition. 
> 
> In order to test this it's also necessary to change translate-c to make Translator append the right `-framework` argument to the arocc invocation (and similarly the repro itself also needs to be updated).
>

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -215,6 +215,9 @@ pub fn deinit(comp: *Compilation) void {
         pragma.deinit(pragma, comp);
     }
     for (comp.sources.values()) |source| {
+        if (source.umbrella_framework_path) |path| {
+            assert(path.ptr == source.path.ptr);
+        }
         gpa.free(source.path);
         gpa.free(source.buf);
         gpa.free(source.splice_locs);
@@ -2086,14 +2089,14 @@ pub fn hasInclude(
     include_type: IncludeType,
     /// __has_include vs __has_include_next
     which: WhichInclude,
-    umbrella_framework: ?[]const u8,
     opt_dep_file: ?*DepFile,
 ) Compilation.Error!bool {
-    if (try FindInclude.run(comp, filename, umbrella_framework, include_type, switch (which) {
-        .next => .{ .only_search_after_dir = comp.getSource(includer_token_source).path },
+    const includer_source = comp.getSource(includer_token_source);
+    if (try FindInclude.run(comp, filename, includer_source, include_type, switch (which) {
+        .next => .{ .only_search_after_dir = includer_source.path },
         .first => switch (include_type) {
             .cli => unreachable,
-            .quotes => .{ .allow_same_dir = comp.getSource(includer_token_source).path },
+            .quotes => .{ .allow_same_dir = includer_source.path },
             .angle_brackets => .only_search,
         },
     })) |found| {
@@ -2117,14 +2120,12 @@ const FindInclude = struct {
         source: Source.Id,
         kind: Source.Kind,
         used_ms_search_rule: bool,
-        /// Allocated using the compilation arena allocator.
-        umbrella_framework: ?[]const u8 = null,
     };
 
     fn run(
         comp: *Compilation,
         include_path: []const u8,
-        umbrella_framework: ?[]const u8,
+        includer_source: Source,
         include_type: IncludeType,
         search_strat: union(enum) {
             allow_same_dir: []const u8,
@@ -2180,11 +2181,8 @@ const FindInclude = struct {
             if (try find.checkMsCwdIncludeDir(source_id)) |res| return res;
         }
 
-        if (umbrella_framework) |uf| {
-            if (try find.checkSubframeworkDir(uf, switch (include_type) {
-                .quotes, .cli => .user,
-                .angle_brackets => .system,
-            })) |res| return res;
+        if (includer_source.umbrella_framework_path) |umbrella_framework_path| {
+            if (try find.checkSubframeworkDir(umbrella_framework_path, includer_source.kind)) |res| return res;
         }
 
         return null;
@@ -2207,7 +2205,7 @@ const FindInclude = struct {
     fn checkFrameworkDir(find: *FindInclude, framework_dir: []const u8, kind: Source.Kind) Allocator.Error!?Result {
         // For an include like 'Foo/Bar.h', search in '<framework_dir>/Foo.framework/Headers/Bar.h'.
         const framework_name: []const u8, const header_sub_path: []const u8 = f: {
-            const i = std.mem.indexOfScalar(u8, find.include_path, '/') orelse return null;
+            const i = mem.findScalar(u8, find.include_path, '/') orelse return null;
             break :f .{ find.include_path[0..i], find.include_path[i + 1 ..] };
         };
         var bfa_buf: [path_buf_stack_limit]u8 = undefined;
@@ -2216,27 +2214,23 @@ const FindInclude = struct {
 
         // If the search succeeds, we consider the curent framework the umbrella framework for all
         // downstream includes.
-        const umbrella_framework = try std.fmt.allocPrint(bfa, "{f}", .{std.fs.path.fmtJoin(&.{
-            framework_dir,
-            try std.fmt.allocPrint(bfa, "{s}.framework", .{framework_name}),
-        })});
-        defer bfa.free(umbrella_framework);
-        std.debug.print("checking = '{s}'\n", .{umbrella_framework});
+        const framework_lookup = try std.fmt.allocPrint(bfa, "{s}.framework", .{framework_name});
+        defer bfa.free(framework_lookup);
 
-        var res = try find.check(&.{
-            umbrella_framework,
+        const res = try find.check(&.{
+            framework_dir,
+            framework_lookup,
             "Headers",
             header_sub_path,
         }, kind, false) orelse return null;
-
-        find.comp.normalizePath(umbrella_framework);
-        res.umbrella_framework = try find.comp.arena.dupe(u8, umbrella_framework);
+        const new_source = &find.comp.sources.values()[@intFromEnum(res.source.index)];
+        new_source.umbrella_framework_path = new_source.path[0 .. framework_dir.len + 1 + framework_lookup.len];
         return res;
     }
 
-    fn checkSubframeworkDir(find: *FindInclude, umbrella_framework: []const u8, kind: Source.Kind) Allocator.Error!?Result {
+    fn checkSubframeworkDir(find: *FindInclude, umbrella_framework_path: []const u8, kind: Source.Kind) Allocator.Error!?Result {
         const framework_name: []const u8, const header_sub_path: []const u8 = f: {
-            const i = std.mem.indexOfScalar(u8, find.include_path, '/') orelse return null;
+            const i = mem.findScalar(u8, find.include_path, '/') orelse return null;
             break :f .{ find.include_path[0..i], find.include_path[i + 1 ..] };
         };
         var bfa_buf: [path_buf_stack_limit]u8 = undefined;
@@ -2245,20 +2239,20 @@ const FindInclude = struct {
         const framework_lookup = try std.fmt.allocPrint(bfa, "{s}.framework", .{framework_name});
         defer bfa.free(framework_lookup);
 
-        var result = try find.check(&.{
-            umbrella_framework,
+        const res = try find.check(&.{
+            umbrella_framework_path,
             "Frameworks",
             framework_lookup,
             "Headers",
             header_sub_path,
         }, kind, false) orelse return null;
-
         // Subframeworks are assumed to not be able to contain other
         // subframeworks (i.e. they can't be umbrella frameworks), but they
         // can reference one another, meaning that we keep the same
         // umbrella framework.
-        result.umbrella_framework = umbrella_framework;
-        return result;
+        const new_source = &find.comp.sources.values()[@intFromEnum(res.source.index)];
+        new_source.umbrella_framework_path = new_source.path[0..umbrella_framework_path.len];
+        return res;
     }
 
     fn check(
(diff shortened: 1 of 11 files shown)


kristoff-it:
Testing by invoking clang without explicitly linking to the framework seems to suggest that the compiler should be able to resolve AE/AE.h without relying on `-framework`, but I'd like to receive an expert opinion on this.

src/aro/Compilation.zig
```diff
@@ -2156,6 +2160,14 @@ const FindInclude = struct {
                 find.wait_for = std.fs.path.dirname(other_file);
             },
         }
+
+        if (umbrella_framework) |uf| {
+            if (try find.checkSubframeworkDir(uf, switch (include_type) {
```
kristoff-it: I'm not sure if this is the correct approach or if the authoritative information is the kind of import path that we originally used to get into this framework. If this second option is the right one, then a new field needs to be added to Preprocessor to keep track of that.

src/aro/Compilation.zig
```diff
@@ -2198,15 +2212,51 @@ const FindInclude = struct {
...
+        std.debug.print("checking = '{s}'\n", .{umbrella_framework});
+
+        var res = try find.check(&.{
+            umbrella_framework,
+            "Headers",
+            header_sub_path,
+        }, kind, false) orelse return null;
+
+        find.comp.normalizePath(umbrella_framework);
+        res.umbrella_framework = try find.comp.arena.dupe(u8, umbrella_framework);
```
kristoff-it: In theory it should be possible to avoid this dupe, but it would require changing significantly the current code structure to pull up the buffer that contains this data inside `Preprocessor.include`, as that's where the recursion happens. (see later comment for more info on this)

src/aro/Preprocessor.zig
```diff
@@ -3585,22 +3588,24 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
...
     if (pp.verbose) {
-        pp.verboseLog(first, "include file {s}", .{new_source.path});
+        pp.verboseLog(first, "include file {s}", .{match.source.path});
     }
 
     const token_state = pp.getTokenState();
-    try pp.addIncludeStart(new_source.id);
-    const eof = pp.preprocessExtra(new_source) catch |er| switch (er) {
+    try pp.addIncludeStart(match.source.id);
+    pp.umbrella_framework = match.umbrella_framework;
```
kristoff-it: This is the most debatable part of the PR, here's what's happening.

My understanding of how subframework search should work is that it's essentially an argument that should be passed down when recursing.

1. we start from a user source file (`umbrella_framework = null`)
2. we encounter the first `<CoreServices/CoreServices.h>`
3. we find the header file inside of a framework include path 
4. from now on `umbrella_framework` is set to the current framework 
5. as we analyze the header file and expand new includes, different values of `umbrella_framework` should be passed down the call stack as we recourse (the value should stay the same as long as we are within the same framework and be set to null when exiting the umbrella framework)
6. as we go up, we pop new values of `umbrella_framework` off the stack, recovering older ones to be used for subsequent includes

Originally I started adding `umbrella_framework` as an argument to functions, but it was very invasive, so I decided to stash the value inside of Preprocessor, and have 2 lines of code that manage its state to mimic the same behavior.

It might be less error-prone to keep it as a function argument, but I wanted first to have somebody double check my understanding and ask Vexu which style is preferred. 

I'll mark the other location where I manage this state.

src/aro/Preprocessor.zig
```diff
@@ -3585,22 +3588,24 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
...
+    const eof = pp.preprocessExtra(match.source) catch |er| switch (er) {
         error.StopPreprocessing => {
             for (pp.expansion_entries.items(.locs)[token_state.expansion_entries_len..]) |loc| TokenWithExpansionLocs.free(loc, gpa);
             pp.restoreTokenState(token_state);
             return;
         },
         else => |e| return e,
     };
-    try eof.checkMsEof(new_source, pp.comp);
+    pp.umbrella_framework = match.umbrella_framework;
```
kristoff-it: this mimics going up from the call stack

Vexu: I think you meant to set it to the previous value of itself or at least that is needed to pass this test case:

```sh
mkdir -p frameworks/Framework.framework/Headers
mkdir -p frameworks/Framework.framework/Frameworks/SubFrameworkA.framework/Headers
mkdir -p frameworks/Framework.framework/Frameworks/SubFrameworkB.framework/Headers

echo "#include <Framework/Framework.h>

_Static_assert(CONSTANT_A == 1, \"\");
_Static_assert(CONSTANT_B == 1, \"\");" > framework-test.c
echo "#include <SubFrameworkA/A.h>
#include <SubFrameworkB/B.h>" > frameworks/Framework.framework/Headers/Framework.h
echo "#define CONSTANT_A 1" > frameworks/Framework.framework/Frameworks/SubFrameworkA.framework/Headers/A.h
echo "#define CONSTANT_B 1" > frameworks/Framework.framework/Frameworks/SubFrameworkB.framework/Headers/B.h
```

```diff
diff --git a/src/aro/Preprocessor.zig b/src/aro/Preprocessor.zig
index 4a771188..a498d865 100644
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -3594,6 +3594,7 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
 
     const token_state = pp.getTokenState();
     try pp.addIncludeStart(match.source.id);
+    const old_framework = pp.umbrella_framework;
     pp.umbrella_framework = match.umbrella_framework;
     const eof = pp.preprocessExtra(match.source) catch |er| switch (er) {
         error.StopPreprocessing => {
@@ -3603,7 +3604,7 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
         },
         else => |e| return e,
     };
-    pp.umbrella_framework = match.umbrella_framework;
+    pp.umbrella_framework = old_framework;
     try eof.checkMsEof(match.source, pp.comp);
     if (pp.preserve_whitespace and pp.tokens.items(.id)[pp.tokens.len - 1] != .nl) {
         try pp.addToken(.{ .id = .nl, .loc = .{
```

src/aro/Preprocessor.zig
Vexu: Did you try to add a framework flag to `Source`? That should make it possible to implement this without adding any state to the prerprocessor.

src/aro/Compilation.zig
```diff
@@ -2198,15 +2213,51 @@ const FindInclude = struct {
...
+        var bfa_buf: [path_buf_stack_limit]u8 = undefined;
+        var bfa_state: std.heap.BufferFirstAllocator = .init(&bfa_buf, find.comp.gpa);
+        const bfa = bfa_state.allocator();
         const framework_lookup = try std.fmt.allocPrint(bfa, "{s}.framework", .{framework_name});
         defer bfa.free(framework_lookup);
+
+        // A match into a subframework resets the active umbrella framework
+        // back to null (by not setting the relative Result field) because
+        // we assume that subframeworks don't see one another within the
+        // same umbrella framework.
```
Vexu: This doesn't seem to be true:
```sh
mkdir -p frameworks/Framework.framework/Headers
mkdir -p frameworks/Framework.framework/Frameworks/SubFrameworkA.framework/Headers
mkdir -p frameworks/Framework.framework/Frameworks/SubFrameworkB.framework/Headers

echo "#include <Framework/Framework.h>

_Static_assert(CONSTANT_A == 1, \"\");
_Static_assert(CONSTANT_B == 1, \"\");" > framework-test.c
echo "#include <SubFrameworkA/A.h>
#include <SubFrameworkB/B.h>" > frameworks/Framework.framework/Headers/Framework.h
echo "#define CONSTANT_A 1

#include <SubFrameworkB/B.h>

_Static_assert(CONSTANT_B == 1, \"\");" > frameworks/Framework.framework/Frameworks/SubFrameworkA.framework/Headers/A.h
echo "#define CONSTANT_B 1" > frameworks/Framework.framework/Frameworks/SubFrameworkB.framework/Headers/B.h

```

src/aro/Preprocessor.zig
Vexu: I went ahead and tried this version:

<details>
  <summary>umbrealla_framework_path in Source diff</summary>




```diff
diff --git a/src/aro/Compilation.zig b/src/aro/Compilation.zig
index 33c348fe..8cfe21e7 100644
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -2086,14 +2086,14 @@ pub fn hasInclude(
     include_type: IncludeType,
     /// __has_include vs __has_include_next
     which: WhichInclude,
-    umbrella_framework: ?[]const u8,
     opt_dep_file: ?*DepFile,
 ) Compilation.Error!bool {
-    if (try FindInclude.run(comp, filename, umbrella_framework, include_type, switch (which) {
-        .next => .{ .only_search_after_dir = comp.getSource(includer_token_source).path },
+    const includer_source = comp.getSource(includer_token_source);
+    if (try FindInclude.run(comp, filename, includer_source, include_type, switch (which) {
+        .next => .{ .only_search_after_dir = includer_source.path },
         .first => switch (include_type) {
             .cli => unreachable,
-            .quotes => .{ .allow_same_dir = comp.getSource(includer_token_source).path },
+            .quotes => .{ .allow_same_dir = includer_source.path },
             .angle_brackets => .only_search,
         },
     })) |found| {
@@ -2117,14 +2117,12 @@ const FindInclude = struct {
         source: Source.Id,
         kind: Source.Kind,
         used_ms_search_rule: bool,
-        /// Allocated using the compilation arena allocator.
-        umbrella_framework: ?[]const u8 = null,
     };
 
     fn run(
         comp: *Compilation,
         include_path: []const u8,
-        umbrella_framework: ?[]const u8,
+        includer_source: Source,
         include_type: IncludeType,
         search_strat: union(enum) {
             allow_same_dir: []const u8,
@@ -2180,11 +2178,8 @@ const FindInclude = struct {
             if (try find.checkMsCwdIncludeDir(source_id)) |res| return res;
         }
 
-        if (umbrella_framework) |uf| {
-            if (try find.checkSubframeworkDir(uf, switch (include_type) {
-                .quotes, .cli => .user,
-                .angle_brackets => .system,
-            })) |res| return res;
+        if (includer_source.umbrella_framework_path) |umbrella_framework_path| {
+            if (try find.checkSubframeworkDir(umbrella_framework_path, includer_source.kind)) |res| return res;
         }
 
         return null;
@@ -2207,7 +2202,7 @@ const FindInclude = struct {
     fn checkFrameworkDir(find: *FindInclude, framework_dir: []const u8, kind: Source.Kind) Allocator.Error!?Result {
         // For an include like 'Foo/Bar.h', search in '<framework_dir>/Foo.framework/Headers/Bar.h'.
         const framework_name: []const u8, const header_sub_path: []const u8 = f: {
-            const i = std.mem.indexOfScalar(u8, find.include_path, '/') orelse return null;
+            const i = mem.findScalar(u8, find.include_path, '/') orelse return null;
             break :f .{ find.include_path[0..i], find.include_path[i + 1 ..] };
         };
         var bfa_buf: [path_buf_stack_limit]u8 = undefined;
@@ -2216,27 +2211,23 @@ const FindInclude = struct {
 
         // If the search succeeds, we consider the curent framework the umbrella framework for all
         // downstream includes.
-        const umbrella_framework = try std.fmt.allocPrint(bfa, "{f}", .{std.fs.path.fmtJoin(&.{
-            framework_dir,
-            try std.fmt.allocPrint(bfa, "{s}.framework", .{framework_name}),
...skipping...
     }
 
-    if (pp.dep_file) |dep| try dep.addDependency(gpa, match.source.path);
+    if (pp.dep_file) |dep| try dep.addDependency(gpa, new_source.path);
     if (pp.verbose) {
-        pp.verboseLog(first, "include file {s}", .{match.source.path});
+        pp.verboseLog(first, "include file {s}", .{new_source.path});
     }
 
     const token_state = pp.getTokenState();
-    try pp.addIncludeStart(match.source.id);
-    pp.umbrella_framework = match.umbrella_framework;
-    const eof = pp.preprocessExtra(match.source) catch |er| switch (er) {
+    try pp.addIncludeStart(new_source.id);
+    const eof = pp.preprocessExtra(new_source) catch |er| switch (er) {
         error.StopPreprocessing => {
             for (pp.expansion_entries.items(.locs)[token_state.expansion_entries_len..]) |loc| TokenWithExpansionLocs.free(loc, gpa);
             pp.restoreTokenState(token_state);
@@ -3603,8 +3600,7 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
         },
         else => |e| return e,
     };
-    pp.umbrella_framework = match.umbrella_framework;
-    try eof.checkMsEof(match.source, pp.comp);
+    try eof.checkMsEof(new_source, pp.comp);
     if (pp.preserve_whitespace and pp.tokens.items(.id)[pp.tokens.len - 1] != .nl) {
         try pp.addToken(.{ .id = .nl, .loc = .{
             .id = tokenizer.source,
@@ -3761,8 +3757,7 @@ fn findIncludeFilenameToken(
     return filename_tok;
 }
 
-/// On success returns a new source and, if the new source is part of a macOS umbrella framework, a non-null framework path.
-fn findIncludeSource(pp: *Preprocessor, tokenizer: *Tokenizer, first: RawToken, which: Compilation.WhichInclude) !Compilation.FoundInclude {
+fn findIncludeSource(pp: *Preprocessor, tokenizer: *Tokenizer, first: RawToken, which: Compilation.WhichInclude) !Source {
     const filename_tok = try pp.findIncludeFilenameToken(first, tokenizer, .expect_nl_eof);
     defer TokenWithExpansionLocs.free(filename_tok.expansion_locs, pp.comp.gpa);
 
@@ -3781,7 +3776,7 @@ fn findIncludeSource(pp: *Preprocessor, tokenizer: *Tokenizer, first: RawToken,
         else => unreachable,
     };
 
-    return (try pp.comp.findInclude(filename, pp.umbrella_framework, first, include_type, which)) orelse
+    return (try pp.comp.findInclude(filename, first, include_type, which)) orelse
         return pp.fatalNotFound(filename_tok, filename);
 }
 
diff --git a/src/aro/Source.zig b/src/aro/Source.zig
index 28d80ce0..0d35c399 100644
--- a/src/aro/Source.zig
+++ b/src/aro/Source.zig
@@ -58,6 +58,9 @@ id: Id,
 splice_locs: []const u32,
 kind: Kind,
 
+// Path of the umbrella framework. Slice of `path`.
+umbrella_framework_path: ?[]const u8 = null,
+
 pub fn lineCol(source: Source, loc: Location) ExpandedLocation {
     var start: usize = 0;
     // find the start of the line which is either a newline or a splice
```




</details>

kristoff-it: clang is able to resolve these includes without any flag, I think we tested this during our last call (I'm confident of the fact, less so whether we tested it together or not)

src/aro/Preprocessor.zig
kristoff-it: oh damn yeah thanks

src/aro/Preprocessor.zig
kristoff-it: oh sorry I misread your comment, wrt adding state to Source, I avoided it because it seems that the umbrella framework state is not necessary anymore once preprocessing is done, do Sources only exist during preprocessing? in that case it might make no difference

kristoff-it:
I've addressed the PR review feedback, except for the change about where to store tracking info for `umbrella_framework`. @Vexu do you mind taking over from here? I think the work that remains is best done by you, as you're more familiar with the codebase as a whole.

Vexu:
Sure.

PR #1002, opened 2026-05-03
Improve clang compat in __is_target and add some more test cases

brodo:
This PR tries to align the __is_targets builtins closer to what clang does. The motivation for this was an error
in [`TargetConditionals.h`](https://codeberg.org/brodo/xcode-frameworks/src/branch/main/include/TargetConditionals.h) when compiling [dvui](https://github.com/david-vanderson/dvui) which I currently can't reproduce. I took the tests for the `__is_targets` builtins from the clang test suite and ported them. I found no incompatibilities that way, but I still kept the tests to make sure that we don't break anything in the future. 

I found several incompatibilities by looking at `TargetContionals.h` and the clang test suite:

- Target OSes like `ios12` are not supported. **This PR fixes that.**
- Zig does not support the `darwin` OS tag and stores it as `.macos`. In clang all Apple OSes are `darwin` but not all of them are Mac. This means we can't be 100% compatible.
- clang does not have `mccatalyst` as an OS tag, but it is supported by Zig. This means we can't be 100% compatible there either.

I've added some tests for OS version postfixes (ios12) and updated `parseOsName` to be prefix-based and `parseOS`
to extract the right version.

### DIFF
--- a/src/aro/Target.zig
+++ b/src/aro/Target.zig
@@ -878,16 +878,14 @@ test parseArchName {
 pub fn parseOsName(query: []const u8) ?Os.Tag {
     var buf: [64]u8 = undefined;
     const lower = toLower(query, &buf) orelse return null;
-    
-    return std.meta.stringToEnum(Os.Tag, lower) orelse std.StaticStringMap(Os.Tag).initComptime(.{
-        .{ "darwin", .macos },
-        .{ "macosx", .macos },
-        .{ "tvos", .tvos },
-        .{ "visionos", .visionos },
-        .{ "watchos", .watchos },
-        .{ "win32", .windows },
-        .{ "xros", .visionos },
-    }).get(lower) orelse return null;
+
+    return std.meta.stringToEnum(Os.Tag, lower) orelse
+        std.StaticStringMap(Os.Tag).initComptime(.{
+            .{ "darwin", .macos },
+            .{ "macosx", .macos },
+            .{ "win32", .windows },
+            .{ "xros", .visionos },
+        }).get(lower) orelse return null;
 }
 
 pub fn isOs(target: *const Target, query_str: []const u8) bool {
--- a/src/aro/Target.zig
+++ b/src/aro/Target.zig
@@ -878,7 +878,6 @@ test parseArchName {
 pub fn parseOsName(query: []const u8) ?Os.Tag {
     var buf: [64]u8 = undefined;
     const lower = toLower(query, &buf) orelse return null;
-
     return std.meta.stringToEnum(Os.Tag, lower) orelse
         std.StaticStringMap(Os.Tag).initComptime(.{
             .{ "darwin", .macos },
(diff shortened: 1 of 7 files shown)


src/aro/Target.zig
```diff
@@ -878,18 +867,26 @@ test parseArchName {
...
+        .{ .prefix = "driverkit", .tag = .driverkit },
+        .{ .prefix = "ios", .tag = .ios },
+        .{ .prefix = "macos", .tag = .macos },
+        .{ .prefix = "tvos", .tag = .tvos },
+        .{ .prefix = "visionos", .tag = .visionos },
+        .{ .prefix = "watchos", .tag = .watchos },
+        .{ .prefix = "win32", .tag = .windows },
+        .{ .prefix = "xros", .tag = .visionos },
+    }) |entry| {
+        if (mem.startsWith(u8, lower, entry.prefix)) return entry.tag;
```
Vexu: You should not need `startsWith` here, did you run into some specific case?

src/aro/Target.zig
```diff
@@ -1690,6 +1687,15 @@ pub fn parseOs(result: *std.Target.Query, text: []const u8, version_string: ?*[]
...
         },
     } else .{ checkOs(text) orelse return error.UnknownOs, "" };
+
+    var version_buf: [64]u8 = undefined;
+    const version_text = if (tag.isDarwin() and
+        raw_version_text.len > 0 and
+        !mem.containsAtLeastScalar(u8, raw_version_text, 1, '.'))
+        std.fmt.bufPrint(&version_buf, "{s}.0.0", .{raw_version_text}) catch return error.InvalidOsVersion
+    else
+        raw_version_text;
```
Vexu: Clang allows `linux6` so instead of special casing Darwin OSes  you should copy `std.Target.Query.parseVersion` here and modify it to allow omitting the minor version.

src/aro/Target.zig
brodo: Yes, `__is_target_os` uses `Target.isOs` and that uses `Target.parseOsName` and not `Target.parseOs`. So `__is_target_os(ios17)` always fails. There are tests for that.

Vexu: `Target.isOs` should then be removed in favor of using `parseOs`.

brodo:
@Vexu changes are in!

src/aro/Target.zig
```diff
@@ -878,23 +867,31 @@ test parseArchName {
...
-            .{ "darwin", .macos },
-            .{ "macosx", .macos },
-            .{ "win32", .windows },
-            .{ "xros", .visionos },
-        }).get(lower) orelse return null;
+    return std.meta.stringToEnum(Os.Tag, lower) orelse std.StaticStringMap(Os.Tag).initComptime(.{
+        .{ "darwin", .macos },
+        .{ "driverkit", .driverkit },
+        .{ "ios", .ios },
+        .{ "macos", .macos },
```
Vexu: These identical ones can be reverted now since they are already checked with the `stringToEnum`.

Vexu (approved):

brodo:
@Vexu I deleted it.

PR #1004, opened 2026-05-13
Parser: Add support for __declspec(thread) for thread local variables

Finistere:
Add support for __declspec(thread) for thread local variables.


Finistere:
Sorry for the noise.

PR #1006, opened 2026-05-13
Parser: Add support for GCC __thread

### LINKED ISSUE #1005: Parser: GNU __thread is not parsed correctly
Hi! Today, we use `translate-c` for dpdk for which we need a few patches. One of those involves GNU's `__thread`. My understanding of it, after some investigation with the help of AI, is that Aro doesn't parse it correctly as thread local. I've made a commit: https://github.com/Vexu/arocc/commit/99d8f341ac362575961f655c04f1e0569e2d9091 that solves the issue for me. If you get a chance, I'd appreciate if you could let me know whether that's the right fix or if I should do differently.

Finistere:
Adding support for GCC __thread with pedantic warning if before `extern`/`static`.
I've imitated `__Thread_local` keyword for `nextStmt` and the `"keyword"` test.

Closes: #1005

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -11,6 +11,7 @@ const char_info = @import("char_info.zig");
 const Compilation = @import("Compilation.zig");
 const Diagnostics = @import("Diagnostics.zig");
 const InitList = @import("InitList.zig");
+pub const Diagnostic = @import("Parser/Diagnostic.zig");
 const Preprocessor = @import("Preprocessor.zig");
 const record_layout = @import("record_layout.zig");
 const Source = @import("Source.zig");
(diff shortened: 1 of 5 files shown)


Finistere:
CI fails because of Zig update, I can also take a look at this in a separate PR

src/aro/Parser.zig
```diff
@@ -1752,7 +1753,12 @@ fn storageClassSpec(p: *Parser, d: *DeclSpec) Error!bool {
                 }
                 if (d.thread_local != null) {
                     switch (id) {
-                        .keyword_extern, .keyword_static => {},
+                        .keyword_extern, .keyword_static => {
+                            const thread_tok = d.thread_local.?;
```
Vexu: This could use a capture.

Vexu (approved):
I did the Zig update, this just needs a rebase.

src/aro/Parser.zig
Finistere: Ah yes, thanks.

PR #1009, opened 2026-05-15
Parser: Add support for __builtin_elementwise_(fshl|fshr)

Finistere:
Adding support for `__builtin_elementwise_fshl` and `__builtin_elementwise_fshr` that were added in CLang 22 

| Builtin | Description | Types |
|---|---|---|
| T __builtin_elementwise_fshl(T x, T y, T z) | Perform a funnel shift left. Concatenate x and y (x is the most significant bits of the wide value), shift the combined value left by z modulo the bit width of the original arguments, and extract the most significant bits to produce a result the same size as the original arguments. | integer types |
| T __builtin_elementwise_fshr(T x, T y, T z) | Perform a funnel shift right. Concatenate x and y (x is the most significant bits of the wide value), shift the combined value right by z modulo the bit width of the original arguments, and extract the least significant bits to produce a result the same size as the original arguments. | integer types |

https://releases.llvm.org/22.1.0/tools/clang/docs/LanguageExtensions.html#vector-builtins


Finistere:
I didn't check after rebase that those functions were already added. I am working on an earlier version of aro in my project.

PR #1010, opened 2026-05-18
Pragma, Preprocessor: add "beforeInclude" hook, extend "once"

vancluever:
This adds the `beforeInclude` pragma hook, which fires before a file is included. Its current and only implementation is in the "once" pragma, which has been extended to use it to ensure that `#pragma once` causes future inclusions of a file to be skipped outright, versus processed until the pragma is hit again.

This prevents scenarios where the directive has not been placed at the top of the file. This is a practice that is carried out in at least the GTK headers, and the current control flow causes inner-inclusion guards
to be hit when they should not be.

Note that since this is a deep-preprocessor event, it has not been added to the compilation-scoped `PragmaEvent` set, and just implemented as a hook local in the include function. If more event hooks get added in the future, it might be a good idea to review the event system to allow for scope-agnostic event types and payloads.

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -3553,6 +3553,39 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
         if (pp.defines.contains(guard)) return;
     }
 
+    // Run any beforeInclude pragma hooks.
+    //
+    // NOTE: This is currently the only deep-preprocessor hook other than the
+    // actual token handler itself. If more are added, an event hook system
+    // similar to the compilation-scoped one should be considered with a
+    // payload structure that can handle context-relevant event data (e.g.,
+    // here, the source file data is needed).
+    {
+        var pragma_handler_it = pp.comp.pragma_handlers.iterator();
+        while (pragma_handler_it.next()) |pragma_handler| {
+            const handler_pragma_name = pragma_handler.key_ptr.*;
+            const handler_pragma = pragma_handler.value_ptr.*;
+            if (handler_pragma.beforeInclude) |func| {
+                func(handler_pragma, pp, new_source) catch |handler_err| {
+                    switch (handler_err) {
+                        error.SkipInclude => {
+                            if (pp.verbose) {
+                                pp.verboseLog(
+                                    first,
+                                    "skipping include file {s} under direction of \"{s}\" pragma",
+                                    .{ new_source.path, handler_pragma_name },
+                                );
+                            }
+
+                            return;
+                        },
+                        else => |e| return e,
+                    }
+                };
+            }
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


vancluever:
This was discovered during our migration of Ghostty to Zig 0.16.0 (current PR is ghostty-org/ghostty#12726). It appears that the `#pragma once` in `gdk/version/gdkversionmacros.h` has been placed _after_ the guards in the file, which causes some deeper includes to fail, particularly the include path that starts at `gtk/gtkaboutdialog.h` in `gtk/gtk.h`, by which point the GDK includes have been walked enough to undef `__GDK_H_INSIDE__` (in addition to `__GTK_CSS_H_INSIDE__` through the `gtk/css/gtkcss.h` path).

Managed to isolate the case to what is seen in the test here. A reproduction in clang seems to be fine, so I'm guessing this is generally the expected behavior.

Let me know if this needs anything!

ehaas:
The test looks good. The fix works as a targeted solution to the problem; it would be nice if the pragma system itself could express this sort of thing (as it stands, a consumer of aro can provide their own pragmas, but they would not be able to provide a custom pragma that behaved the same as `#pragma once` since its behavior is now hardcoded into the preprocessor itself.)

I'm imagining something like another `PragmaEvent` like "before_include" or "should_include" or something like that; the default implementation would just return true but then custom pragmas could override it to allow deciding whether the file should be included.

I don't think implementing that is a blocker, especially since this issue is preventing use on a popular real-world library, just a thought for how we could improve the pragma system.

vancluever:
@ehaas I think this would be a trivial thing (or at least not a complicated thing) to implement to start out. I'll update the PR in a bit!

vancluever:
@ehaas PR updated - commentary in the commit and PR description!

ehaas (approved):

src/aro/Preprocessor.zig
```diff
@@ -3550,6 +3553,39 @@ fn include(pp: *Preprocessor, tokenizer: *Tokenizer, which: Compilation.WhichInc
...
+                            }
+
+                            return;
+                        },
+                        else => |e| return e,
+                    }
+                };
+            }
+        }
+    }
```
Vexu:
```suggestion
    for (pp.comp.pragma_handlers.keys(), pp.comp.pragma_handlers.values()) |handler_pragma_name, handler_pragma| {
        if (handler_pragma.beforeInclude) |func| {
            func(handler_pragma, pp, new_source) catch |handler_err| {
                switch (handler_err) {
                    error.SkipInclude => {
                        if (pp.verbose) {
                            pp.verboseLog(
                                first,
                                "skipping include file {s} under direction of \"{s}\" pragma",
                                .{ new_source.path, handler_pragma_name },
                            );
                        }

                        return;
                    },
                    else => |e| return e,
                }
            };
        }
    }
```

Vexu (approved):

vancluever:
Thanks to the both of you!

PR #1011, opened 2026-05-22
Preprocessor fixes

### LINKED ISSUE #919: Macro expansion difference vs other preprocessors
Found this example here: https://marc.info/?l=boost&m=118835769257658&w=2

```c
#define A(m) m( B(f)
#define B(x) A(x)

#define C(x) < x >

A(C) ) *
```

clang, GCC, MSVC, and tinycc output:
```
< f( B(f) > *
```

arocc output:
```
<  A(f)  > *
```

The post says that our output is correct; and that "Very very few preprocessors get this completely right". Not sure if this is a bug or not.

### LINKED ISSUE #949: Placemarker token at the start of preprocessor expression
```c
#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)
# define FLEXARRAY
#else
# define FLEXARRAY 1
#endif

#if FLEXARRAY+1 > 1
# define SZ_KEYINFO_0   offsetof(KeyInfo,aColl)
#else
# define SZ_KEYINFO_0   sizeof(KeyInfo)
#endif
```

```sh-session
$ zig build run -- a.c
<scratch space>:0:1: error: invalid token at start of a preprocessor expression

^
1 error generated.
```

I fixed it by adding a call to `removePlacemarkers` but I suspect that this also affects pragmas and incudes:
```diff
diff --git a/src/aro/Preprocessor.zig b/src/aro/Preprocessor.zig
index 57d7b5fb..85eb8c47 100644
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -1139,6 +1139,7 @@ fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) MacroError!bool {
         pp.expansion_source_loc = pp.top_expansion_buf.items[0].loc;
         pp.hideset.clearRetainingCapacity();
         try pp.expandMacroExhaustive(tokenizer, &pp.top_expansion_buf, 0, pp.top_expansion_buf.items.len, false, .expr);
+        removePlacemarkers(gpa, &pp.top_expansion_buf);
     }
     for (pp.top_expansion_buf.items) |tok| {
         if (tok.id == .macro_ws) continue;
```

ehaas:
Fixes a few preprocessor edge cases - see individual commits.

Closes #919
Closes #949


test/cases/expanded/rparen hideset in argument expansion.c
```diff
@@ -0,0 +1 @@
+<  f( B(f)  > *
```
ehaas, line 1: Note: our whitespace doesn't exactly match clang's output; not sure how much of an issue that is.

ehaas:
Regarding #949, I wasn't able to trigger the placemarker issue with pragma directives but I'm not sure if that's because we don't have a problem there or if I just wasn't able to find it.

Vexu (approved):

PR #1014, opened 2026-05-24
Compilation: Add ROPI/RWPI macros for ARM

brodo:
This allows users to invoke Aro with three new options

-fno-ropi: disables ROPI
-frwpi: enables RWPI
-fno-rwpi: disables RWPI

This en/disables the __ARM_ROPI and __ARM_RWPI macros.

@Vexu I just added the new options and booleans. I hope this is fine. The tests are ported from Clang.


src/aro/Compilation.zig
```diff
@@ -912,6 +784,14 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                 try define(w, "__ARM_FEATURE_RCPC");
             }
 
+            if (comp.code_gen_options.is_ropi) {
```
Vexu: Clang doesn't support `-fropi`/`-frwpi` on aarch64.

Vexu (commented):
This change doesn't necessarily need a test and definitely not four of them.

brodo:
> This change doesn't necessarily need a test and definitely not four of them.

Ok, I'll just keep one.

src/aro/Compilation.zig
brodo: I'll check if it does for other ARM architectures. I finally went ahead and cloned clang (it's so big...), so I can grep the codebase locally...

brodo:
@Vexu done now

Vexu:
Thanks!

PR #1016, opened 2026-05-24
Compilation: add ACLE macro to aarch64

brodo:
This also moves unconditional defines to the top of the switch branches. I also added the `__arm__` macros to `aarch64` because Clang does that too.


Vexu:
> I also added the `__arm__` macros to `aarch64` because Clang does that too.

It doesn't for me. The aarch64 version of `__ARM_ACLE` is also more complicated:

```c
#define __ARM_ACLE_VERSION(year,quarter,patch) (100 * (year) + 10 * (quarter) + (patch))
#define __ARM_ACLE 202420
```

brodo:
I was just looking at that: https://clang.llvm.org/doxygen/Basic_2Targets_2ARM_8cpp_source.html
I cloned llvm now and can see the definition...

brodo:
I've got some nushell scripts in place now which allow me to compare the Clang and the Aro output for a given target.
 I've added some more unconditional defines. The value of `ARM_ACLE` is always `202420` for aarch64 and `200` for everything else. Set in clang by `Basic/Targets/Arm64.cpp` and  `Basic/Targets/ARM.cpp` in respectively.

brodo:
@Vexu This PR is done now. I've got another PR in the pipeline with more ARM defines...

src/aro/Compilation.zig
```diff
@@ -825,6 +823,16 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
         },
         .aarch64, .aarch64_be => |arch| {
             try define(w, "__aarch64__");
+            try define(w, "__ARM_64BIT_STATE");
+            try define(w, "__ARM_ARCH_ISA_A64");
+            try w.writeAll("#define __ARM_ARCH 8\n");
```
Vexu: This could also be `9` with the `.v9a` feature.

src/aro/Compilation.zig
```diff
@@ -825,6 +823,16 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
+            try w.writeAll("#define __ARM_ARCH_PROFILE 'A'\n");
```
Vexu: This could also be `R` with the `.v8r` feature.

brodo: Really? `aarch` does not mean `A` profile?

Edit: yes, one can indeed call `clang` with an `aarch64` target and a `v8r` architecture.

Vexu: I'm no expert on ARM but that's what I get with `clang -E -target aarch64-linux -march=armv8r`.

brodo: I've just looked it up: Wikipedia says `aarch64` just means 64-bit ARM architecture. The 'A' is just for 'ARM' not for A profile. However, there is no 'M' processor which has 64-bit support, so we don't have to check that.

brodo:
The current version is not 100% compatible with clang, because clang does not support setting 'R' profile or 'v9' as part of the target triplet. E.g.:

` "" | clang -dM -E -target "aarch64v9a-apple-darwin" -` will result in `__ARM_ARCH 8`, not 9. If you set it explicitly using `-march`, it will work: `"" | clang -dM -E -target "aarch64-apple-darwin" -march="armv9a" -`. I think that's a clang bug.

brodo:
I also changed the order of the features so the newest ones come first. This is needed because some new features also imply the old ones and these are detected first then.

@Vexu rdy to merge if you think it's ok.

src/aro/Compilation.zig
```diff
@@ -822,16 +848,49 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
+                .{ .v8_5a, "8_5A" },
+                .{ .v8_4a, "8_4A" },
+                .{ .v8_3a, "8_3A" },
+                .{ .v8_2a, "8_2A" },
+                .{ .v8_1a, "8_1A" },
+                .{ .v8a, "8A" },
+                .{ .v8r, "8R" },
+            }) |fs| {
+                if (arm_features.isEnabled(@intFromEnum(fs[0]))) {
+                    try w.print("#define __ARM_ARCH_{s}__ 1\n", .{fs[1]});
```
Vexu, line 885: Clang and gcc don't define this macro but it's probably no harm to add.

brodo: I'm going to fix that in the next PR.

PR #1023, opened 2026-05-30
Frontend: add basic __mfp8 support

ehaas:
Adds basic frontend support for `__mfp8`. This is just basic parser support; right now it is not target-specific (this is an ARM / ACLE thing) and doesn't include things like FEAT_FP8 and I'm not sure whether the frontend needs to care about modality (E5M2 vs E4M3). It's modeled as a new kind of "opaque float" that can be copied but can't be coerced or casted to/from other types or used in arithmetic expressions. We will also need `neon_vector_type` attribute eventually.


src/aro/CodeGen.zig
```diff
@@ -158,6 +158,7 @@ fn genType(c: *CodeGen, qt: QualType) !Interner.Ref {
         .int => |int| .{ .int_ty = int.bits(c.comp) },
         .bit_int => |bit_int| .{ .int_ty = bit_int.bits },
         .float => |float| .{ .float_ty = float.bits(c.comp) },
+        .storage_float => |storage_float| .{ .int_ty = storage_float.bits() },
```
ehaas, line 161: This is just a placeholder, not sure what we should be using here.

src/aro/Tokenizer.zig
```diff
@@ -1109,6 +1112,7 @@ pub const Token = struct {
         // clang keywords
         .{ "__fp16", .keyword_fp16 },
         .{ "__bf16", .keyword_bf16 },
+        .{ "__mfp8", .keyword_mfp8 },
```
Vexu: `__mfp8` is an implicit typedef in Clang.

test/cases/__mfp8 invalid.c
```diff
@@ -0,0 +1,101 @@
...
+    take_int(a);
+    _Complex __mfp8 complex_mfp8;
+}
+
+
+/** manifest:
+syntax
+args = -std=c23 --target=aarch64-linux-gnu
+
+__mfp8 invalid.c:2:7: error: invalid operands to binary expression ('__mfp8' (aka '__mfp8') and '__mfp8' (aka '__mfp8'))
```
Vexu, line 57: >  `'__mfp8' (aka '__mfp8')`

I'll improve the type desugaring logic in a follow-up.

Vexu (approved):

PR #1028, opened 2026-06-01
Builtins: use generated ARM Neon builtins

Vexu:
With this I'm able to successfully include `arm_neon.h`.


vancluever:
@Vexu confirmed over here that this successfully translates for me as well with this and #1027 applied!

(Not too sure if #1027 was needed but I figured I'd include it since it came from the MFP8 work.)

I will do the necessary backporting and what not over the next few days (so that we can use this in Ghostty and Zig 0.16.0) and test some more and report if anything else is an issue. Thank you and @ehaas very much for your quick work on this!!!

PR #1029, opened 2026-06-01
Update to new std.lang.Type; Update to Zig 0.17.0-dev.657+2faf8debf

der-teufel-programming:
Merging this PR should probably wait until a Zig version of at least 657 is available to download

PR #1032, opened 2026-06-02
build: update zig version

brodo:
Updates Zig to newest master. This is already used in CI.

Aro currently does not build with the Zig version in the Zon file.


brodo:
@Vexu I've got a PR with 13 more ARM defines ready based on this commit.

Vexu:
This should have been changed in a57886d20e3056bb0a227e090fa63896b7b4ffa4 but there will be new breaking changes soon (#1029) so lets just wait for them to hit the download page.

PR #1036, opened 2026-06-03
Compilation: Add even more ARM defines

brodo:
This defines 13 new ARM macros and also contains more fallback cases in the ARM profile and version detection code.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -699,26 +699,9 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                 try w.print("#define __ARM_ARCH_ISA_THUMB {s}\n", .{num});
             }
 
-            if (target.cpu.has(.arm, .vldn_align)) {
-                try define(w, "__ARM_FEATURE_UNALIGNED");
-            }
-
             // ARM ISA means we are not M profile
             if (!target.cpu.has(.arm, .mclass)) {
                 try define(w, "__ARM_ARCH_ISA_ARM");
-            }
-
-            const profile = if (target.cpu.has(.arm, .aclass))
-                "A"
-            else if (target.cpu.has(.arm, .rclass))
-                "R"
-            else if (target.cpu.has(.arm, .mclass))
-                "M"
-            else
-                "";
-
-            if (profile.len > 0) {
-                try w.print("#define __ARM_ARCH_PROFILE '{s}'\n", .{profile});
             }
 
             var arm_version: u8 = 6;
@@ -752,6 +735,7 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                 .{ .v7r, "7R" },
                 .{ .v7m, "7M" },
                 .{ .v7em, "7EM" },
+                .{ .has_v7, "7A" }, // bare armv7 with no profile: default to A
 
                 .{ .v6t2, "6T2" },
                 .{ .v6kz, "6KZ" },
@@ -780,6 +764,19 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                     try w.print("#define __ARM_ARCH {d}\n", .{arm_version});
                     break;
                 }
+            }
+
+            const arm_profile: ?u8 =
+                if (target.cpu.has(.arm, .aclass))
+                    'A'
+                else if (target.cpu.has(.arm, .rclass))
+                    'R'
+                else if (target.cpu.has(.arm, .mclass))
+                    'M'
+                else
+                    null;
+            if (arm_profile) |p| {
+                try w.print("#define __ARM_ARCH_PROFILE '{c}'\n", .{p});
             }
 
             if (arm_version == 5 or (arm_version == 6 and !target.cpu.has(.arm, .mclass)) or arm_version > 6) {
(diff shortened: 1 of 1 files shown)


src/aro/Compilation.zig
```diff
@@ -857,11 +847,24 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
+
+            const min_enum_size: u8 = if (comp.langopts.short_enums) 1 else 4;
+            try w.print("#define __ARM_SIZEOF_MINIMAL_ENUM {d}\n", .{min_enum_size});
+            try w.print("#define __ARM_SIZEOF_WCHAR_T {d}\n", .{comp.type_store.wchar.sizeof(comp)});
         },
         .aarch64, .aarch64_be => |arch| {
             try define(w, "__aarch64__");
             try define(w, "__ARM_64BIT_STATE");
             try define(w, "__ARM_ARCH_ISA_A64");
+            try define(w, "__ARM_FEATURE_CLZ");
```
Vexu, line 866: These are now duplicated with the ones I added below.

src/aro/Compilation.zig
```diff
@@ -896,16 +899,20 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                 .{ .v8_1a, "8_1A" },
                 .{ .v8a, "8A" },
                 .{ .v8r, "8R" },
+                .{ .neon, "8A" }, // generic aarch64 baseline: always at least ARMv8-A
             }) |fs| {
                 if (arm_features.isEnabled(@intFromEnum(fs[0]))) {
-                    try w.print("#define __ARM_ARCH_{s}__ 1\n", .{fs[1]});
-                    const arm_version = fs[1][0];
-                    try w.print("#define __ARM_ARCH {c}\n", .{arm_version});
+                    const arm_version = fs[1][0] - '0';
```
Vexu: If you're not going to define `__ARM_ARCH_XXX__` then this could be simplified to:
```zig
            const arm_version: u8 = if (target.cpu.has(.aarch64, .v9a)) 9 else 8;
            const arm_profile: u8 = if (target.cpu.has(.aarch64, .v8r)) 'R' else 'A';
```

src/aro/Compilation.zig
```diff
@@ -778,6 +766,8 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                     try w.print("#define __ARM_ARCH_{s}__ 1\n", .{fs[1]});
                     arm_version = fs[1][0] - '0';
                     try w.print("#define __ARM_ARCH {d}\n", .{arm_version});
+                    const profile = fs[1][fs[1].len - 1];
+                    try w.print("#define __ARM_ARCH_PROFILE '{c}'\n", .{profile});
```
Vexu: With this the profile could also be `N`, `E`, `2`, `Z`, `K`, `J`, `6`, `T`, `4`, `3` or `2`.

Vexu (changes requested):

src/aro/Compilation.zig
brodo: I removed the duplicates

src/aro/Compilation.zig
brodo: Oh yes. I've switched to your version. It's also way shorter this way.

src/aro/Compilation.zig
brodo: I moved the profile detection out of that block and used `target.cpu.has(.arm, .$class)`.

brodo:
@Vexu review plz!

src/aro/Compilation.zig
```diff
@@ -785,6 +773,19 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
 
+            const arm_profile: u8 =
+                if (target.cpu.has(.arm, .aclass))
+                    'A'
+                else if (target.cpu.has(.arm, .rclass))
+                    'R'
+                else if (target.cpu.has(.arm, .mclass))
+                    'M'
+                else
+                    '-';
```
Vexu: `null` is better suited for this.

src/aro/Compilation.zig
```diff
@@ -860,54 +861,40 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
             try define(w, "__ARM_FEATURE_FMA");
             try w.writeAll("#define __ARM_FEATURE_LDREX 0xF\n");
-            try define(w, "__ARM_FEATURE_IDIV");
-            try define(w, "__ARM_FEATURE_DIV");
+            try define(w, "__ARM_FEATURE_IDIV"); // As specified in ACLE
+            try define(w, "__ARM_FEATURE_DIV"); // For backwards compatibility
+            try define(w, "__ARM_FEATURE_NUMERIC_MAXMIN");
+            try w.writeAll("#define __ARM_FEATURE_DIRECTED_ROUNDING 1\n");
+            try w.writeAll("#define __ARM_STATE_ZA 1\n");
+            try w.writeAll("#define __ARM_STATE_ZT0 1\n");
```
Vexu: Mostly a stylistic nit but using `define` for macros with the value is `1` makes other ones stand out.

brodo:
Everything is done. I've also deleted another duplicate: `__ARM_FEATURE_UNALIGNED`. There was a check that enabled it when `target.cpu.has(.arm, .vldn_align)` which is not what clang checks. [`__ARM_FEATURE_UNALIGNED`](https://developer.arm.com/documentation/101754/0624/armclang-Reference/armclang-Command-line-Options/-munaligned-access---mno-unaligned-access) means "this CPU allows unaligned access". `vldn-align` means ["VLDn instructions that access unaligned data take one more cycle"](https://github.com/llvm/llvm-project/blob/7bc9882cce313b73b263b4c1377ccd322794f413/llvm/lib/Target/ARM/ARMFeatures.td#L336).

PR #1038, opened 2026-06-05
Preprocessor: preprocessor expressions must end with eof

### LINKED ISSUE #687: Preprocessor allows extraneous tokens in preprocessor expression
This doesn't issue an error even though it should:

```c
#if 1 < ABC 2
#endif
```

Identifiers, preprocessor numbers, and unary operators all seem to be incorrectly allowed currently. They're allowed regardless of position (before, after, or even in the middle of an expression).

ehaas:
Fixes #687

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -1154,6 +1154,7 @@ fn restoreTokenState(pp: *Preprocessor, state: TokenState) void {
 fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) MacroError!bool {
     const gpa = pp.comp.gpa;
     const token_state = pp.getTokenState();
+    const error_count = pp.diagnostics.errors;
     defer {
         for (pp.top_expansion_buf.items) |tok| TokenWithExpansionLocs.free(tok.expansion_locs, gpa);
         pp.restoreTokenState(token_state);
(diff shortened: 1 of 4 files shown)


test/cases/msp430 builtin types.c
```diff
@@ -1,7 +1,7 @@
 #include <stddef.h>
 #include <stdarg.h>
 
-#if !defined(MSP430) or !defined(__MSP430__)
+#if !defined(MSP430) || !defined(__MSP430__)
```
ehaas, line 4: we were silently accepting zig syntax!

src/aro/Preprocessor.zig
```diff
@@ -1278,7 +1279,7 @@ fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) MacroError!bool {
         .string_ids = undefined,
     };
     defer parser.strings.deinit(gpa);
-    return parser.macroExpr();
+    return parser.macroExpr(pp.diagnostics.errors == error_count);
```
ehaas, line 1282: Should we loop over added diagnostics and check if there are errors? warnings should probably not suppress this check.

Vexu: Warnings are counted separately.

PR #1039, opened 2026-06-07
Parser: ARM exclusive call type checking

vancluever:
This adds custom type checking for the following ARM exclusive builtins:

```
__builtin_arm_ldrex
__builtin_arm_ldrexd
__builtin_arm_ldaex
__builtin_arm_strex
__builtin_arm_strexd
__builtin_arm_stlex
```

The custom type checking follows what is in LLVM/clang, although there are some slight divergences on part of the fact that arg count and return types are handled separately from arg checking.

The following breakdown of these exclusive calls are as follows:

* Load calls take 1 argument (pointer source), store calls take two   (scalar value source, pointer destination).
* Pointer arguments must be a int/float/pointer type.
* Pointer arguments are coerced to `const volatile *` (load) or   `volatile *` (store). Warnings are surfaced appropriately through the standard coercion code path.
* Value arguments are coerced to the child type of the pointer argument.
* A check is performed to ensure that the pointer argument child is supported - currently, this is fetched from the compilation through a new optional field - the avoids having to repeat the complex version checking that has already been done by `Compilation`.

Tests are included for the old LLVM/clang 21 section in `arm_acle.h`. This is required by the Arm NEON headers, which is keeping us from using it on appropriate systems, e.g., Zig 0.16.0.

Also currently included is the main test ported from LLVM/clang for armv7a. This adequately allows us to test most points of the code in one go. LLVM has other headers for other more restrictive architectures that we can pull in if needed.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -705,65 +705,10 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
             }
 
             var arm_version: u8 = 6;
-            const arm_features = target.cpu.features;
-            for ([_]struct { std.Target.arm.Feature, []const u8 }{
-                .{ .v9_6a, "9_6A" },
-                .{ .v9_5a, "9_5A" },
-                .{ .v9_4a, "9_4A" },
-                .{ .v9_3a, "9_3A" },
-                .{ .v9_2a, "9_2A" },
-                .{ .v9_1a, "9_1A" },
-                .{ .v9a, "9A" },
-
-                .{ .v8_9a, "8_9A" },
-                .{ .v8_8a, "8_8A" },
-                .{ .v8_7a, "8_7A" },
-                .{ .v8_6a, "8_6A" },
-                .{ .v8_5a, "8_5A" },
-                .{ .v8_4a, "8_4A" },
-                .{ .v8_3a, "8_3A" },
-                .{ .v8_2a, "8_2A" },
-                .{ .v8_1a, "8_1A" },
-                .{ .v8_1m_main, "8_1M_MAIN" },
-                .{ .v8a, "8A" },
-                .{ .v8r, "8R" },
-                .{ .v8m_main, "8M_MAIN" },
-                .{ .v8m, "8M_BASE" },
-
-                .{ .v7ve, "7VE" },
-                .{ .v7a, "7A" },
-                .{ .v7r, "7R" },
-                .{ .v7m, "7M" },
-                .{ .v7em, "7EM" },
-                .{ .has_v7, "7A" }, // bare armv7 with no profile: default to A
-
-                .{ .v6t2, "6T2" },
-                .{ .v6kz, "6KZ" },
-                .{ .v6k, "6K" },
-                .{ .v6j, "6J" },
-                .{ .v6sm, "6SM" },
-                .{ .v6m, "6M" },
-                .{ .v6, "6" },
-
-                .{ .v5tej, "5TEJ" },
-                .{ .v5te, "5TE" },
-                .{ .v5t, "5T" },
-
-                .{ .v4t, "4T" },
-                .{ .v4, "4" },
-
-                .{ .v3m, "3M" },
-                .{ .v3, "3" },
-
-                .{ .v2a, "2A" },
-                .{ .v2, "2" },
-            }) |fs| {
-                if (arm_features.isEnabled(@intFromEnum(fs[0]))) {
-                    try w.print("#define __ARM_ARCH_{s}__ 1\n", .{fs[1]});
-                    arm_version = fs[1][0] - '0';
-                    try w.print("#define __ARM_ARCH {d}\n", .{arm_version});
-                    break;
-                }
+            if (target.armVersion()) |v| {
+                arm_version = v.version;
+                try w.print("#define __ARM_ARCH_{s}__ 1\n", .{v.string});
+                try w.print("#define __ARM_ARCH {d}\n", .{v.version});
             }
 
             const arm_profile: ?u8 =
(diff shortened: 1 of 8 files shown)


test/cases/builtins-arm-exclusive.c
```diff
@@ -0,0 +1,232 @@
...
+  res |= __builtin_arm_strex(4, addr);
+  res |= __builtin_arm_strex(42, (short *)addr);
+  res |= __builtin_arm_strex(42, (int *)addr);
+  res |= __builtin_arm_strex(42, (long long *)addr);
+  res |= __builtin_arm_strex(2.71828f, (float *)addr);
+  res |= __builtin_arm_strex(3.14159, (double *)addr);
+  res |= __builtin_arm_strex(&var, (struct Simple **)addr);
+
+  res |= __builtin_arm_strex(42, (volatile char *)addr);
+  // res |= __builtin_arm_strex(42, (char *const)addr);
```
vancluever: PS: This and other similar lines are currently commented out because it looks like the const-ness is currently being stripped here: https://github.com/Vexu/arocc/blob/017c4d2bc28e1d44feb89ef574e6c54586a33475/src/aro/Parser.zig#L7456-L7471

vancluever:
Pretty sure this gets us past our ARM translation issues. I'm wrestling some iOS build issues in that dependency, but I think that might be a toolchain issue on my side, because translating the dependency (wuffs) in isolation with Zig 0.16.0 with this change works.

src/aro/Compilation.zig
```diff
@@ -838,6 +840,7 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
                 else => 0,
             };
             if (ldrex != 0) {
+                comp.arm_ldrex = ldrex;
```
Vexu: This isn't the correct place to set this value since these builtins should be properly checked even with `-undef`. There currently isn't a correct place to do so but you could add a `setTargetOptions` to `LangOpts.zig` and call it from `Driver.parseArgs` after it sets `d.comp.target`.

You could also a packed struct for the type.

src/aro/Parser.zig
```diff
@@ -7589,13 +7657,14 @@ pub const Result = struct {
...
                     {
-                        try p.err(tok, switch (c) {
-                            .assign => .ptr_assign_discards_quals,
-                            .init => .ptr_init_discards_quals,
-                            .ret => .ptr_ret_discards_quals,
-                            .arg => .ptr_arg_discards_quals,
+                        const data: struct { diag: Diagnostic, qts: [2]QualType } = switch (c) {
+                            .assign => .{ .diag = .ptr_assign_discards_quals, .qts = .{ dest_qt, src_original_qt } },
+                            .init => .{ .diag = .ptr_init_discards_quals, .qts = .{ dest_qt, src_original_qt } },
+                            .ret => .{ .diag = .ptr_ret_discards_quals, .qts = .{ dest_qt, src_original_qt } },
```
Vexu: This should also be flipped.

src/aro/Parser.zig
```diff
@@ -9608,6 +9677,151 @@ fn checkComplexArg(p: *Parser, builtin_tok: TokenIndex, first_after: TokenIndex,
...
+            .child = ptr_arg_child_coerced,
+        } });
+        try arg.coerce(p, qt, param_tok, .{ .arg = null });
+    }
+
+    // Child should be a int, float, or pointer
+    const ptr_arg_child_scalar = ptr_arg_child.scalarKind(p.comp);
+    if (!ptr_arg_child_scalar.isInt() and
+        !ptr_arg_child_scalar.isFloat() and
+        !ptr_arg_child_scalar.isPointer())
```
Vexu: Missing a `isReal()` check.

test/cases/builtins-arm-exclusive.c
Vexu: That is correct behavior, you cannot cast qualifiers into an rvalue.

src/aro/Parser.zig
vancluever: The reason I didn't do this wholesale was this in the `typeof.c` test (amongst other tests):

```
======= expected to find error =======
typeof.c:97:29: warning: initializing 'const typeof(int *)' (aka 'int *const') from incompatible type 'const int [2]' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]

=== but output does not contain it ===
...
typeof.c:97:29: warning: initializing 'const int [2]' from incompatible type 'const typeof(int *)' (aka 'int *const') discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
    typeof(int *)const p4 = arr; // type listed in error message is incorrect
```

This seems correct?

clang gives the same warning (line number different as I just copied the affected function):

```
> clang -fsyntax-only typeof.c
...
typeof.c:6:24: warning: initializing 'const typeof(int *)' (aka 'int *const') with an expression of type
      'const int[2]' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
    6 |     typeof(int *)const p4 = arr; // type listed in error message is incorrect
      |                        ^    ~~~
1 warning generated.
```

src/aro/Compilation.zig
```diff
@@ -876,35 +806,10 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
-            }) |fs| {
-                if (arm_features.isEnabled(@intFromEnum(fs[0]))) {
-                    try w.print("#define __ARM_ARCH_{s}__ 1\n", .{fs[1]});
-                    const arm_version = fs[1][0];
-                    try w.print("#define __ARM_ARCH {c}\n", .{arm_version});
-                    const profile = fs[1][fs[1].len - 1];
-                    try w.print("#define __ARM_ARCH_PROFILE '{c}'\n", .{profile});
-                    break;
-                }
+            if (target.armVersion()) |v| {
```
Vexu: You can't check for arm features on aarch64. #1036 also simplifies this part.

src/aro/Target.zig
```diff
@@ -1789,6 +1789,70 @@ pub fn parseOs(result: *std.Target.Query, text: []const u8, version_string: ?*[]
...
+        .{ .v3m, "3M" },
+        .{ .v3, "3" },
+
+        .{ .v2a, "2A" },
+        .{ .v2, "2" },
+    }) |fs| {
+        if (target.cpu.features.isEnabled(@intFromEnum(fs[0]))) {
+            return .{
+                .version = fs[1][0] - '0',
+                .profile = fs[1][fs[1].len - 1],
```
Vexu: `J` is not a valid profile.

src/aro/Parser.zig
Vexu: The `.assign` and `.init` cases are correct, I was commenting on the `.ret` case specifically.

src/aro/Target.zig
```diff
@@ -1789,6 +1789,70 @@ pub fn parseOs(result: *std.Target.Query, text: []const u8, version_string: ?*[]
     };
 }
 
+pub fn armVersion(target: Target) ?struct { version: u8, profile: u8, string: []const u8 } {
```
Vexu: This function should `assert(target.cpu.arch.isArm())`.

src/aro/Target.zig
vancluever: Just removing the .profile field since it's not used anymore (was only used in the aarch64 bits and since it's being "statically" set now, I think we can drop it, for the time being at least).

test/cases/builtins-arm-exclusive.c
vancluever: I've un-commented these and and just added the warnings/errors to the expected list.

Vexu (approved):

vancluever:
Thanks @Vexu!

PR #1047, opened 2026-06-13
Parser: array element size must be a multiple of alignment

### LINKED ISSUE #1046: Array element size not required to be multiple of alignment
Noticed this while working on #1045 (it's present in master and that PR):
```c
void array_element_typedef(void) {
    typedef int elem __attribute__((aligned(8)));
    typedef elem arr[4];
}
```

aro accepts this; clang reports:
```
test.c:3:21: error: size of array element of type 'elem' (aka 'int') (4 bytes) isn't a multiple of its alignment (8 bytes)
    3 |     typedef elem arr[4];
      |                     ^
```

(this was on an aarch64-mac)

ehaas:
Fixes #1046

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -3661,6 +3661,13 @@ const Declarator = struct {
                     try p.err(source_tok, .array_func_elem, .{});
                     return .nested_invalid;
                 }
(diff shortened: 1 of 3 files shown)


ehaas:
Incorporated into #1083

PR #1048, opened 2026-06-14
Rewrite attributes

### LINKED ISSUE #367: MSVC isn't parsing `__declspec`?

### LINKED ISSUE #944: Incorrect attribute application for nested pointers
This should be a single pointer to a single pointer, but the attribute gets applied twice to the outer pointer:
```c
#include <ptrcheck.h>

void foo(int *__single *__single p) {
	p[0][5] = 5;
}
```

```
% arocc -fexperimental-bounds-safety test.c
test.c:3:25: warning: pointer annotated with single multiple times. Annotate only once to remove this warning [-Wbounds-attributes-redundant]
void foo(int *__single *__single p) {
                        ^
/Users/ehaas/source/arocc/zig-out/include/ptrcheck.h:11:33: note: expanded from here
#define __single __attribute__((__single__))
                                ^
```

```
% clang -Xclang -fexperimental-bounds-safety test.c
test.c:4:6: error: array subscript on single pointer 'p[0]' must use a constant index of 0 to be in bounds
    4 |         p[0][5] = 5;
      |         ~~~~^~
1 error generated.
```

Vexu:
Will close #944
Will close #367


Vexu:
For anyone interested in taking a look this is now mostly ready with 21 integration tests failing with `TODO: implement 'X' attribute` and 204 failing record layout tests (likely caused by maybe 3 bugs).

ehaas:
I'm away from my main computer until the weekend but thank you for taking this on, the old attribute implementation had a bit too much comptime :)

test/cases/_BitInt change size.c
```diff
@@ -1,9 +1,7 @@
 _BitInt(10) x = 1.2;
-_Complex unsigned _BitInt(10) y = 1.2;
 
 /** manifest:
 syntax
 
 _BitInt change size.c:1:17: warning: implicit conversion from 'double' to 'signed _BitInt(10)' changes value from 1.2 to 1 [-Wfloat-conversion]
-_BitInt change size.c:2:35: warning: implicit conversion from 'double' to '_Complex unsigned _BitInt(10)' changes value from 1.2 to 1 [-Wfloat-conversion]
```
dotcarmen, line 8: did you intend to remove this test? seems helpful and like it shouldn't be removed

src/aro/Attribute.zig
```diff
@@ -709,830 +274,343 @@ const attributes = struct {
...
+    riscv: Riscv,
+    keyword: Keyword,
+
+    pub const Standard = enum {
+        deprecated,
+        fallthrough,
+        maybe_unused,
+        nodiscard,
+        noreturn,
+        _Noreturn, // Deprecated
```
dotcarmen: mind moving this comment and others below to be doc-comments?
```suggestion
        /// Deprecated in c23
        _Noreturn,
```

src/aro/Attribute.zig
```diff
@@ -709,830 +274,343 @@ const attributes = struct {
...
-                .float, .fp16, .bf16 => true,
-                .double => p.comp.target.cpu.arch.isAARCH64(),
-                else => false,
+            .gnu => {
+                assert(opt_vendor == null);
+                inline for (.{ "gnu", "clang", "aro" }) |vendor_name| {
+                    if (std.meta.stringToEnum(@FieldType(Namespaced, vendor_name), name)) |tag| {
+                        return @unionInit(Namespaced, vendor_name, tag);
+                    }
+                }
```
dotcarmen, line 571: IMO this should match the compiler being emulated


```suggestion
                switch (comp.langopts.emulate) {
                    .msvc => {}, // TODO
                    inline .clang, .gcc => |vendor| if (std.meta.stringToEnum(@FieldType(Namespaced, vendor_name), name)) |tag| {
                        return @unionInit(Namespaced, vendor_name, tag);
                    },
                    .no => if (std.meta.stringToEnum(Namespaced.Aro, name)) |tag| {
                        return @unionInit(Namespaced, "aro", tag);
                    },
                }
```

src/aro/Attribute/Map.zig
```diff
@@ -0,0 +1,272 @@
+const std = @import("std");
+const mem = std.mem;
+const Attribute = @import("../Attribute.zig");
+const Compilation = @import("../Compilation.zig");
+const Tree = @import("../Tree.zig");
+const TokenIndex = Tree.TokenIndex;
+
+const Map = @This();
```
dotcarmen: mind moving this next to the field defs? easier to see the type when using zls go to definition (when ZLS works again...)

src/aro/Attribute/Map.zig
```diff
@@ -0,0 +1,272 @@
...
+}
+
+pub fn put(map: *Map, gpa: mem.Allocator, attribute: Attribute) !Ref {
+    var repr: Repr = .{
+        .tag = undefined,
+        .data = attribute.tok,
+        .name = attribute.name,
+        .syntax = attribute.syntax,
+    };
+    _ = &repr;
```
dotcarmen: no longer necessary

src/aro/Attribute/Map.zig
```diff
@@ -0,0 +1,272 @@
...
+    var max_requested: ?u32 = null;
+    for (map.attrs(node)) |ref| {
+        const attr = map.get(ref);
+        if (attr.args != .alignment) continue;
+
+        const requested = if (attr.args.alignment) |alignment| alignment else comp.target.defaultAlignment();
+        if (max_requested == null or max_requested.? < requested) {
+            max_requested = requested;
+        }
+    }
```
dotcarmen, line 305: it seems most attributes will be using the last attribute present (by using `getAttribute`). is this implementation the same behavior of clang/gcc/msvc, or can this be simplified/removed to follow the conforming behavior?

dotcarmen (commented):
still haven't reviewed Wip, Parser, and TypeStore. lgtm so far, a lot better than the old way :) i'll try rebasing my blocks PR when i have time in the next few days

test/cases/_BitInt change size.c
Vexu: I made `_Complex _BitInt` an error as it is in gcc and clang.

src/aro/Attribute/Map.zig
Vexu: The biggest specified alignment is used so this logic needs to exist either when applying attributes or when fetching them.

src/aro/Attribute.zig
Vexu: If we were to go that route then we should also check which specific attributes each compiler supports but seeing as attributes are an extension and unknown ones are ignored I don't really want to do that.

ehaas:
Previously we had a warning for noreturn functions that do return - 

```c
__attribute__((noreturn))  void f(void) { return; }
```
```
test.c:1:49: warning: function 'f' declared 'noreturn' should not return [-Winvalid-noreturn]
__attribute__((noreturn))  void f(void) { return; }
```

Looks like that isn't triggering any more.

PR #1050, opened 2026-06-15
Diagnostics: treat extra tokens at end of macro directive as warning

### LINKED ISSUE #1049: Extra tokens at end of macro directive being treated as error
```c
#infdef FOO(x)
#define FOO(x) x
#endif
```
Output: 
```
foo/test.def.in:1:12: error: extra tokens at end of macro directive
#ifndef FOO(x)
           ^
1 error generated.
```

Aro treats this only as an error. Clang and GCC emit a warning about the trailing `(x)` after FOO in the first line. While GCC doesn't have an explicit flag for this warning, [Clang does](https://clang.llvm.org/docs/DiagnosticsReference.html#wextra-tokens). 

Clang output:
```
foo/test.def.in:1:12: warning: extra tokens at end of #ifndef directive [-Wextra-tokens]
    1 | #ifndef FOO(x)
      |            ^
      |            //
```

arthur-teixeira:
Fixes #1049

### DIFF
--- a/src/aro/Diagnostics.zig
+++ b/src/aro/Diagnostics.zig
@@ -204,6 +204,7 @@ pub const Option = enum {
     @"base-file-extension",
     @"include-level-extension",
     @"blocks-extension",
+    @"extra-tokens",
 
     /// GNU extensions
     pub const gnu = [_]Option{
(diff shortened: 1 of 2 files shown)


Vexu:
Thanks!

Issue #1044, opened 2026-06-10
Preprocessor quadratic-time expansion on long linear object-like macros


ym-xie:
Hi, I try the following code

```sh
python3 - <<'PY' > macro-chain.c
N = 45000
for i in range(N):
    print(f"#define A{i} A{i + 1}")
print(f"#define A{N} int x;")
print("A0")
PY

zig translate-c macro-chain.c > /dev/null
```

it shows that preprocessor can spend quadratic time expanding a linear chain of
object-like macros.

| Macro chain length | Result |
| ---: | ---: |
| 30000 | 16.204s |
| 45000 | 35.024s |
| 60000 | exceeded 60s and was killed |

```sh
zig version
0.17.0-dev.830+1f43e049d
```

I have opened a issue in zig: https://codeberg.org/ziglang/translate-c/issues/412

PR #1051, opened 2026-06-17
Hideset: store precomputed identifier slices

arthur-teixeira:
improves #1044 

Before:
- 7k long macro chain (I couldn't profile a larger input, Tracy OOMed and my pc crashed):
image

- 25k long macro chain:

```
$ time arocc -E macro-chain.c -nostdinc >/dev/null
real    0m7.643s
user    0m7.615s
sys     0m0.005s
```

After:
- 45k long chain:
image

```
$ time ./zig-out/bin/arocc -nostdinc -E macro-chain.c > /dev/null
real    0m0.398s
user    0m0.391s
sys     0m0.006s
```

This doesn't fix the quadratic-time expansion, but it removes the largest overhead in the process, which is re-tokenizing all expansion locations stored in the hideset. I didn't try that hard to fix the quadratic-time issue because I am not very familiar with how hidesets work, but I would be glad to try to improve it more.

### DIFF
--- a/src/aro/Hideset.zig
+++ b/src/aro/Hideset.zig
@@ -13,25 +13,12 @@ const Allocator = mem.Allocator;
 
 const Compilation = @import("Compilation.zig");
 const Source = @import("Source.zig");
-const Tokenizer = @import("Tokenizer.zig");
 
 pub const Hideset = @This();
 
 const Identifier = struct {
     id: Source.Id = .unused,
     byte_offset: u32 = 0,
-
-    fn slice(self: Identifier, comp: *const Compilation) []const u8 {
-        var tmp_tokenizer: Tokenizer = .{
-            .buf = comp.getSource(self.id).buf,
-            .langopts = comp.langopts,
-            .index = self.byte_offset,
-            .source = .generated,
-            .splice_locs = &.{},
-        };
-        const res = tmp_tokenizer.next();
-        return tmp_tokenizer.buf[res.start..res.end];
-    }
 
     fn fromLocation(loc: Source.Location) Identifier {
         return .{
@@ -43,6 +30,7 @@ const Identifier = struct {
 
 const Item = struct {
     identifier: Identifier = .{},
+    str: []const u8,
     next: Index = .none,
 
     const List = std.MultiArrayList(Item);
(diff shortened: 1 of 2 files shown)


src/aro/Hideset.zig
```diff
@@ -58,24 +46,32 @@ map: std.AutoHashMapUnmanaged(Identifier, Index) = .{},
 /// until hideset is deinit'ed
 tmp_map: std.AutoHashMapUnmanaged(Identifier, void) = .{},
 linked_list: Item.List = .{},
-comp: *const Compilation,
+gpa: Allocator,
+
+const Temp = struct { identifier: Identifier, str: []const u8 };
```
ehaas: This seems like it's left over from other work you were doing? It's not used.

src/aro/Hideset.zig
```diff
@@ -13,26 +13,13 @@ const Allocator = mem.Allocator;
 
 const Compilation = @import("Compilation.zig");
```
ehaas, line 14: `Compilation` is also no longer used in this file so the import can be removed

src/aro/Hideset.zig
arthur-teixeira: Right, I removed it but forgot to commit it

ehaas:
My only concern here (and this is going off of memory from originally implementing the hidesets) the potential issue is the lifetime of slices into the generated buffer. I'm not sure if this approach will potentially end up storing slices that get invalidated when the generated buffer grows (if so we would need source id + start index + length struct)

Vexu:
I was able to find a case where this causes one incorrect expansion to change into another more incorrect expansion:

```c
#define bb aa
#define aa short(b,b)
#define short(a, b) a ## b
#define foo(x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x) short(x, x)
foo(a)
```

The source + index + length struct would likely match the current behavior

arthur-teixeira:
Great point, I didn't think about generated tokens

Vexu:
Thanks!

PR #1054, opened 2026-06-22
Update to Zig 0.17.0-dev.947+36069a2a7

Vexu:
This caused a bunch of record layout tests to fail due to `cTypeBitSize` changes, I assume the new values are correct.


alexrp:
> This caused a bunch of record layout tests to fail due to `cTypeBitSize` changes, I assume the new values are correct.

What are the new values and for what targets?

Vexu:
I was wrong, I'll try to bisect for the actual reason.

ehaas:
@Vexu @alexrp 

All of the failures are with target `mips-linux-gnuabi64` and mcpu `mips64r6` or `mips32`, GCC emulation.

In our tests we expect `sizeof(long) == 8` but now Zig is saying `sizeof(long) == 4` for that target. I haven't dug in more yet to see which is correct; I'm not super familiar with the mips platform.

alexrp:
`mips-linux-gnuabi64` looks like a nonsensical target to me; `gnuabi64`/`muslabi64` are only valid for `mips64`/`mips64el`.

ehaas:
Ok, seems like the right fix might be to update the text fixtures then.

PR #1058, opened 2026-06-23
Pragmas: store null in the stack.

Fahien:
Hi there :)

Discovered while debugging this: https://github.com/zig-gamedev/zphysics/pull/27


Vexu:
Thanks!

PR #1061, opened 2026-06-24
Preprocessor: skip over comments in expressions

### LINKED ISSUE #1060: Regression: preserving comments causes preprocessor errors
test.c:
```c
#if !defined(_SAL_VERSION_SAL2) // [
foo
#endif
```

Without `--comments` (`arocc -E test.c`) everything works as expected:

```c
# 1 "test.c" 1
# 1 "<builtin>" 1
# 380 "<builtin>"
# 1 "<command line>" 1
# 1 "test.c" 2

foo
```

With `--comments` (`arocc -E --comments test.c`):

```
test.c:1:33: error: token is not a valid binary operator in a preprocessor subexpression
#if !defined(_SAL_VERSION_SAL2) // [
                                ^
```

This is a regression, not sure exactly when it happened (EDIT: likely caused by https://github.com/Vexu/arocc/pull/1038). Also reproduces with `-C` and `-CC`

squeek502:
Fixes #1060, partially addresses #1065

---

Without this fix, the modified tests would fail with

```
========= new error ==========
preserve comments.c:17:7: error: token is not a valid binary operator in a preprocessor subexpression

=== not in expected errors ===
in case 'preserve comments'

========= new error ==========
preserve comments in macros.c:17:7: error: token is not a valid binary operator in a preprocessor subexpression

=== not in expected errors ===
in case 'preserve comments in macros'
```

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -1167,6 +1167,10 @@ fn expr(pp: *Preprocessor, tokenizer: *Tokenizer) MacroError!bool {
             .nl, .eof => break tok,
             .whitespace => if (pp.top_expansion_buf.items.len == 0) continue,
             .comment => continue,
+            .unterminated_comment => {
+                try pp.err(tok, .unterminated_comment, .{});
+                continue;
+            },
             else => {},
         }
         try pp.top_expansion_buf.append(gpa, tokFromRaw(tok));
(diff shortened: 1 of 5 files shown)


Vexu:
Comment tokens should not be passed to the parser from `Preprocessor.expr` at all:

```c
#if 0 /*bar*/ == 2
#endif
```

squeek502:
Pushed a fix that's in line with what you said, and also expanded the test case a bit more.

PR #1066, opened 2026-06-25
Attribute: avoid dangling function params in applyCallingConvention

happystraw:
`applyCallingConvention` updates a function type by changing func.cc and writing the full type back through `TypeStore.set`.

`func.params` may be a borrowed slice into type_store.extra.

https://github.com/Vexu/arocc/blob/8f4df188ac8256fc4cd9bc069a3b8dd68fd62b68/src/aro/Attribute/Wip.zig#L1117-L1119

During `TypeStore.set`, type_store.extra may reallocate, causing `func.params` to become a dangling reference to invalid memory.

https://github.com/Vexu/arocc/blob/8f4df188ac8256fc4cd9bc069a3b8dd68fd62b68/src/aro/TypeStore.zig#L2147

https://github.com/Vexu/arocc/blob/8f4df188ac8256fc4cd9bc069a3b8dd68fd62b68/src/aro/TypeStore.zig#L2153

### DIFF
--- a/src/aro/Attribute/Wip.zig
+++ b/src/aro/Attribute/Wip.zig
@@ -1214,12 +1214,11 @@ fn applyCallingConvention(wip: *Wip) !void {
 
     // We cannot pass the function type directly here because the pointer to
     // type_store.extra might get invalidated while setting the updated type.
-    var owned_params: ?[]Type.Func.Param = null;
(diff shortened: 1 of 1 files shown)


Vexu:
Thanks!

PR #1068, opened 2026-07-01
TypeStore: round record size up to alignment for MSVC

ehaas:
This regresses 3 record tests but brings the overall number of failing ones down to 61. I included the msvc-enum branch just to avoid conflicts, https://github.com/Vexu/arocc/pull/1067 should be merged first.

### DIFF
--- a/test/cases/alignment on typedef msvc.c
+++ b/test/cases/alignment on typedef msvc.c
@@ -36,6 +36,13 @@ __declspec(align(16)) typedef enum {
 
 _Static_assert(_Alignof(AlignedEnum) == 16, "");
 
+__declspec(align(16)) typedef struct {
+  int x;
+} AlignedStruct;
+
+_Static_assert(_Alignof(AlignedStruct) == 16, "");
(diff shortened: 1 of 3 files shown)


test/cases/alignment on typedef msvc.c
```diff
@@ -36,6 +36,13 @@ __declspec(align(16)) typedef enum {
 
 _Static_assert(_Alignof(AlignedEnum) == 16, "");
 
+typedef struct __declspec(align(16)) {
+  int x;
+} AlignedStruct;
+
+_Static_assert(_Alignof(AlignedStruct) == 16, "");
```
Vexu, line 56: Could you add a test case that only works with this change applied?

ehaas: Nice catch, I was testing a bunch of combos and used the wrong one - I rebased and switched it so the `__declspec` is before the `typedef`

ehaas:
Pushed a few more fixes - prevent underaligned enums in msvc and a fix for record alignment through multiple layers of typedefs

Issue #1052, opened 2026-06-18
Issue parsing MacOS CoreGraphics


vancluever:
```
% cat c.h
#include <CoreGraphics/CoreGraphics.h>
```

```
arocc c.h \
  -nostdlibinc \
  -O0 \
  -idirafter "/nix/store/elided-zig-0.16.0/lib/zig/include" \
  -isystem "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk/usr/include" \
  -isystem "/opt/homebrew/include" \
  -iframework "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk/System/Library/Frameworks"
```

Output below.

I think this is possibly related to what @dotcarmen is working on in #971 but I'm not too sure. Debug sessions would indicate that this is failing on actually printing the diagnostic on sections that are very close to the ones that are actually printing in relation to the lack of ability to null the block types, so I'm not too sure if it's something that would be better covered once this work lands, or if all of this is a red herring. 🙂 

## Output

```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGPDFArray.h:113:40: error: nullability specifier cannot be applied to non-pointer type 'CGPDFArrayApplierBlock' (aka '_Bool (^)(unsigned long, struct CGPDFObject *, void *)')
    CF_NOESCAPE CGPDFArrayApplierBlock cg_nullable block, void * __nullable info)
                                       ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGBase.h:352:24: note: expanded from here
#   define cg_nullable __nullable
                       ^
<builtin>:79:20: note: expanded from here
#define __nullable _Nullable
                   ^
[...]

PR #1072, opened 2026-07-06
TypeStore: add QualType.isCombinedDeclarator

vancluever:
This might be useful to help us catch some panics while the WIP parts of the compiler are being worked on still - particularly I used it to keep a panic from happening that was a result of the reduction case in #1052.


Vexu:
A `.declarator_combine` appearing in either of those places is a bug and should cause a panic.

vancluever:
Yeah no worries, maybe best to close this then for now (unless there's merit to having it somewhere else), it just helped to get these crashes out of the way so I could get at what needed implementing. #1073 works independent of it so it's not needed.

PR #1074, opened 2026-07-07
Compilation: define __BLOCKS__ when blocks are enabled

vancluever:
This gets defined when blocks are enabled in clang: https://github.com/llvm/llvm-project/blob/9e22a5950e51ace7f50729d7f556ca99d06fa8bb/clang/lib/Frontend/InitPreprocessor.cpp#L1010

Found while trying to translate `#include <CoreVideo/CoreVideo.h>` on MacOS.

PS: Not too sure if this should be implicit or not somewhere, but `-fblocks` specifically needs to be enabled when doing C translation for the MacOS frameworks that employ blocks. I don't recall if this is necessary or not in Zig right now currently (it might just be in our flags somewhere that I haven't seen yet).

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -1189,6 +1189,11 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
             }
         },
     }
+
+    // Blocks
+    if (comp.langopts.blocks) {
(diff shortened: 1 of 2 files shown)


Vexu:
This is included in #971

vancluever:
Ahh yeah, I see it - no problems - will close and keep an eye out for that!

PR #1075, opened 2026-07-08
Ensure __has_extension treats features as extensions

vancluever:
Confirmed that this was the behavior in clang for prior art, both as documented, and also as implemented (in [`PPMacroExpansion.cpp`](https://github.com/llvm/llvm-project/blob/da4bbda3dcefbaf88c2f336afd03c4a6ebb41a27/clang/lib/Lex/PPMacroExpansion.cpp#L1145-L1146)).

Additionally, as documented in clang and implemented in the aforementioned, we always return the result from `hasFeature` if `-pedantic-errors` is used.

### DIFF
--- a/test/cases/feature as extension pedantic.c
+++ b/test/cases/feature as extension pedantic.c
@@ -1,0 +1,15 @@
+// has_extension pedantic test (c_alignas is a C11 feature, so this should
+// return unsupported on C99).
+
+#if __has_extension(c_alignas)
+#error expected c_alignas to be unavailable as an extension
+#endif
+
+int main(void) {
+  return 0;
+}
+
+/** manifest:
+syntax
+args = -pedantic-errors -std=c99
+*/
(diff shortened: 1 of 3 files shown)


Vexu (commented):
Could you split the check into two like it is in the clang version? I find that much easier to understand. The pedantic error check could also use a switch for readability.

vancluever:
@Vexu done!

test/cases/feature as extension pedantic.c
```diff
@@ -0,0 +1,15 @@
+// has_extension pedantic test (c_alignas is a C11 feature, so this should
+// return unsupported on C99).
+
+#if __has_extension(c_alignas)
+#error expected c_alignas to be unavailble as an extension
```
ehaas: Since this is a pedantic test I feel justified pointing out the ~type~ typo in `unavailble` :)

vancluever: Will fix shortly 🤣

Vexu:
Thanks!

PR #1080, opened 2026-07-17
Preprocessor: do not double-advance after include macro expansion

vancluever:
Macro filename expansion in `#include` directives already advances to end-of-line, so the trailing token check at the end was mistakenly double-advancing in this case, causing the line after the initial `#include` to be skipped over.

This commit fixes it so that the state is only advanced when we are not doing macro expansion, by just moving the handling of the trailing check into the non-macro/macro handler prongs.

### DIFF
--- a/src/aro/Preprocessor.zig
+++ b/src/aro/Preprocessor.zig
@@ -3700,11 +3700,26 @@ fn findIncludeFilenameToken(
     }
 
     const source_tok = tokFromRaw(first);
-    const filename_tok, const expanded_trailing = switch (source_tok.id) {
-        .string_literal, .macro_string => .{ source_tok, false },
-        else => expanded: {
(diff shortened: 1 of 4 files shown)


vancluever:
I found this attempting to translate FreeType through a simple `#include <freetype/freetype.h>`, and also through HarfBuzz with FreeType integration.

Vexu (approved):
Nice catch! The trailing tokens aren't handled correctly in the `ignore_trailing_tokens` case but that matches master so it is out of scope for this PR.

vancluever:
Thanks @Vexu! Yeah I didn't really look at anything beyond this case; particularly I didn't want to work around the problem so I didn't look at what would happen if I toggled that on.

In any case this seems like (currently) the last thing affecting Ghostty, will send along what I find if that changes though!

PR #1083, opened 2026-07-23
record_runner: Use clang for msvc targets for certain tests

ehaas:
In the original repr-c repo 0055 - 0059 were using clang for MSVC targets e.g. - https://github.com/mahkoh/repr-c/blob/master/repc/tests/testfiles/0055/config.toml

Previously we set the --emulate option implicitly from the target; now we use the value encoded in the test file itself; also update those tests to specify clang instead of msvc. This revealed that several of those tests are not currently passing; a follow up step will be to check if the encoded layout is correct or not.


ehaas:
@Vexu once the zig update is rebased I think this is in good shape - I went through and validated all the records tests against all the msvc targets using godbolt. For some I had to update the tests to use correct values, for some I had to pull different targets apart into separate defines, for some we were actually passing when using the correct values and for some we weren't actually passing.

Vexu:
Including the ones that were already passing? Nice, I'm starting to not trust the repr-c results. We should probably also check the clang and gcc tests as well. Clang at least should be easy to check since it can cross compile.

ehaas:
Only MSVC so far but I can check the others. All the failures (MSVC tests themselves being incorrect) were with the "extra" tests; not sure if they are part of the original repr-c or added in during the original port.

Edit: yes verified all 88 tests on godbolt for all 4 msvc's (x86, x64, arm32, arm64)

ehaas:
I cherry-picked the fix from #1047 and used that to update the record tests that were invalid due to that error. I verified the clang targets locally using clang21 that I built from source - none of them failed due to incorrect alignment specifications, just that array element alignment issue.

ehaas:
8 and 10 are giving a bunch of array element alignment not multiple of size in various gcc's

45 is failing in a bunch of targets

`x86_64-x86_64-windows-gnu` is failing 68 and 82 using MinGW

Vexu (approved):

ehaas:
Ok to the extent possible I went through and validated the GCC layout tests (x86-windows gcc doesn't have a godbolt compiler available). 8, 10, and 45 are erroring with `error: size of array element is not a multiple of its alignment` on GCC, but not on clang. I think it has something to do with GCC looking deeper into nested typedefs but I'm not sure. However the actual layout numbers appear to be correct (I can assert on the element size and align, and the overall array size, and still get the same answer on GCC and arocc).

So, at this point unless anything else comes up I'm willing to conclude my investigation.

PR #1084, opened 2026-07-23
update to zig v0.17.0-dev.1454+5faa79730

dotcarmen:


Vexu (changes requested):
Missing one `Debug` in `build.zig`, needs to update `build.zig.zon`.

dotcarmen:
what about `deps/zig/`? apparently this passes CI as-is, but `deps/zig/arch/x86_64/Mir.zig` has a reference to `.Debug` and `.ReleaseSafe`. is that not currently tested at all? i'm assuming you'll update that dir to keep in sync with zig?

Vexu:
It's not currently used, I should just delete it, it's very out of date anyways.

PR #1088, opened 2026-07-29
Parser: use precedence table for binary expression parsing

Vexu:
Smaller improvement but an improvement still.
```
Benchmark 1 (50 runs): bin/arocc-master sqlite3.c -fsyntax-only -w
  measurement          mean ± σ            min … max           outliers         delta
  wall_time           203ms ± 5.58ms     200ms …  241ms          1 ( 2%)        0%
  peak_rss           80.1MB ±  218KB    79.5MB … 80.4MB          0 ( 0%)        0%
  cpu_cycles         1.05G  ± 4.01M     1.04G  … 1.06G           0 ( 0%)        0%
  instructions       2.91G  ± 14.8      2.91G  … 2.91G           1 ( 2%)        0%
  cache_references   78.3M  ±  226K     77.8M  … 79.0M           1 ( 2%)        0%
  cache_misses       1.51M  ± 17.7K     1.47M  … 1.55M           0 ( 0%)        0%
  branch_misses      8.99M  ± 46.7K     8.90M  … 9.09M           0 ( 0%)        0%
Benchmark 2 (51 runs): bin/arocc-branch sqlite3.c -fsyntax-only -w
  measurement          mean ± σ            min … max           outliers         delta
  wall_time           198ms ±  985us     196ms …  201ms          1 ( 2%)        ⚡-  2.8% ±  0.8%
  peak_rss           80.0MB ±  165KB    79.5MB … 80.3MB          0 ( 0%)          -  0.2% ±  0.1%
  cpu_cycles         1.02G  ± 3.81M     1.01G  … 1.03G           0 ( 0%)        ⚡-  2.5% ±  0.1%
  instructions       2.79G  ± 9.02      2.79G  … 2.79G           0 ( 0%)        ⚡-  4.0% ±  0.0%
  cache_references   65.4M  ±  512K     64.2M  … 66.4M           0 ( 0%)        ⚡- 16.5% ±  0.2%
  cache_misses       1.47M  ± 13.4K     1.45M  … 1.51M           1 ( 2%)        ⚡-  2.4% ±  0.4%
  branch_misses      8.96M  ± 36.5K     8.90M  … 9.04M           0 ( 0%)          -  0.3% ±  0.2%
```

### DIFF
--- a/src/aro/Parser.zig
+++ b/src/aro/Parser.zig
@@ -11,7 +11,6 @@ const char_info = @import("char_info.zig");
 const Compilation = @import("Compilation.zig");
 const Diagnostics = @import("Diagnostics.zig");
 const InitList = @import("InitList.zig");
-pub const Diagnostic = @import("Parser/Diagnostic.zig");
 const Preprocessor = @import("Preprocessor.zig");
 const record_layout = @import("record_layout.zig");
 const Source = @import("Source.zig");
@@ -30,6 +29,8 @@ const Type = TypeStore.Type;
 const QualType = TypeStore.QualType;
 const Value = @import("Value.zig");
 const ArmLdrex = @import("LangOpts.zig").ArmLdrex;
+
+pub const Diagnostic = @import("Parser/Diagnostic.zig");
 
 const NodeList = std.ArrayList(Node.Index);
 const TentativeDefinitionKey = struct {
@@ -145,7 +146,6 @@ maybe_unused_decl: std.AutoArrayHashMapUnmanaged(Node.Index, bool) = .empty,
 local_unused_err_count: u32 = 0,
 
 // configuration and miscellaneous info
-no_eval: bool = false,
 extension_suppressed: bool = false,
 contains_address_of_label: bool = false,
 label_count: u32 = 0,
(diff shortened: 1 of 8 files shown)


src/aro/Parser.zig
```diff
@@ -8248,424 +8048,478 @@ fn constExpr(p: *Parser, decl_folding: ConstDeclFoldingMode) Error!Result {
     defer p.const_decl_folding = const_decl_folding;
     p.const_decl_folding = decl_folding;
 
-    const res = try p.expect(condExpr);
+    const res = try p.expect(assignExpr);
```
ehaas, line 8052: Was this intentional? Seems like you shouldn't be able to use an assignment expression when expecting a constant expression

ehaas: The old parser rejects the following code but the new one accepts it:
```c
_Static_assert((int){1} = 0, "");
```

Vexu: Not intentional but could be an improvement. It matches GCC and shouldn't cause conflicts anywhere.

ehaas: I'm not seeing it accepted by GCC or clang:
GCC https://godbolt.org/z/E54bfWP8h
clang https://godbolt.org/z/ajqnvv4n3

Vexu: The assertion is not supposed to succeed (and now doesn't after I fixed a different bug) but in GCC it's a semantic error while in clang it's a parsing error.

PR #1090, opened 2026-07-30
record layout: use MSVC layout if targeting MSVC abi with clang emulation

ehaas:
The diff looks messier than it actually is - previously we always used SysV layout when emulating clang, gcc, and no emulation; and MSVC layout when emulating MSVC. The change is to now use MSVC layout when emulating clang, _if_ the target is MSVC abi.

There is still some issue with 55 on x86/x86_64. And this regresses `test/cases/msvc anonymous default pack clang.c` so something about aligned bitfields is not quite right with msvc layout.

### DIFF
--- a/src/aro/record_layout.zig
+++ b/src/aro/record_layout.zig
@@ -573,50 +573,53 @@ pub fn compute(
     const packed_attr = am.hasAttribute(opt_decl, .@"packed");
     const requested_align = am.requestedAlignment(opt_decl, comp);
 
-    switch (comp.langopts.emulate) {
-        .no, .gcc, .clang => {
-            var context: SysVContext = .{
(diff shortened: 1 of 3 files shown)


src/aro/record_layout.zig
```diff
@@ -573,50 +573,53 @@ pub fn compute(
...
-                .size_bits = 0,
-                .comp = comp,
-                .am = am,
-                .ongoing_bitfield = null,
-            };
+    const use_sysv_layout = switch (comp.langopts.emulate) {
+        .msvc => false,
+        .clang => comp.target.abi != .msvc,
+        .gcc, .no => true,
+    };
```
Vexu, line 580: I think `no` should match clang here.

Vexu (approved):

PR #1101, opened 2026-08-10
record_layout: in MSVC required alignment overrides pragma pack

ehaas:


Vexu (approved):
Nice! Looks like it won't be long until that `compErr` can be removed.

PR #1102, opened 2026-08-12
Add AVR int24 support

### LINKED ISSUE #1089: AVR layout allows fun integer sizes for packed enums
From 0062_test.c; this should pass with target `--target=avr-other-none`

```c
typedef enum {
    F121 = 8388608,
} __attribute__((packed)) A028;
_Static_assert(sizeof(A028) == 3, "");
```

https://godbolt.org/z/sWxq6dW7Y

I'm not entirely sure how that is supposed to work but I don't think it can possibly fit with our current enum representation since we have to pick a concrete integer type, but nothing is actually 3 bytes.

ehaas:
When adding a _Complex __int24 test I noticed that our `sizeof` for `_Complex` types was wrong!

Fixes #1089

### DIFF
--- a/src/aro/Tokenizer.zig
+++ b/src/aro/Tokenizer.zig
@@ -285,6 +285,8 @@ pub const Token = struct {
         keyword_thread,
         /// __float128
         keyword_float128_1,
+        keyword_int24,
+        keyword_uint24,
         keyword_int128,
         keyword_imag1,
         keyword_imag2,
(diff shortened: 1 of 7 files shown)


src/aro/Tokenizer.zig
```diff
@@ -285,6 +285,8 @@ pub const Token = struct {
         keyword_thread,
         /// __float128
         keyword_float128_1,
+        keyword_int24,
+        keyword_uint24,
```
Vexu: `__uint24` should be an implicit typedef.

Vexu (approved):

PR #1104, opened 2026-09-05
Compilation: add target OS macros

vancluever:
This commit adds the target OS macros as they currently exist in LLVM 22.

The logic around adding them is designed to mirror how `TargetOSMacros.def` defines them, e.g., checking the triple for each macro versus say handling it in a switch per OS/abi/etc.

This also 1) adds a helper to make defining boolean values for system defines, and 2) adds a few extra bits to ensure that targeting visionOS does not cause crashes due to unreachable code.

### DIFF
--- a/src/aro/Compilation.zig
+++ b/src/aro/Compilation.zig
@@ -275,6 +275,11 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
             , .{ name, name });
         }
     }.defineStd;
+    const defineBool = struct {
+        fn defineBool(_w: *Io.Writer, name: []const u8, value: bool) !void {
+            try _w.print("#define {s} {d}\n", .{ name, @intFromBool(value) });
+        }
+    }.defineBool;
     const target = &comp.target;
     const ptr_width = target.ptrBitWidth();
     const is_gnu = comp.langopts.standard.isGNU();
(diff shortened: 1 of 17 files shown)


src/aro/Compilation.zig
```diff
@@ -336,6 +341,69 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
+        // Windows
+        try defineBool(w, "TARGET_OS_WIN32", target.os.tag == .windows);
+        try defineBool(w, "TARGET_OS_WINDOWS", target.os.tag == .windows);
+
+        // Linux
+        try defineBool(w, "TARGET_OS_LINUX", target.os.tag == .linux);
+
+        // Unix
+        //
+        // Note that LLVM has Solaris too but it seems to be missing from Zig.
```
Vexu: Solaris support was intentionally removed https://github.com/ziglang/zig/pull/25706.

src/aro/Compilation.zig
```diff
@@ -336,6 +341,69 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
+            switch (target.os.tag) {
+                .freebsd, .openbsd, .netbsd => true,
+                else => false,
+            },
+        );
+
+        // Apple Targets
+        try defineBool(w, "TARGET_OS_MAC", target.os.tag.isDarwin());
+        // NOTE: This mirrors LLVM isMacOSX
+        try defineBool(w, "TARGET_OS_OSX", target.os.tag.isDarwin() and target.os.tag == .macos);
```
Vexu:
```suggestion
        try defineBool(w, "TARGET_OS_OSX", target.os.tag == .macos);
```

src/aro/Compilation.zig
```diff
@@ -336,6 +341,69 @@ fn generateSystemDefines(comp: *Compilation, w: *Io.Writer) !void {
...
+            switch (target.os.tag) {
+                .ios, .tvos, .watchos => true,
+                else => false,
+            },
+        );
+        try defineBool(w, "TARGET_OS_IOS", target.os.tag == .ios);
+        try defineBool(w, "TARGET_OS_TV", target.os.tag == .tvos);
+        try defineBool(w, "TARGET_OS_WATCH", target.os.tag == .watchos);
+        // Currently no visionOS support in Zig
+        try defineBool(w, "TARGET_OS_VISION", false);
```
Vexu: This should just check for `.visionos` no?

vancluever: I swear I looked for this and didn't see it 🤦‍♂️ will fix!

src/aro/Compilation.zig
vancluever: :thumbsup: thoughts on adding `.illumos` to this macro in lieu?

vancluever: @vexu I just went and added `.illumos` - I think this should be correct for this macro at least (I see https://github.com/llvm/llvm-project/issues/53919, just briefly perusing LLVM issues, so I'm guessing if anything, LLVM has the issue of not making enough of a distinction between Solaris and IllumOS, something we wouldn't have to worry about).

vancluever:
ty again @Vexu!

PR #1106, opened 2026-09-06
Compilation: define __BLOCKS__ when blocks are enabled

vancluever:
This is a re-submission of #1074.

@Vexu can this be merged ahead of #971? Just to make sure we don't get regressions over in Ghostty when backporting (and working on future fixes on master as well).


vancluever:
ty @Vexu!

